PR #1, opened 2016-11-22
Install libjemalloc1

marian-pritsak:
redis-tools package depends on libjemalloc1


qiluo-msft:
We should remove it from the docker-base. It is only used in docker-database.

PR #4, opened 2017-02-01
Added debug packages installation when SONIC_CONFIG_DEBUG enabled.

antonpatenko:
When SONIC_CONFIG_DEBUG enabled, this script performs installation of the debug packages, listed in docker-base.mk


Dockerfile.j2
```diff
@@ -0,0 +1,48 @@
...
+RUN apt-get -y purge                    \
+    exim4                               \
+    exim4-base                          \
+    exim4-config                        \
+    exim4-daemon-light
+
+## Install common debug-packages
+{% if docker_base_dbgs %}
+RUN apt-get -y install \
+{% for dbg_pkg in docker_base_dbgs.split(' ') -%}
```
marian-pritsak: No need to keep list of these packages in buildimage since they are not built bu it but installed directly from apt instead
Just put a list of packages here and use debug config directly in conditional

Dockerfile.j2
```diff
@@ -0,0 +1,48 @@
...
+
+COPY rsyslog.conf /etc/rsyslog.conf
+
+RUN apt-get -y purge                    \
+    exim4                               \
+    exim4-base                          \
+    exim4-config                        \
+    exim4-daemon-light
+
+## Install common debug-packages
```
marian-pritsak: Comment is relevant only when debug option is enabled
Move it inside conditional check

Dockerfile.j2
```diff
@@ -33,9 +33,16 @@ RUN apt-get -y purge                    \
...
 
-## Clean up apt
-## Remove /var/lib/apt/lists/*, could be obsoleted for derived images
-RUN apt-get clean -y; apt-get autoclean -y; apt-get autoremove -y;      \
-    rm -rf /var/lib/apt/lists/*;                                        \
-    rm -rf /tmp/*;
+{% if docker_base_dbgs %}
+## Install common debug-packages
+RUN apt-get -y install \
+{% for dbg_pkg in docker_base_dbgs.split(' ') -%}
```
qiluo-msft: >dbg_pkg [](start = 7, length = 7)

What is dbg_pkg for docker-base?

Dockerfile.j2
```diff
@@ -33,9 +33,16 @@ RUN apt-get -y purge                    \
...
+    {{ dbg_pkg }}{{' '}}
+{%- endfor %}
+{% endif %}
 
+## Clean up
+RUN apt-get clean -y; apt-get autoclean -y; apt-get autoremove -y
+## Note: NO removing /var/lib/apt/lists/*, shared by all derived images
```
qiluo-msft: Why change this part? /var/lib/apt/lists/ will possibly be obsoleted and there is no benefit to be share then.

Dockerfile.j2
antonpatenko: See this pull request.
https://github.com/Azure/sonic-buildimage/pull/234

PR #5, opened 2017-06-06
Add sonic-ledd (Front-panel LED control daemon)

jleveque:
- Also add README.md

### DIFF
--- a/sonic-ledd/scripts/ledd
+++ b/sonic-ledd/scripts/ledd
@@ -216,7 +216,7 @@ def main():
                 port_status_dict[key] = status
                 led_control.port_link_state_change(key, status)
 
-        time.sleep(0.25)
+        time.sleep(1)
         pass
 
 if __name__ == '__main__':
--- a/sonic-ledd/scripts/ledd
+++ b/sonic-ledd/scripts/ledd
@@ -1,0 +1,224 @@
+#!/usr/bin/env python
+#
+# ledd
+#
+# Front-panel LED control daemon for SONiC
+#
+
+try:
+    import ast
+    import getopt
+    import os
+    import imp
+    import signal
+    import subprocess
+    import swsssdk
+    import sys
+    import syslog
+    import time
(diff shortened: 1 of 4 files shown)


sonic-ledd/scripts/ledd
```diff
@@ -0,0 +1,224 @@
...
+    while True:
+        for (key, value) in port_status_dict.iteritems():
+            status = swss.get(swss.APPL_DB, PORT_TABLE_PREFIX + key, 'oper_status')
+
+            # If the status has changed, update it in our dictionary and report to our plugin
+            if value != status:
+                port_status_dict[key] = status
+                led_control.port_link_state_change(key, status)
+
+        time.sleep(0.25)
```
lguohan: >time.sleep(0.25) [](start = 8, length = 16)

by default sleep 1 second.

sonic-ledd/scripts/ledd
```diff
@@ -0,0 +1,224 @@
...
+    for key in port_table_keys:
+        # Remove table name prefix
+        port_name = key[len(PORT_TABLE_PREFIX):]
+
+        # TODO: Once the DB is fixed and this 'ConfigDone' entry is gone, remove this check
+        if port_name == 'ConfigDone':
+            continue
+
+        port_status_dict[port_name] = 'down'
+        led_control.port_link_state_change(port_name, port_status_dict[port_name])
```
lguohan: one problem is that you only get the port_table once, it may not be a complete list.

sonic-ledd/scripts/ledd
jleveque: Resolved.

sonic-ledd/scripts/ledd
lguohan: temporarily, we can include get port in the while loop.

jleveque: Resolved.

PR #8, opened 2017-06-21
wait till the PORT_TABLE in appdb is initialized

lguohan:


sonic-ledd/scripts/ledd
```diff
@@ -208,6 +208,11 @@ def main():
         # Get a list of all ports from the database
         port_table_keys = swss.keys(swss.APPL_DB, PORT_TABLE_PREFIX + '*');
 
+        # when db is not intialized and port_table_keys is None, sleep for 1 second and retry
+        if port_table_keys == None:
+            time.sleep(1)
+            continue
```
qiluo-msft: How about using blocking?
'blocking = True' have a timeout 30s. If it fits here, you can use.

PR #13, opened 2018-04-10
[ledd] Now subscribe to notifications from APPL_DB rather than polling

jleveque:
Refactor ledd such that it no longer periodically polls for interface status changes, and instead subscribes to interface status changes from APPL_DB.

### DIFF
--- a/sonic-ledd/scripts/ledd
+++ b/sonic-ledd/scripts/ledd
@@ -204,7 +204,7 @@ def main():
 
     # Listen indefinitely for changes to the PORT table in the Application DB
     while True:
-        (state, c, fd) = sel.select()
+        (state, c) = sel.select()
         if state != swsscommon.Select.OBJECT:
             log_warning("sel.select() did not return swsscommon.Select.OBJECT")
             continue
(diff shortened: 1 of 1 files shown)


sonic-ledd/scripts/ledd
```diff
@@ -188,38 +186,40 @@ def main():
...
+    # Subscribe to PORT table notifications in the Application DB
+    sel = swsscommon.Select()
+    sst = swsscommon.SubscriberStateTable(appl_db, swsscommon.APP_PORT_TABLE_NAME)
+    sel.addSelectable(sst)
 
-        # Get a list of all ports from the database
-        port_table_keys = swss.keys(swss.APPL_DB, PORT_TABLE_PREFIX + '*', blocking = True);
+    # Listen indefinitely for changes to the PORT table in the Application DB
+    while True:
+        (state, c, fd) = sel.select()
```
pavel-shirshov: Can you please change it to  (state, c) = sel.select()
The interface was changed just recently.

jleveque: Done. Thanks for the heads-up, Pavel!

PR #20, opened 2018-12-18
[sonic-psud] add a new daemon sonic-psud to platform monitor

kevinwangsk:
* This new daemon will periodcally read PSU presence and status from sysfs and write to DB
* CLI will get the PSU presence and status from DB directly.

Signed-off-by: Kevin Wang <kevinw@mellanox.com>

### DIFF
--- a/sonic-psud/scripts/psud
+++ b/sonic-psud/scripts/psud
@@ -1,0 +1,190 @@
+#!/usr/bin/env python2
+
+"""
+    Psud
+    PSU information update daemon for SONiC
+    This daemon will loop to collect PSU related information and then write the information to state DB.
+    Currently it is implemented based on old plugins rather than new platform APIs. So the PSU information just
+    includes three things: number of PSU, PSU presence and PSU status which is supported by old plugins.
+    The loop interval is PSU_INFO_UPDATE_PERIOD_SECS in seconds.
+"""
+
+try:
+    import getopt
+    import os
+    import imp
+    import signal
+    import subprocess
+    import sys
+    import syslog
+    import time
+    from swsscommon import swsscommon
+except ImportError, e:
+    raise ImportError (str(e) + " - required module not found")
+
+#============================= Constants =============================
+
+VERSION = '1.0'
+
+SYSLOG_IDENTIFIER = os.path.basename(__file__)
+PLATFORM_SPECIFIC_MODULE_NAME = "psuutil"
+PLATFORM_SPECIFIC_CLASS_NAME = "PsuUtil"
+
+# Platform root directory inside docker
+PLATFORM_ROOT_DOCKER = "/usr/share/sonic/platform"
+SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
+HWSKU_KEY = 'DEVICE_METADATA.localhost.hwsku'
+PLATFORM_KEY = 'DEVICE_METADATA.localhost.platform'
+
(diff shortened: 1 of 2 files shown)


sonic-psud/scripts/psud
```diff
@@ -0,0 +1,188 @@
...
+    # Load platform-specific psuutil class
+    err = load_platform_psuutil()
+    if err != 0:
+        log_error("failed to load psuutil")
+        sys.exit(1)
+
+    state_db = swsscommon.DBConnector(swsscommon.STATE_DB,
+                                     REDIS_HOSTNAME,
+                                     REDIS_PORT,
+                                     REDIS_TIMEOUT_MSECS)
```
jleveque: Add one more leading space to lines 165-167 so the arguments visually align with the first argument on line 164.

sonic-psud/scripts/psud
```diff
@@ -0,0 +1,188 @@
...
+    psu_tbl = swsscommon.Table(state_db, "PSU_INFO")
+    chassis_tbl = swsscommon.Table(state_db, "CHASSIS_INFO")
+    psu_num = platform_psuutil.get_num_psus()
```
jleveque: I think this variable should be renamed either `num_psus` or `psu_count` for clarity.

sonic-psud/scripts/psud
```diff
@@ -0,0 +1,188 @@
...
+                                stderr=subprocess.STDOUT)
+        stdout = proc.communicate()[0]
+        proc.wait()
+        hwsku = stdout.rstrip('\n')
+    except OSError, e:
+        raise OSError("Cannot detect platform")
+
+    return (platform, hwsku)
+
+
```
liat-grozovik: Remove extra line

sonic-psud/scripts/psud
```diff
@@ -0,0 +1,188 @@
...
+# Loads platform specific psuutil module from source
+def load_platform_psuutil():
```
liat-grozovik: I believe load platform will be the same for each daemon and while using the defines such as PLATFORM_SPECIFIC_MODULE_NAME and PLATFORM_SPECIFIC_CLASS_NAME can make it a generic utility for each daemon in pmon. What do u think?

sonic-psud/scripts/psud
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env python2
```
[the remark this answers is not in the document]
kevinwangsk: Just follow the convention of other daemons

sonic-psud/scripts/psud
```diff
@@ -0,0 +1,188 @@
...
+
+REDIS_HOSTNAME = "localhost"
+REDIS_PORT = 6379
+REDIS_TIMEOUT_MSECS = 0
+
+PSU_INFO_UPDATE_PERIOD_SECS = 3
+
+#========================== Syslog wrappers ==========================
+
+def log_info(msg, also_print_to_console=False):
```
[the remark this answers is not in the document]
kevinwangsk: Yes, it is better to have a common file for the generic function of Pmon daemons. But I would suggest it should not in this commit as it is to implement PSU daemon not to improve the generic functions.

sonic-psud/scripts/psud
kevinwangsk: Yes, same with the above comments. Should be in a separate commit to abstract them, not in this commit.

PR #22, opened 2019-02-19
[sonic-platform-daemons] Refactor daemons based on DaemonBase class 

kevinwangsk:
Signed-off-by: Kevin Wang <kevinw@mellanox.com>

### DIFF
--- a/sonic-ledd/scripts/ledd
+++ b/sonic-ledd/scripts/ledd
@@ -15,6 +15,7 @@ try:
     import sys
     import syslog
     from swsscommon import swsscommon
+    from sonic_daemon_base.daemon_base import DaemonBase
 except ImportError, e:
     raise ImportError (str(e) + " - required module not found")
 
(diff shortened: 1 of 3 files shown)


kevinwangsk:
Joe, please help to have a test with ledd. Thanks.
This PR should not be merged until the parent PR https://github.com/Azure/sonic-buildimage/pull/2570 is merged in sonic-buildimage repo.

sonic-ledd/scripts/ledd
```diff
@@ -34,133 +35,18 @@ Options:
...
-
-    log_info("Instantiated class '%s.%s'." % (LED_MODULE_NAME, LED_CLASS_NAME))
-
-    return led_control
-
 #=============================== Main ================================
 
 def main():
-    log_info("Starting up...")
+    ledd_util = DaemonBase()
```
andriymoroz-mlnx: hmm, this breaks the entire concept of having the base class...
I suggest to declare
`class DaemonLed(DaemonBase):`
and use DaemonLed even if it does not override anything

sonic-ledd/scripts/ledd
```diff
@@ -183,21 +69,17 @@ def main():
...
+    if not led_control:
+        ledd_util.log_error("failed to load ledutil")
+        sys.exit(1)
 
     # Open a handle to the Application database
     appl_db = swsscommon.DBConnector(swsscommon.APPL_DB,
-                                     REDIS_HOSTNAME,
-                                     REDIS_PORT,
-                                     REDIS_TIMEOUT_USECS)
+                                     ledd_util.redis_hostname,
```
andriymoroz-mlnx: DB connector creation can also go to base class. 3 of 4 parameters belong to base
here can be only:
appl_db = my_daemon.DBConnect(APPL_DB)

sonic-ledd/scripts/ledd
```diff
@@ -34,133 +35,25 @@ Options:
...
+class DaemonLedd(DaemonBase):
+    def __init__(self):
+        DaemonBase.__init__(self)
+
+    def __exit__(self):
+        DaemonBase.__exit__(self)
 
 #=============================== Main ================================
 
 def main():
```
jleveque: Suggest making this function a method of the daemon class and calling it `run()`. This function can be declared in the DaemonBase class which raises a `NotImplementedError` because it is expected each daemon which inherits the base class needs to implement its own `run()` method.

Then, the `main()` function in each daemon would resemble the following:

```
def main():
    daemon_ledd = DaemonLedd()
    if not daemon_ledd:
        print "Failed to instantiate LED daemon"
        sys.exit(1)

    daemon_ledd.run()
```

This applies to all daemons.

sonic-ledd/scripts/ledd
```diff
@@ -34,133 +35,25 @@ Options:
...
-    log_info("Starting up...")
+    ledd_util = DaemonLedd()
```
jleveque: Suggest renaming object from `ledd_util` to `daemon_led` as detailed in my comment above. This applies to all daemons.

sonic-ledd/scripts/ledd
kevinwangsk: Done

sonic-ledd/scripts/ledd
kevinwangsk: Done

sonic-ledd/scripts/ledd
```diff
@@ -34,200 +35,88 @@ Options:
...
-PLATFORM_KEY = "platform"
+class DaemonLedd(DaemonBase):
+    def __init__(self):
+        DaemonBase.__init__(self)
+
+    def __exit__(self):
+        DaemonBase.__exit__(self)
+
+    def run(self):
+        if not os.geteuid() == 0:
```
jleveque: I feel like this check for root user belongs at the top of the `main()` function, and not in the daemon class. @kevinwangsk: do you agree here?

sonic-xcvrd/scripts/xcvrd
```diff
@@ -348,107 +314,26 @@ class dom_info_update_task:
         self.task_stopping_event.set()
         self.task_timer.join()
 
-#=============================== Main ================================
+# Returns path to port config file
+def get_path_to_port_config_file():
```
jleveque: Should this function become a method of DaemonBase class so it can be shared among daemons, if needed?

sonic-xcvrd/scripts/xcvrd
```diff
@@ -61,122 +41,108 @@ VOLT_UNIT = 'Volts'
...
 
-    return port_config_file_path
+        # Clean all the information from DB and then exit
+        logical_port_list = platform_sfputil.logical
+        for logical_port_name in logical_port_list:
+            del_port_sfp_dom_info_to_db(logical_port_name, int_tbl, dom_tbl)
+        return 1
 
 # find out the underneath physical port list by logical name
 def logical_port_name_to_physical_port_list(port_name):
```
jleveque: Should this function become a method of DaemonXcvrd class?

sonic-ledd/scripts/ledd
kevinwangsk: Agree.

sonic-xcvrd/scripts/xcvrd
kevinwangsk: Given that this function will only be used by xcvrd, I think we can put this function to DaemonXcvrd rather than DaemonBase.

jleveque: OK. We can consider moving it in the future if it ever needs to be shared.

PR #25, opened 2019-03-15
[syseeprom] Add a new daemon to populate syseeprom info to state DB

keboliu:
syseeprom:  Add a new daemon to populate the syseeprom info to state DB

* This daemon intends to collect the syseeprom information and write it to state DB, so CLI can get this info from DB instead of access HW. This daemon will keep monitoring the DB table, if table was deleted, it will write the info to DB again.

There will be a followed PR to add make files and modify start.sh after this one merged.

Signed-off-by: Kebo Liu kebol@mellanox.com


sonic-post-syseeprom/scripts/post-syseeprom
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python2
+
+'''
+    post-syseeprom
+    Syseeprom infomation gathering task for SONiC
```
jleveque: Typo: 'information'

sonic-post-syseeprom/scripts/post-syseeprom
```diff
@@ -0,0 +1,47 @@
...
+
+def main():
+    Daemon_Base = DaemonBase()
+    (platform_path, hwsku_path) = Daemon_Base.get_path_to_platform_and_hwsku()
+    
+    # Arista platform have their own implementation for eeprom tlv parsing, it's not inherited from eeprom base class.
+    # So for now this task can not support Arista platform, but this will not cause anny issue on them. 
+    # decode-syseeprom scripts will cover Arista case, it will still read from hw/cache instead from DB on Arista platform.  
+    if 'arista' in platform_path:
+        Daemon_Base.log_warning('Arista platform not support this yet')
```
jleveque: Add "does": "Arista platform __does__ not support this yet"

sonic-post-syseeprom/scripts/post-syseeprom
```diff
@@ -0,0 +1,47 @@
...
+    eeprom_util = Daemon_Base.load_platform_util(PLATFORM_SPECIFIC_MODULE_NAME, PLATFORM_SPECIFIC_CLASS_NAME)
+
+    eeprom = eeprom_util.read_eeprom()
+    if eeprom is None :
+        Daemon_Base.log_error('Failed to read eeprom')
+        return ERR_FAILED_EEPROM
+
+    err = eeprom_util.update_eeprom_db(eeprom)
+    if err:
+        Daemon_Base.log_error('Failed to update eeprom to database')
```
jleveque: Reword: "Failed to update eeprom info in database"

sonic-post-syseeprom/setup.py
```diff
@@ -0,0 +1,29 @@
+from setuptools import setup
+
+setup(
+    name='sonic-post-syseeprom',
+    version='1.0',
+    description='Syseeprom gathering task for SONiC',
```
jleveque: Suggest change description to "Tool which posts syseeprom info to DB in SONiC"

sonic-post-syseeprom/scripts/post-syseeprom
keboliu: fixed

sonic-post-syseeprom/scripts/post-syseeprom
keboliu: revised

sonic-post-syseeprom/scripts/post-syseeprom
keboliu: revised

sonic-post-syseeprom/setup.py
keboliu: revised

lguohan:
is this eeprom information in state db? the state db is clearly after sonic config load or swss restart. we need to deal with such situation.

keboliu:
@lguohan the plan is that after this PR merged there will be another PR to call this script in the 'start.sh' of pmon container and make change to related make file in sonic-buildimage repo, I think 'start.sh' will be executed during config reload/swss restart thus this script will post the syseeprom info to state DB again, is this sound good to you?

keboliu:
@lguohan added '-w' and '-c' options to the scripts to handle write DB and clear DB functions respectively.

sonic-post-syseeprom/scripts/post-syseeprom
```diff
@@ -0,0 +1,64 @@
...
+    # decode-syseeprom scripts will cover Arista case, it will still read from hw/cache instead from DB on Arista platform.  
+    if 'arista' in platform_path:
+        Daemon_Base.log_warning('Arista platform does not support this yet')
+        return ERR_ARISTA_PLATFORM
+
+    if args.w:
+        eeprom_util = Daemon_Base.load_platform_util(PLATFORM_SPECIFIC_MODULE_NAME, PLATFORM_SPECIFIC_CLASS_NAME)
+        eeprom = eeprom_util.read_eeprom()
+        if eeprom is None :
+            Daemon_Base.log_error('Failed to read eeprom')
```
nazariig: @keboliu This is invalid logging API. Please use Logger from the sonic_daemon_base.daemon_base.

sonic-post-syseeprom/scripts/post-syseeprom
```diff
@@ -0,0 +1,64 @@
...
+            return ERR_FAILED_EEPROM
+
+        err = eeprom_util.update_eeprom_db(eeprom)
+        if err:
+            Daemon_Base.log_error('Failed to update eeprom info to database')
```
nazariig: @keboliu This is invalid logging API. Please use Logger from the sonic_daemon_base.daemon_base.

sonic-post-syseeprom/scripts/post-syseeprom
```diff
@@ -0,0 +1,64 @@
...
+            return ERR_FAILED_UPDATE_DB
+    elif args.c:
+        state_db = Daemon_Base.db_connect(swsscommon.STATE_DB)
+        eeprom_tbl = swsscommon.Table(state_db, EEPROM_TABLE_NAME)
+        keys = eeprom_tbl.getKeys()
+        for key in keys:
+            eeprom_tbl._del(key)
+    else:
+        Daemon_Base.log_error("Invalid parameter")
```
nazariig: @keboliu This is invalid logging API. Please use Logger from the sonic_daemon_base.daemon_base.

sonic-post-syseeprom/scripts/post-syseeprom
jleveque: To add to this, the DaemonBase class is intended to be inherited by the main class of a daemon. I think it's odd to instantiate an object of type DaemonBase just to call the methods within it.

I understand this is a one-shot application, not a daemon, but it might be better to inherit the DaemonBase class, since it doesn't currently have any dameon-specific functionality (forking, etc.). If we ever consider making DaemonBase do things like forking, then we can consider breaking out another subclass which handles logging, etc.

keboliu: I refactored this scripts with adding a class which inherits DaemonBase, and rewrite the log code with latest Logger class.

sonic-post-syseeprom/scripts/post-syseeprom
keboliu: revised

sonic-post-syseeprom/scripts/post-syseeprom
keboliu: revised

sonic-post-syseeprom/scripts/post-syseeprom
```diff
@@ -0,0 +1,90 @@
...
+            return ERR_ARISTA_PLATFORM
+
+        state_db = daemon_base.db_connect(swsscommon.STATE_DB)
+        eeprom_tbl = swsscommon.Table(state_db, EEPROM_TABLE_NAME)
+        keys = eeprom_tbl.getKeys()
+        for key in keys:
+            eeprom_tbl._del(key)
+        
+
+def main():
```
jleveque: When inheriting DaemonBase, the main() function should end with a call to my_daemon.run(). This isn't a daemon, and it accepts arguments, so it really doesn't make sense here. I don't think DaemonBase is the appropriate choice here after all.

I think eventually, the Logger class should be broken out of the daemon_base file and placed in a separate package with other common functions all SONiC applications can share (including DaemonBase).

I think you should revert this change, and simply use

```
    from sonic_daemon_base.daemon_base import Logger
```

and

```
SYSLOG_IDENTIFIER = "post-syseeprom"

# Global logger class instance
logger = Logger(SYSLOG_IDENTIFIER)
```

Sorry for the confusion and extra work.

keboliu: agree, changed.

nazariig: > I think eventually, the Logger class should be broken out of the daemon_base file and placed in a separate package with other common functions all SONiC applications can share (including DaemonBase).

But sonic_daemon_base initially was designed in the way to implement DaemonBase.
What's the point in moving DaemonBase from sonic_daemon_base to a separate package eventually?

IMHO, the design of sonic-daemon-base was initially incorrect: DaemonBase should be refactored and equipped only with the mandatory API (signal handlers and run method). The other stuff should be moved to daemon_base directly. Such approach will completely solve all your review comments but requires some extra work to do with aligning the existing stuff of sonic-platform-daemons.

P.S: for logging purposes we can abandon our custom Logger implementation and consider using python embedded mechanisms/libraries.

jleveque: @nazariig: I didn't mention moving the DaemonBase class out of sonic_daemon_base.py, but rather moving the Logger class somewhere where it can be shared, along with other functions that we might want to share between daemons and also one-shot applications. Just a thought to keep in mind as SONiC grows -- how can we reuse code most efficiently.

liat-grozovik:
@jleveque @lguohan can you please review? this PR is the last piece of the pmon refactoring to be part of 201904

lguohan:
it does not address my comments : "need to deal with state clear due to config reload or swss restart."

nazariig:
> 
> 
> it does not address my comments : "need to deal with state clear due to config reload or swss restart."

@lguohan can we consider having an EEPROM daemon?

henry-ma:
@lguohan could you please approve this or raise your comments? We should get it merged as part of Pmon enhancement work.

keboliu:
> it does not address my comments : "need to deal with state clear due to config reload or swss restart."

@lguohan, as I mentioned in the description there will be a followd PR against sonic-buildimage which will include the code to call this utility to post info to DB during pmon start and clear DB when pmon stopped.  I was planning to create it after this one merged.  
Now in order to give you a whole picture of this solution, I just submitted that one: https://github.com/Azure/sonic-buildimage/pull/2866  you can find the code where to clear DB when pmon stop, please also review.

keboliu:
> > it does not address my comments : "need to deal with state clear due to config reload or swss restart."
> 
> @lguohan can we consider having an EEPROM daemon?

@nazariig I think it kind of overkill to have a daemon here since eeprom info never changed.

sonic-post-syseeprom/scripts/post-syseeprom
```diff
@@ -0,0 +1,70 @@
...
+
+EEPROM_TABLE_NAME = 'EEPROM_INFO'
+SYSLOG_IDENTIFIER = 'post-syseeprom'
+
+# Global logger class instance
+logger = Logger(SYSLOG_IDENTIFIER)
+
+def main():
+    parser = argparse.ArgumentParser(description='post-syseeprom tool')
+    parser.add_argument("-w", help="write syseeprom info into state DB", action='store_true')
```
qiluo-msft: >"-w" [](start = 24, length = 4)

you can also give a long name such as '--write' and make your code more readable 'args.write'

sonic-post-syseeprom/scripts/post-syseeprom
```diff
@@ -0,0 +1,70 @@
...
+    from sonic_daemon_base import daemon_base
+    import argparse
+    from swsscommon import swsscommon
+except ImportError, e:
+    raise ImportError (str(e) + " - required module not found")
+
+PLATFORM_SPECIFIC_MODULE_NAME = 'eeprom'
+PLATFORM_SPECIFIC_CLASS_NAME = 'board'
+
+ERR_ARISTA_PLATFORM = 1
```
qiluo-msft: >ERR_ARISTA_PLATFORM [](start = 0, length = 19)

ERR_PLATFORM_NOTSUPPORT ?

sonic-post-syseeprom/scripts/post-syseeprom
```diff
@@ -0,0 +1,70 @@
...
+        return ERR_ARISTA_PLATFORM
+
+    if args.w:
+        eeprom_util = Daemon_Base.load_platform_util(PLATFORM_SPECIFIC_MODULE_NAME, PLATFORM_SPECIFIC_CLASS_NAME)
+        eeprom = eeprom_util.read_eeprom()
+        if eeprom is None :
+            logger.log_error("Failed to read eeprom")
+            return ERR_FAILED_EEPROM
+
+        err = eeprom_util.update_eeprom_db(eeprom)
```
qiluo-msft: >update_eeprom_db [](start = 26, length = 16)

Are you going to write MAC address into StateDB? If yes, there is already one in ConfigDB, how do you handle data consistency?

keboliu: The purpose here is to put all the syseeprom info into state DB and afterward, all the request from CLI or somewhere else to check the syseeprom info will access the state DB instead of reading the from the hardware.  The one in the config DB may or may not consistent with the one burnt to syseeprom and depends on the implementaion of specific platform?  anyway these info is only for show not for other purpose.

sonic-post-syseeprom/scripts/post-syseeprom
keboliu: revised.

sonic-post-syseeprom/scripts/post-syseeprom
keboliu: add long arguments names.

sonic-post-syseeprom/scripts/post-syseeprom
qiluo-msft: Please check old PR: [https://github.com/Azure/sonic-buildimage/pull/1397](https://github.com/Azure/sonic-buildimage/pull/1397)

Digging into the history, we introduced 'DEVICE_METADATA|localhost :: mac' only for Mellanox platform, because other platform could easily get the mac from eth0.

Here you introduce a duplication of mac in Redis database. Could you please consolidate into one?

---
In reply to: [284053727](https://github.com/Azure/sonic-platform-daemons/pull/25#discussion_r284053727) [](ancestors = 284053727)

sonic-post-syseeprom/scripts/post-syseeprom
```diff
@@ -0,0 +1,70 @@
...
+        state_db = daemon_base.db_connect(swsscommon.STATE_DB)
+        eeprom_tbl = swsscommon.Table(state_db, EEPROM_TABLE_NAME)
+        keys = eeprom_tbl.getKeys()
+        for key in keys:
+            eeprom_tbl._del(key)
+    else:
+        logger.log_error("Invalid parameter")
+
+if __name__ == '__main__':
+    main()
```
qiluo-msft: >main() [](start = 4, length = 6)

Did you return error code in all cases?

keboliu: Added return code for a missed case.

sonic-post-syseeprom/scripts/post-syseeprom
keboliu: I know there is lots of twist on how to get the switch mac address. 

- The mac address in config db is essential to the start of syncd(or swss?), need to be fetched from syseeprom at an early stage of switch starting, in this PR. the syseeprom info(including the mac address) is read in a later phase(at the starting of the pmon), so we can not feed the mac address in config DB with this scripts.

- on the other hand, if we want to skip the mac address in this script and try to reuse the mac address in config db, then we need to refactor the "decode-syseeprom" code to get the base mac address from another source, that will cause the code more complicated, I am thinking the benefit what we can get from reducing the mac address here is not so significant.

- you mentioned that on other platforms just directly use the "eth0" as the mac address,  in this scripts the mac adress read from syseeprom is the base mac, usually they are different from the "eth0" mac address.

qiluo-msft: I think you still need `return main()`

---
In reply to: [283998899](https://github.com/Azure/sonic-platform-daemons/pull/25#discussion_r283998899) [](ancestors = 283998899)

qiluo-msft: I don't get the point here. However, I unblocked this PR and try to discuss offline and get better understanding of the motivation.

---
In reply to: [285456136](https://github.com/Azure/sonic-platform-daemons/pull/25#discussion_r285456136) [](ancestors = 285456136)

keboliu: > I think you still need `return main()`
> 
> In reply to: [283998899](https://github.com/Azure/sonic-platform-daemons/pull/25#discussion_r283998899) [](ancestors = 283998899)

amended.

keboliu:
@lguohan I have addressed new comments with Qi and Joe, would you please approve if it's OK to you?

sonic-syseepromd/scripts/syseepromd
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python2
+
+'''
+    syseepromd
+    Syseeprom information gathering daemon for SONiC
+    This daemon will be started during the start phase of pmon container, gathering syseeprom info and write to state DB.
+    It will continue monitoring the state DB for the syseeprom table, it table deleted, will write again.
```
stcheng: if

keboliu: revised, thanks.

sonic-syseepromd/scripts/syseepromd
```diff
@@ -0,0 +1,143 @@
...
+        
+        state_db = daemon_base.db_connect(swsscommon.STATE_DB)
+        self.eeprom_tbl = swsscommon.Table(state_db, EEPROM_TABLE_NAME)
+        self.eepromtbl_keys = []
+
+    def load_eeprom_util(self):
+        try:
+            self.eeprom_util = self.load_platform_util(PLATFORM_SPECIFIC_MODULE_NAME, PLATFORM_SPECIFIC_CLASS_NAME)
+        except Exception as e:
+            logger.log_error("Failed to load psuutil: %s" % (str(e)), True)
```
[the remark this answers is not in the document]
keboliu: a copy/paste error, fixed.

sonic-syseepromd/scripts/syseepromd
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python2
+
+'''
+    syseepromd
+    Syseeprom information gathering daemon for SONiC
+    This daemon will be started during the start phase of pmon container, gathering syseeprom info and write to state DB.
+    It will continue monitoring the state DB for the syseeprom table, if table been deleted, it will write again.
```
jleveque: Change wording: "if table been deleted" --> "if table was deleted"

sonic-syseepromd/scripts/syseepromd
```diff
@@ -0,0 +1,143 @@
...
+SYSLOG_IDENTIFIER = 'syseepromd'
+
+# Global logger class instance
+logger = Logger(SYSLOG_IDENTIFIER)
+
+class DaemonSyseeprom(DaemonBase):
+    def __init__(self):
+        DaemonBase.__init__(self)
+
+        self.stop = threading.Event()
```
jleveque: I think the name "stop" can make the code a bit confusing. Suggest rename to "stop_event"

sonic-syseepromd/scripts/syseepromd
keboliu: revised.

sonic-syseepromd/scripts/syseepromd
keboliu: renamed.

jleveque:
@lguohan: Please review the latest changes.

sonic-syseepromd/scripts/syseepromd
```diff
@@ -0,0 +1,143 @@
...
+    from sonic_daemon_base.daemon_base import Logger
+    from sonic_daemon_base import daemon_base
+    from swsscommon import swsscommon
+except ImportError, e:
+    raise ImportError (str(e) + " - required module not found")
+
+PLATFORM_SPECIFIC_MODULE_NAME = 'eeprom'
+PLATFORM_SPECIFIC_CLASS_NAME = 'board'
+
+EEPROM_INFO_UPDATE_PERIOD_SECS = 10
```
lguohan: 60

jleveque:
@keboliu: A couple comments:

1. When you add this daemon to the supervisor config file, make sure you set it to restart always to ensure it's always running.
2. This storage of syseeprom info to STATE_DB should be viewed as a cache to improve performance for non-critical applications (SNMP, streaming telemetry, etc.). However, there are critical applications which rely on obtaining the MAC address. These applications should never fail. Thus, they should rely on calling decode-syseeprom. Therefore, we should not remove the ability for decode-syseeprom to access the hardware directly. At the most, we could modify decode-syseeprom to use the STATE_DB as a cache, and return the data if found, otherwise, it should poll the hardware directly as it currently does.

sonic-syseepromd/scripts/syseepromd
keboliu: changed.

keboliu:
> @keboliu: A couple comments:
> 
> 1. When you add this daemon to the supervisor config file, make sure you set it to restart always to ensure it's always running.

updated in another PR.

> 2. This storage of syseeprom info to STATE_DB should be viewed as a cache to improve performance for non-critical applications (SNMP, streaming telemetry, etc.). However, there are critical applications which rely on obtaining the MAC address. These applications should never fail. Thus, they should rely on calling decode-syseeprom. Therefore, we should not remove the ability for decode-syseeprom to access the hardware directly. At the most, we could modify decode-syseeprom to use the STATE_DB as a cache, and return the data if found, otherwise, it should poll the hardware directly as it currently does.

the current decode-syseeprom logic is implemented as you mentioned, when it failed to get data from DB, it will fall back on access the hardware. So we are ok here.

PR #26, opened 2019-03-28
Dynamic transceiver tuning support

dgsudharsan:
**- What I did**
Dynamic transceiver tuning support
**- How I did it**
Dynamically program pre-emphasis and other settings based on media detected in xcvrd. The various supported media are defined in media_settings.json

**- How to verify it**
Perform OIR of different media types. Dump the pre-emphasis values from hardware and verify that they got reflected

Support for dynamically programming pre-emphasis

### DIFF
--- a/sonic-xcvrd/scripts/xcvrd
+++ b/sonic-xcvrd/scripts/xcvrd
@@ -357,7 +357,7 @@ def get_media_settings_key(physical_port, transceiver_dict):
     if sup_compliance_str in media_compliance_dict:
         media_compliance_code = media_compliance_dict[sup_compliance_str]
 
-    media_type = transceiver_dict[physical_port]['short_type']
+    media_type = transceiver_dict[physical_port]['type_abbrv_name']
 
     media_key = ''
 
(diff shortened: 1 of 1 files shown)


sonic-xcvrd/scripts/xcvrd
```diff
@@ -291,11 +471,17 @@ class sfp_state_update_task:
...
+        transceiver_dict = {}
         # Connect to STATE_DB and create transceiver dom/sfp info tables
         state_db = daemon_base.db_connect(swsscommon.STATE_DB)
         int_tbl = swsscommon.Table(state_db, TRANSCEIVER_INFO_TABLE)
         dom_tbl = swsscommon.Table(state_db, TRANSCEIVER_DOM_SENSOR_TABLE)
 
+        # Connect to ASIC_DB to notify Media notifications
+        asic_db = daemon_base.db_connect(swsscommon.ASIC_DB)
+        media_notifier = swsscommon.NotificationProducer(asic_db,
+                                                         "MEDIANOTIFICATION")
```
lguohan: should be app db.

dgsudharsan:
> There is a corner case that user might insert a SFP during warmboot, In such case that function post_port_sfp_dom_info_to_db seems will skip programming serdes tuning value to asic and will never be programmed for this newly plugged SFP until OIR again.
> Could you consider such case even is a very rare case?

Hi @stevenlu99 Yes I think that event will be very rare case and is it worth to handle such a case? I am not sure if someone might do a warm reboot and in parallel change cabling (It defeats the purpose of warm reboot)

PR #28, opened 2019-05-17
[Event listener] Add a new supervisord event listener to pmon container

keboliu:
Event-listener: Add a new supervisord event listener to pmon container to clear the state DB  when pmon was shut down.

* This event listener will monitor the 'SUPERVISOR_STATE_CHANGE_STOPPING' event of supervisord when it is shutting down.   On this event, the event listener will call command 'post-syseeprom -c' to clear the state DB.

Signed-off-by: Kebo Liu kebol@mellanox.com

### DIFF
--- a/sonic-event-listener/scripts/event-listener
+++ b/sonic-event-listener/scripts/event-listener
@@ -1,0 +1,49 @@
+#!/usr/bin/env python2
+
+'''
+    event-listener
+    Supervisord event listener for SONiC pmon container
+    This event listener will monitor the 'SUPERVISOR_STATE_CHANGE_STOPPING' event of supervisord when it is shutting down.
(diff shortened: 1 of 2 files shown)


jleveque:
Since this is specific to the PMON container, why not add the event listener script in https://github.com/Azure/sonic-buildimage/tree/master/dockers/docker-platform-monitor?

keboliu:
> Since this is specific to the PMON container, why not add the event listener script in https://github.com/Azure/sonic-buildimage/tree/master/dockers/docker-platform-monitor?

I was thinking to do so but not quite sure whether this is a good practice at that time. Since you all prefer this way, I moved it and updated PR https://github.com/Azure/sonic-buildimage/pull/2866

PR #30, opened 2019-06-21
[xcvrd] Fix missing sfp info table entry issue 

keboliu:
Issue and fix description: 
 
- Currently, Xcvrd relies on the "PortConfigDone" flag to start to retrieve SFP info. But in some case, even the "PortConfigDone" flag set, not all the SFP devices are ready, so there could be some SFP info missed in the SFP info table due to the delay. 

-  Add a period SFP table integrity check and recover function can fix this issue.  Another benefit of this fix is that can make sure SFP table info always recovered in any case.

- What I did in this fix
    - add a recover check function to check if all SFP info in the table, if not write it to the table.
    - run this recover function in the main task each 60s

### DIFF
--- a/sonic-xcvrd/scripts/xcvrd
+++ b/sonic-xcvrd/scripts/xcvrd
@@ -38,6 +38,7 @@ SELECT_TIMEOUT_MSECS = 1000
 
 DOM_INFO_UPDATE_PERIOD_SECS = 60
 TIME_FOR_SFP_READY_SECS = 1
+XCVRD_MAIN_THREAD_SLEEP_SECS = 60
 
 SFP_STATUS_INSERTED = '1'
 SFP_STATUS_REMOVED = '0'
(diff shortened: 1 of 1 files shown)


keboliu:
@jleveque would you please help to review when you have time?

sonic-xcvrd/scripts/xcvrd
```diff
@@ -262,6 +263,19 @@ def del_port_sfp_dom_info_from_db(logical_port_name, int_tbl, dom_tbl):
             logger.log_error("This functionality is currently not implemented for this platform")
             sys.exit(NOT_IMPLEMENTED_ERROR)
 
+# recover missing sfp table entries if any
+def recover_missing_sfp_table_entries(sfp_util, int_tbl, stop=threading.Event()):
```
jleveque: Why the default parameter (`stop=threading.Event()`)? I don't think this is what you intended here.

keboliu: you are correct, revised.

keboliu: @jleveque please check the revised code, should be OK now.

sonic-xcvrd/scripts/xcvrd
```diff
@@ -573,7 +587,7 @@ class DaemonXcvrd(DaemonBase):
     def __init__(self):
         DaemonBase.__init__(self)
 
-        self.timeout = 1
+        self.timeout = XCVRD_MAIN_THREAD_SLEEP_SECS
         self.stop = threading.Event()
```
jleveque: Can you rename this "self.stop_event" so that it is clear it is an event, not a bool or something else?

keboliu: @jleveque changed all 'stop' to 'stop_event'.

PR #34, opened 2019-07-25
[xcvrd] recovery SFP modules (backport PR#30 to 201811)

stephenxs:
This is the backport of the [PR#30](https://github.com/Azure/sonic-platform-daemons/pull/30) to 201811 with necessary adjustment.

1. periodically check whether all SFPs' information is in the database and insert it if not.
2. retry for 5 times when a newly inserted SFP module isn't "present"


sonic-xcvrd/scripts/xcvrd
```diff
@@ -440,6 +462,13 @@ def main():
             # TODO: next step need to define more error types to handle accordingly.
             break
 
+        time_now = time.time()
+        time_diff = time_now - time_last_recovery_run
+        # time.time() returns value in unit of seconds while XCVRD_MAIN_THREAD_SLEEP_MSECS is defined in unit of milliseconds
```
liat-grozovik: please remove unused code

stephenxs: Hi Liat, this code is not unused.
this part of code is backported from master which has an independent thread to recover_missing_sfp_table_entries periodically. 
In 201811 we don't have a thread to do that and have to make use of the main thread. to avoid consuming too much cpu time this function shouldn't be called too frequently. currently it is called every minute, which is implemented by adding a timeout argument to get_transceiver_change_event.
However, in case of multiple SFP modules inserted/removed, get_transceiver_change_event can be woken up more frequently thus recover_missing_sfp_table_entries being called several times in a short period.
To avoid that, we implement a time_diff calculation.

PR #35, opened 2019-07-29
[syseepromd] Support both new platform API and old platform plugins

stephenxs:
Support new platform api with plugin mode compatible.
In this PR, new platform api is supported in following steps:
1. initialization, load new-platform-api-based class eeprom or chassis for syseepromd and psud respectively, if failed then load old style plugin.
2. for each api, call new platform api, old style plugin is called if the former raised a NotImplementError; other Exceptions raised will be passed to caller without handling.

now only two APIs are called in syseepromd, update_eeprom_db and read_eeprom.
five three APIs for psud, get_num_psus, get_presence and get_powergood_status.
it depends on the [sonic-platform-common #48 add get_eeprom to Chassis for syseepromd to access](https://github.com/Azure/sonic-platform-common/pull/48)

### DIFF
--- a/sonic-psud/scripts/psud
+++ b/sonic-psud/scripts/psud
@@ -45,16 +45,44 @@ PSUUTIL_LOAD_ERROR = 1
 
 logger = Logger(SYSLOG_IDENTIFIER)
 
+platform_psuutil = None
+platform_chassis = None
+
+# temporary wrappers that are compliable with both new platform api and old-style plugin mode
+def _wrapper_get_num_psus():
+    if platform_chassis is not None:
+        try:
+            return platform_chassis.get_num_psus()
+        except NotImplementedError:
+            pass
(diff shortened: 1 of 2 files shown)


jleveque:
@stephenxs: Do you mind splitting the syseepromd and psud changes into separate PRs? It will make the changes more modular in case of the need to revert anything.

stephenxs:
> @stephenxs: Do you mind splitting the syseepromd and psud changes into separate PRs? It will make the changes more modular in case of the need to revert anything.

Done.
[PR 36](https://github.com/Azure/sonic-platform-daemons/pull/36) for syseeprom and [PR 37](https://github.com/Azure/sonic-platform-daemons/pull/37) for psud

jleveque:
So shall we close this PR?

stephenxs:
sure.

PR #36, opened 2019-07-31
[syseepromd] Support both new platform API and old platform plugins

stephenxs:
Support new platform api with plugin mode compatible.
In this PR, new platform api is supported in following steps:
1. initialization, load new-platform-api-based class eeprom or chassis for syseepromd and psud respectively, if failed then load old style plugin.
2. for each api, call new platform api, old style plugin is called if the former raised a NotImplementError; other Exceptions raised will be passed to caller without handling.

now only two APIs are called in syseepromd, update_eeprom_db and read_eeprom.
it depends on the [sonic-platform-common #48 add get_eeprom to Chassis for syseepromd to access](https://github.com/Azure/sonic-platform-common/pull/48)

### DIFF
--- a/sonic-syseepromd/scripts/syseepromd
+++ b/sonic-syseepromd/scripts/syseepromd
@@ -43,7 +43,8 @@ class DaemonSyseeprom(DaemonBase):
 
         self.stop_event = threading.Event()
         self.eeprom_util = None
-        
+        self.eeprom = None
+
         state_db = daemon_base.db_connect(swsscommon.STATE_DB)
         self.eeprom_tbl = swsscommon.Table(state_db, EEPROM_TABLE_NAME)
         self.eepromtbl_keys = []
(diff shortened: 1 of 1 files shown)


stephenxs:
Hi Joe,
Can it be well handled if another exception thrown from the finally block? I’m not sure whether it’s a good style. Other thing seems good to me.

BTW, psud can also be updated in this way.
But for xcvrd, it is a little complicated, because the sfputil has to be remained even new platform api supported. I’ve mentioned that in another PR https://github.com/Azure/sonic-platform-daemons/pull/38

jleveque:
The finally block is a bit of a messy approach, I agree. Instead, you should be able to simply check for None:

```
        # First, try to load the new platform API
        try:
            import sonic_platform
            self.chassis = sonic_platform.platform.Platform().get_chassis()
            self.eeprom = self.chassis.get_eeprom()
        except Exception as e:
            logger.log_warning("Failed to load platform-specific eeprom from sonic_platform package due to {}".format(repr(e)))

        # If we didn't successfully load the class from the sonic_platform package, try loading the old plugin
        if not self.eeprom:
            try:
                self.eeprom = self.load_platform_util(PLATFORM_SPECIFIC_MODULE_NAME, PLATFORM_SPECIFIC_CLASS_NAME)
            except Exception as e:
                logger.log_error("Failed to load platform-specific eeprom implementation: %s" % (str(e)), True)

        if not self.eeprom:
            sys.exit(ERR_EEPROMUTIL_LOAD)
```

Great to know psud should be able to use the same approach. I assumed xcvrd would require a different solution.

stephenxs:
Eepromd works in this way.
But I don’t think we should remove the _wrapper_* functions if we want to have a per-API level compat.
PSU doesn’t work in this way, since the it requires both chassis and psu object, unlike eeprom which requires eeprom object only.
The root cause is that,

  1.  there can be more than one PSUs in a switch so we have both APIs to access PSU information on a global scope, like get_num_psus which is provided by chassis, and APIs to access each PSU, like get_psu_presence provided by psu class.
  2.  there is only one eeprom in a switch and it doesn’t require any “global” scope APIs.

jleveque:
Understood. So do you want to make any changes to this PR after all?

stephenxs:
> The finally block is a bit of a messy approach, I agree. Instead, you should be able to simply check for None:
> 
> ```
>         # First, try to load the new platform API
>         try:
>             import sonic_platform
>             self.chassis = sonic_platform.platform.Platform().get_chassis()
>             self.eeprom = self.chassis.get_eeprom()
>         except Exception as e:
>             logger.log_warning("Failed to load platform-specific eeprom from sonic_platform package due to {}".format(repr(e)))
> 
>         # If we didn't successfully load the class from the sonic_platform package, try loading the old plugin
>         if not self.eeprom:
>             try:
>                 self.eeprom = self.load_platform_util(PLATFORM_SPECIFIC_MODULE_NAME, PLATFORM_SPECIFIC_CLASS_NAME)
>             except Exception as e:
>                 logger.log_error("Failed to load platform-specific eeprom implementation: %s" % (str(e)), True)
> 
>         if not self.eeprom:
>             sys.exit(ERR_EEPROMUTIL_LOAD)
> ```
> 
> Great to know psud should be able to use the same approach. I assumed xcvrd would require a different solution.



>         # First, try to load the new platform API
>         try:
>             import sonic_platform
>             self.chassis = sonic_platform.platform.Platform().get_chassis()
>             self.eeprom = self.chassis.get_eeprom()
>         except Exception as e:
>             logger.log_warning("Failed to load platform-specific eeprom from sonic_platform package due to {}".format(repr(e)))
> 
>         # If we didn't successfully load the class from the sonic_platform package, try loading the old plugin
>         if not self.eeprom:
>             try:
>                 self.eeprom = self.load_platform_util(PLATFORM_SPECIFIC_MODULE_NAME, PLATFORM_SPECIFIC_CLASS_NAME)
>             except Exception as e:
>                 logger.log_error("Failed to load platform-specific eeprom implementation: %s" % (str(e)), True)
> 
>         if not self.eeprom:
>             sys.exit(ERR_EEPROMUTIL_LOAD)



> Understood. So do you want to make any changes to this PR after all?

Sure. I'm testing the syseepromd.

stephenxs:
> Understood. So do you want to make any changes to this PR after all?

Hi Joe, syseepromd has been updated. Please review. Thank you.

PR #39, opened 2019-08-01
[xcvrd] Enhance xcvrd to handle new system level event/error

keboliu:
This PR has a dependency on PRs:
https://github.com/Azure/sonic-platform-common/pull/50
https://github.com/Azure/sonic-buildimage/pull/3261

In the above PRs introduced the new system-level event, this PR to enhance xcvrd to be more robust by handling the more new system-level event:
1.  wait for a certain period on  system_not_ready event to overcome race condition when xcvrd started but system not ready.

2. start to routing work on system_become_ready event, this event indicates that the system is ready for SFP change event report.

3. exit on system_fail event

### DIFF
--- a/sonic-xcvrd/scripts/xcvrd
+++ b/sonic-xcvrd/scripts/xcvrd
@@ -27,7 +27,6 @@ SYSLOG_IDENTIFIER = os.path.basename(__file__)
 
 PLATFORM_ROOT_PATH = '/usr/share/sonic/device'
 SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
-MINIGRAPH_PATH = '/etc/sonic/minigraph.xml'
 HWSKU_KEY = 'DEVICE_METADATA.localhost.hwsku'
 PLATFORM_KEY = 'DEVICE_METADATA.localhost.platform'
 
@@ -117,7 +116,7 @@ def get_platform_and_hwsku():
     proc.wait()
     platform = stdout.rstrip('\n')
 
-    proc = subprocess.Popen([SONIC_CFGGEN_PATH, '-m', MINIGRAPH_PATH, '-v', HWSKU_KEY],
+    proc = subprocess.Popen([SONIC_CFGGEN_PATH, '-d', '-v', HWSKU_KEY],
                             stdout=subprocess.PIPE,
                             shell=False,
                             stderr=subprocess.STDOUT)
@@ -233,6 +232,9 @@ def post_port_sfp_info_to_db(logical_port_name, table):
         ganged_port = True
 
     for physical_port in physical_port_list:
+        if not platform_sfputil.get_presence(physical_port):
+            continue
+
         port_name = get_physical_port_name(logical_port_name, ganged_member_num, ganged_port)
         ganged_member_num += 1
 
@@ -266,6 +268,9 @@ def post_port_dom_info_to_db(logical_port_name, table):
         ganged_port = True
 
     for physical_port in physical_port_list:
+        if not platform_sfputil.get_presence(physical_port):
+            continue
+
         port_name = get_physical_port_name(logical_port_name, ganged_member_num, ganged_port)
         ganged_member_num += 1
 
--- a/sonic-xcvrd/scripts/xcvrd
+++ b/sonic-xcvrd/scripts/xcvrd
@@ -1,0 +1,449 @@
+#!/usr/bin/env python2
+
+"""
+    Xcvrd
+    Transceiver information update daemon for SONiC
+"""
+
+try:
+    import getopt
+    import os
+    import imp
+    import signal
+    import subprocess
+    import sys
+    import syslog
+    import time
+    import threading
+    from swsscommon import swsscommon
+except ImportError, e:
+    raise ImportError (str(e) + " - required module not found")
+
+#============================= Constants =============================
+
+VERSION = '1.0'
+
+SYSLOG_IDENTIFIER = os.path.basename(__file__)
+
+PLATFORM_ROOT_PATH = '/usr/share/sonic/device'
+SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
+MINIGRAPH_PATH = '/etc/sonic/minigraph.xml'
+HWSKU_KEY = 'DEVICE_METADATA.localhost.hwsku'
+PLATFORM_KEY = 'DEVICE_METADATA.localhost.platform'
+
+PLATFORM_SPECIFIC_MODULE_NAME = "sfputil"
+PLATFORM_SPECIFIC_CLASS_NAME = "SfpUtil"
+
+# Global platform-specific sfputil class instance
+platform_sfputil = None
+
+# platform directory in base image
+PLATFORM_ROOT = "/usr/share/sonic/device"
+
+# platform root directory inside docker
+PLATFORM_ROOT_DOCKER = "/usr/share/sonic/platform"
+
+REDIS_HOSTNAME = "localhost"
+REDIS_PORT = 6379
+REDIS_TIMEOUT_MSECS = 0
+SELECT_TIMEOUT_MSECS = 1000
+
+DOM_INFO_UPDATE_PERIOD_SECS = 60
+TIME_FOR_SFP_READY_SECS = 1
+
+SFP_STATUS_INSERTED = '1'
+SFP_STATUS_REMOVED = '0'
+
+PHYSICAL_PORT_NOT_EXIST = -1
+SFP_EEPROM_NOT_READY = -2
+
+TEMP_UNIT = 'C'
+VOLT_UNIT = 'Volts'
+POWER_UNIT = 'dBm'
+BIAS_UNIT = 'mA'
+
+#========================== Syslog wrappers ==========================
+
+def log_info(msg, also_print_to_console=False):
+    syslog.openlog(SYSLOG_IDENTIFIER)
+    syslog.syslog(syslog.LOG_INFO, msg)
+    syslog.closelog()
+
+    if also_print_to_console:
+        print msg
+
+def log_warning(msg, also_print_to_console=False):
+    syslog.openlog(SYSLOG_IDENTIFIER)
+    syslog.syslog(syslog.LOG_WARNING, msg)
+    syslog.closelog()
+
+    if also_print_to_console:
+        print msg
+
+def log_error(msg, also_print_to_console=False):
+    syslog.openlog(SYSLOG_IDENTIFIER)
+    syslog.syslog(syslog.LOG_ERR, msg)
+    syslog.closelog()
+
+    if also_print_to_console:
+        print msg
+
+#========================== Signal Handling ==========================
+
+def signal_handler(sig, frame):
+    if sig == signal.SIGHUP:
+        log_info("Caught SIGHUP - ignoring...")
+        return
+    elif sig == signal.SIGINT:
+        log_info("Caught SIGINT - exiting...")
+        sys.exit(128 + sig)
+    elif sig == signal.SIGTERM:
+        log_info("Caught SIGTERM - exiting...")
+        sys.exit(128 + sig)
+    else:
+        log_warning("Caught unhandled signal '" + sig + "'")
+        return
+
+
+#============ Functions to load platform-specific classes ============
+
+# Returns platform and HW SKU
+def get_platform_and_hwsku():
+    proc = subprocess.Popen([SONIC_CFGGEN_PATH, '-H', '-v', PLATFORM_KEY],
+                            stdout=subprocess.PIPE,
+                            shell=False,
+                            stderr=subprocess.STDOUT)
+    stdout = proc.communicate()[0]
+    proc.wait()
+    platform = stdout.rstrip('\n')
+
+    proc = subprocess.Popen([SONIC_CFGGEN_PATH, '-m', MINIGRAPH_PATH, '-v', HWSKU_KEY],
+                            stdout=subprocess.PIPE,
+                            shell=False,
+                            stderr=subprocess.STDOUT)
+    stdout = proc.communicate()[0]
+    proc.wait()
+    hwsku = stdout.rstrip('\n')
+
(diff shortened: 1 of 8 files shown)


sonic-xcvrd/scripts/xcvrd
```diff
@@ -631,11 +675,16 @@ class sfp_state_update_task:
         while not stopping_event.is_set():
             status, port_dict = platform_sfputil.get_transceiver_change_event()
             if status:
+                # Receive event on single SFP
                 for key, value in port_dict.iteritems():
+                    if key == EVENT_ON_ALL_SFP:
+                        logger.log_error("Receive event with unexpected content, key: {}, value: {}".format(key, value))
```
qiluo-msft: >Receive [](start = 42, length = 7)

Past tense is better. As original 'Got'. Many time below.

sonic-xcvrd/scripts/xcvrd
```diff
@@ -647,17 +696,42 @@ class sfp_state_update_task:
...
-                # TODO: next step need to define more error types to handle accordingly.
-                logger.log_error("Failed to get transceiver change event. Exiting...")
-                os.kill(os.getppid(), signal.SIGTERM)
-                break
+                # Receive system level error/event
+                if bool(port_dict):
+                    for key, value in port_dict.iteritems():
+                        if key == EVENT_ON_ALL_SFP:
+                            if value == SYSTEM_NOT_READY:
+                                # system not ready, wait and retry
```
qiluo-msft: ># [](start = 32, length = 1)

wrong indentation.

sonic-xcvrd/scripts/xcvrd
```diff
@@ -647,17 +696,42 @@ class sfp_state_update_task:
...
+                                 if self._retry_for_system_ready():
+                                    continue
```
qiluo-msft: >    [](start = 33, length = 3)

inconsistent indentation

sonic-xcvrd/scripts/xcvrd
```diff
@@ -613,6 +621,42 @@ class sfp_state_update_task:
         self.task_process = None
         self.task_stopping_event = multiprocessing.Event()
 
+    def _retry_for_system_ready(self):
+        retry = 0
+        while retry < RETRY_TIMES_FOR_SYSTEM_READY:
+            #time.sleep(RETRY_PERIOD_FOR_SYSTEM_READY_SECS)
+            time_start  = time.time()
+            rc, event_dict = platform_sfputil.get_transceiver_change_event(RETRY_PERIOD_FOR_SYSTEM_READY_MSECS)
```
qiluo-msft: >platform_sfputil.get_transceiver_change_event [](start = 29, length = 45)

You can call platform_sfputil.get_transceiver_change_event function only once, and implement the logic as a normal state machine.

I don't mean your implementation is wrong, state machine may be much readable and future proof.

stephenxs: Hi Qi,
I admit that rewriting it in a state machine approach can make the code a little shorter by eliminating some duplicate code.
But I'm afraid it can also introduce negative effects: We have the logic to control retry timespan and times and introduce some local variables to control that. Rewriting it as a state machine will combine the retry control logic into the main procedure of xcvrd, which make the main function complicated.

sonic-xcvrd/scripts/xcvrd
stephenxs: fixed.

sonic-xcvrd/scripts/xcvrd
stephenxs: fixed.

sonic-xcvrd/scripts/xcvrd
stephenxs: fixed.

sonic-xcvrd/scripts/xcvrd
qiluo-msft: The benefit for state machine pattern is not shorter code, but a correct design and implementation.

For the design part, you need to consider any sequence of change events returned by vendor plugin code. For example, SYSTEM_NOT_READY followed by an inserted event. You need to draw the state machine on paper first to consider all the events at each state.

For the implementation part, you could design you state transition code in standalone function, and achieve 'control retry timespan and times' or 'local variables'.

---
In reply to: [310075723](https://github.com/Azure/sonic-platform-daemons/pull/39#discussion_r310075723) [](ancestors = 310075723)

stephenxs: Hi Qi,
I understand the advantage of state machine. I just feel that the logic of xcvrd daemon is relatively simple and it seems that using state machine will require a bit more effort.

stephenxs: I've also summarized the state machine which may reflect the logic. you can also check and compare.

states
- Initial state: INIT, before received system ready or a normal event
- Final state: EXIT
- other state: NORMAL, after has received system-ready or a normal event

events
- SYSTEM_NOT_READY
- SYSTEM_BECOME_READY
- NORMAL EVENT
  - sfp insertion/removal
  - timeout returned by sfputil.get_change_event with status = true
- SYSTEM_FAIL

State transmit:
1.	Current state: INIT
    - On receiving event SYSTEM_NOT_READY, handle by on_init_snr
        If retry < RETRY_TIMES_FOR_SYSTEM_READY : retry++
        else: reach max retry, exit
    - On receiving event SYSTEM_BECOME_READY, handle by on_init_sr
        Transmit state to NORMAL
    - On receiving event SYSTEM_FAIL, handle by on_anystate_sf
        Treat as a fatal error, exit
    - On receiving normal event like SFP modules insertion/removal (for the vendors who don’t implement SYSTEM_BECOME_READY), handle by on_anystate_normal
      Transmit state to NORMAL
      Handle the plugin/plugout event normally.
2.	Current state: NORMAL
    - On receiving normal event, handled by on_anystate_normal
        Handle the event normally.
        Don’t transmit state.
    - On receiving SYSTEM_BECOME_READY
        Just log the event and ignored
    - On receiving SYSTEM_NOT_READY, handle by on_normal_snr
        Treat as a fatal error, exit
    - On receiving SYSTEM_FAIL, handle by on_anystate_sf
        Treat as a fatal error, exit

State           |event               next state
INIT            |SYSTEM NOT READY    |INIT / EXIT
INIT            |SYSTEM READY        |NORMAL
ANYSTATE        |SYSTEM FAIL         |EXIT
ANYSTATE        |NORMAL EVENT        |NORMAL
NORMAL          |SYSTEM NOT READY    |EXIT
EXIT            -

stephenxs:
> As comments
> Also blocked on [Azure/sonic-platform-common#50](https://github.com/Azure/sonic-platform-common/pull/50)

code updated with state machine approach

sonic-xcvrd/scripts/xcvrd
```diff
@@ -628,36 +666,163 @@ class sfp_state_update_task:
...
-            else:
-                # If get_transceiver_change_event() return error, will clean up the DB and then exit
-                # TODO: next step need to define more error types to handle accordingly.
-                logger.log_error("Failed to get transceiver change event. Exiting...")
+            next_state = None
+            time_start = time.time()
+            status, port_dict = platform_sfputil.get_transceiver_change_event(timeout)
+            logger.log_debug("Got event {} {} in state {}".format(status, port_dict, state))
+            event = self._mapping_event_from_change_event(status, port_dict)
+            for key, value in port_dict.iteritems():
```
qiluo-msft: >for key, value in port_dict.iteritems(): [](start = 12, length = 40)

Why you put the for-loop outside the event/state branches? If you have an empty port_dict, you just ignore the event handling.

sonic-xcvrd/scripts/xcvrd
```diff
@@ -43,13 +43,27 @@ XCVRD_MAIN_THREAD_SLEEP_SECS = 60
...
 SFP_STATUS_REMOVED = '0'
 
+EVENT_ON_ALL_SFP = '-1'
+# events definition
+SYSTEM_NOT_READY = 'system_not_ready'
+SYSTEM_BECOME_READY = 'system_become_ready'
+SYSTEM_FAIL = 'system_fail'
+NORMAL_EVENT = 'normal'
+# states definition
+STATE_INIT = 'INIT STATE'
```
qiluo-msft: >'INIT STATE' [](start = 13, length = 12)

You can use integer as value, don't rely on python to optimize the string comparison.

sonic-xcvrd/scripts/xcvrd
```diff
@@ -628,36 +666,163 @@ class sfp_state_update_task:
...
-                            logger.log_info("Got SFP removed event")
-                            del_port_sfp_dom_info_from_db(logical_port, int_tbl, dom_tbl)
-                        else:
-                            # TODO, SFP return error code, need handle accordingly.
-                            continue
-            else:
-                # If get_transceiver_change_event() return error, will clean up the DB and then exit
-                # TODO: next step need to define more error types to handle accordingly.
-                logger.log_error("Failed to get transceiver change event. Exiting...")
+            next_state = None
```
qiluo-msft: >next_state [](start = 12, length = 10)

You may initialize it with state, next_state will never be None.

sonic-xcvrd/scripts/xcvrd
```diff
@@ -628,36 +666,163 @@ class sfp_state_update_task:
...
+                elif event == SYSTEM_FAIL:
+                    # no matter which state current it is, it's fatal
+                    next_state = STATE_EXIT
+                    logger.log_error("Got system_fail event on state {}, exiting".format(state))
+                else:
+                    logger.log_warning("Got unknown event {} on state {}.".format(event, state))
+
+            if unexpected_state:
+                next_state = STATE_EXIT
+                logger.log_error("Facing unexpected state {}, exiting...".format(state))
```
qiluo-msft: You can remove unexpected_state. This is redundant logic.

sonic-xcvrd/scripts/xcvrd
```diff
@@ -628,36 +666,163 @@ class sfp_state_update_task:
...
+                elif event == SYSTEM_BECOME_READY:
+                    if state == STATE_INIT:
+                        next_state = STATE_NORMAL
+                        logger.log_info("Got system_become_ready in init state, transmit to normal state")
+                    elif state == STATE_NORMAL:
+                        logger.log_info("Got system_become_ready in normal state, ignored")
+                    else:
+                        unexpected_state = True
+                        next_state = STATE_EXIT
+                        break
```
qiluo-msft: repeated code, you  can extract function.

sonic-xcvrd/scripts/xcvrd
```diff
@@ -628,36 +666,163 @@ class sfp_state_update_task:
...
+
+            if next_state is not None:
+                logger.log_debug("State transmitted from {} to {}".format(state, next_state))
+                state = next_state
+
+            if next_state == STATE_EXIT:
                 os.kill(os.getppid(), signal.SIGTERM)
                 break
+            elif next_state == STATE_NORMAL:
+                timeout = 0
```
qiluo-msft: You can move this part into event handling branches.

sonic-xcvrd/scripts/xcvrd
stephenxs: This is only for log. 
By setting unexpected_state to true we can know why the state machine reach STATE_EXIT.

sonic-xcvrd/scripts/xcvrd
stephenxs: Yes. you are correct. only for NORMAL_EVENT is this for loop required. will update.

sonic-xcvrd/scripts/xcvrd
stephenxs: I think to use string is better since it makes logging information more human-readable.
Even though a dictionary can be adopted to map from integer value to string, the case that the state is not in the key set of the dictionary has to be handled. That's too complicated.
BTW, I'm not sure whether to use string address to compare the string is part of python's language specification or just optimization that depends on python interpreter's implementation?

sonic-xcvrd/scripts/xcvrd
stephenxs: > This is only for log.
> By setting unexpected_state to true we can know why the state machine reach STATE_EXIT.

I decide to remove this. It is very unlikely to enter an unexpected state since all states are handled for each event.

sonic-xcvrd/scripts/xcvrd
stephenxs: It's better to set it here since it can ensure that wherever the state reaches STATE_NORMAL the timeout can be properly set.

sonic-xcvrd/scripts/xcvrd
stephenxs: states have been updated to integer value. 
The events, like SYSTEM_FAIL, are already used by the plugins. In this sence, we'd better not to change them. It seems that to use string has already been a common practice in plugins, like the normal plugin/out events which are represented by '1', '0' respectively.

sonic-xcvrd/scripts/xcvrd
stephenxs: now, only next_state is set when facing unexpected state. in this sense, no need to extract function.

sonic-xcvrd/scripts/xcvrd
```diff
@@ -672,36 +710,151 @@ class sfp_state_update_task:
...
+                # no matter which state current it is, it's fatal
+                next_state = STATE_EXIT
+                logger.log_error("Got system_fail event on state {}, exiting".format(state))
             else:
-                # If get_transceiver_change_event() return error, will clean up the DB and then exit
-                # TODO: next step need to define more error types to handle accordingly.
-                logger.log_error("Failed to get transceiver change event. Exiting...")
+                logger.log_warning("Got unknown event {} on state {}.".format(event, state))
+
+            if not next_state == state:
```
qiluo-msft: >if not next_state == state: [](start = 12, length = 27)

For readability, use `if next_state != state`

stephenxs: fixed.

PR #40, opened 2019-08-07
[xcvrd] Fix transceiver tuning issue

dgsudharsan:
**- What I did**
Fixing transceiver tuning issue where number of lanes equal to number of logical ports in a port. (Example 40G port with 4 lanes is in 4x10G mode)
**- How I did it**

**- How to verify it**
Define a profile where number of lanes equal to number of logical ports and check if pre-emphasis gets programmed

Handling transceiver tuning scenario when number of lanes in a port is equal to number of logical ports created
[preemp_ut.txt](https://github.com/Azure/sonic-platform-daemons/files/3478202/preemp_ut.txt)

### DIFF
--- a/sonic-xcvrd/scripts/xcvrd
+++ b/sonic-xcvrd/scripts/xcvrd
@@ -400,7 +400,7 @@ def get_media_val_str_from_dict(media_dict):
             media_str += lane_separator
     return media_str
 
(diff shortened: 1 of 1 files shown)


dgsudharsan:
Need to port this fix to April branch

PR #45, opened 2019-09-04
[201811 xcvrd] state machine enhancement

keboliu:
backport PR https://github.com/Azure/sonic-platform-daemons/pull/44 from master to 201811 branch

Enhance the state machine in order to overcome a possible platform temporarily fail/unavailable case
    
   1. When receiving system_fail event under NORMAL state, it will transmit to INIT instead of exit directly
   2. In INIT state will handle system_fail event as the same as system_not_ready event, try certain times,
       if the system recovered then transmit to NORMAL state again, if not recovered in a certain period, then exit.
    
 the benefit of this change is that when the system has error/failed temporarily,  xcvrd can survive and recover instead of exit directly, make to more tolerance to errors.

### DIFF
--- a/sonic-xcvrd/scripts/xcvrd
+++ b/sonic-xcvrd/scripts/xcvrd
@@ -472,12 +472,14 @@ def main():
     # Start main loop to listen to the SFP change event.
     # The state migrating sequence:
     # 1. When the system starts, it is in "INIT" state, calling get_transceiver_change_event
-    #    with RETRY_PERIOD_FOR_SYSTEM_READY_MSECS as timeout for as many as RETRY_TIMES_FOR_SYSTEM_READY
-    #    times
+    #    with RETRY_PERIOD_FOR_SYSTEM_READY_MSECS as timeout for before reach RETRY_TIMES_FOR_SYSTEM_READY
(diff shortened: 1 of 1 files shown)


sonic-xcvrd/scripts/xcvrd
```diff
@@ -472,12 +482,14 @@ def main():
     # Start main loop to listen to the SFP change event.
     # The state migrating sequence:
     # 1. When the system starts, it is in "INIT" state, calling get_transceiver_change_event
-    #    with RETRY_PERIOD_FOR_SYSTEM_READY_MSECS as timeout for as many as RETRY_TIMES_FOR_SYSTEM_READY
-    #    times
+    #    with RETRY_PERIOD_FOR_SYSTEM_READY_MSECS as timeout for before reach RETRY_TIMES_FOR_SYSTEM_READY
+    #    times, otherwise it will transmit to "EXIT" state
```
jleveque: I believe the word you're looking for is "transition", not "transmit". Can you please change this throughout the file?

keboliu: thanks for the suggestion, fixed.

keboliu:
@jleveque comments fixed, would you please have a look?

PR #50, opened 2020-01-19
Fix missing DOM threshold info from DB

keboliu:
DOM threshold info is missing from DB because of function to fetch DOM threshold info "post_port_dom_threshold_info_to_db" is not called in "dom_info_update_task", and which will break CLI "show interface transceiver eeprom -d".
The fix is to add call to "post_port_dom_threshold_info_to_db" in "dom_info_update_task".

This fix should be included in the 201911 and bmtor branch.


liat-grozovik:
@jleveque can you please review? please not this should be taken for 201911 and bmtor branches.

jleveque:
@prsunny: FYI, this commit also needs to make its way into the bmtor branch.

PR #54, opened 2020-04-17
[thermalctld] add FAN led management in thermal control daemon

Junchao-Mellanox:
What I did:
1. Set led to red and trigger syslog when fan is broken
2. Add fan drawer information to redis database
3. Trigger a syslog when any fan is absence or broken to alert user that "Insufficient fan number"

How I verify:
Manual test and run platform regression

### DIFF
--- a/sonic-thermalctld/tests/test_thermalctld.py
+++ b/sonic-thermalctld/tests/test_thermalctld.py
@@ -168,9 +168,9 @@ def test_insufficient_fan_number():
     fan_list = chassis.get_all_fans()
     fan_list[0].presence = True
     fan_updater.update()
-    logger.log_warning.assert_called_with('Insufficient number of working fans warning: 1 fans are not working.')
     assert logger.log_notice.call_count == 1
-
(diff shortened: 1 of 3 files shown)


sonic-thermalctld/tests/test_thermalctld.py
```diff
@@ -133,6 +148,35 @@ def test_fanupdater_fan_over_speed():
...
+    fan_updater = FanUpdater(chassis)
+    fan_updater.update()
+    assert logger.log_warning.call_count == 3
+    logger.log_warning.assert_called_with('Insufficient number of working fans warning: 2 fans are not working.')
+
+    fan_list = chassis.get_all_fans()
+    fan_list[0].presence = True
+    fan_updater.update()
+    logger.log_warning.assert_called_with('Insufficient number of working fans warning: 1 fans are not working.')
+    assert logger.log_notice.call_count == 1
```
jleveque, line 172: While there is no functional difference, I suggest swapping lines 171 and 172 to make the order of operations consistent with the stanzas above and below.

Junchao-Mellanox: Fixed

PR #56, opened 2020-04-24
[syseepromd] Prevent the syseepromd from termination

kuanyu99:
Why I did:

- Since the termination of the syseepromd will lead to the termination of the pmon.

What I did:

- Regularly try to update the eeprom data, instead of exiting the syseepromd directly when the update_eeprom_to_db return with failure.


kuanyu99:
@jleveque Could you help to review this, please?

PR #57, opened 2020-05-06
[MultiDB] should use multiDB API

dzhangalibaba:
* Should use new multiDB Connector in daemon base
* This submodule needs updated with daemon_base changes together which is in sonic_buildinage
 

Signed-off-by: Dong Zhang d.zhang@alibaba-inc.com

### DIFF
--- a/sonic-ledd/scripts/ledd
+++ b/sonic-ledd/scripts/ledd
@@ -76,7 +76,7 @@ class DaemonLedd(DaemonBase):
             sys.exit(LEDUTIL_LOAD_ERROR)
 
         # Open a handle to the Application database
-        appl_db = daemon_base.db_connect(swsscommon.APPL_DB)
+        appl_db = daemon_base.db_connect("APPL_DB")
 
         # Subscribe to PORT table notifications in the Application DB
         sel = swsscommon.Select()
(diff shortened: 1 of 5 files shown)


dzhangalibaba:
This PR and https://github.com/Azure/sonic-buildimage/pull/4549 depends on each other.
the submodule needs updated with the merge of this PR together.

dzhangalibaba:
@qiluo-msft

sonic-ledd/scripts/ledd
```diff
@@ -76,7 +76,7 @@ class DaemonLedd(DaemonBase):
             sys.exit(LEDUTIL_LOAD_ERROR)
 
         # Open a handle to the Application database
-        appl_db = daemon_base.db_connect(swsscommon.APPL_DB)
```
lguohan, line 79: why not use swsscommon.APPL_DB?

dzhangalibaba: swsscommon.APPL_DB is db id from swsscommon c++ library. Here we need a string "APPL_DB", For backward compatibility, we leave db id still there and not changed APPL_DB = "APPL_DB"

dzhangalibaba:
@qiluo-msft  can we merge this first? then I'll update another PR "[MultiDB] daemon base should use multiDB DBConnector #4549"  with this submodule update as well to avoid the dependency.

PR #60, opened 2020-06-14
Initial version of pcied

sujinmkang:
pcied will run in pmon container to monitor the pcie device status during run time.
It will update the state db with current pcie status so other monitoring service can read to alert any failure status.

### DIFF
--- a/sonic-pcied/scripts/pcied
+++ b/sonic-pcied/scripts/pcied
@@ -7,13 +7,14 @@
 
 try:
     import os
-    import sys
     import signal
-    import threading
     import subprocess
+    import sys
+    import threading
+
+    import swsssdk
     from sonic_daemon_base.daemon_base import Logger
     from sonic_daemon_base.daemon_base import DaemonBase
-    import swsssdk
 except ImportError, e:
     raise ImportError(str(e) + " - required module not found")
 
--- a/sonic-pcied/scripts/pcied
+++ b/sonic-pcied/scripts/pcied
@@ -60,7 +60,7 @@ class DaemonPcied(DaemonBase):
 
     def check_pcie_devices(self):
         tmp_pcie_status_f = "/tmp/pcie_status_f"
-        os.system('sudo pcieutil pcie-check | grep "%s" > %s' % (PCIE_RESULT_REGIX, tmp_pcie_status_f))
+        os.system('sudo pcieutil pcie-check | grep "%s" > %s' % (PCIE_RESULT_REGEX, tmp_pcie_status_f))
         f = open(tmp_pcie_status_f, "r")
         pcie_status = f.readline()
         pcie_db_state = self.read_state_db("PCIE_STATUS|", "PCIE_DEVICES")
--- a/sonic-pcied/scripts/pcied
+++ b/sonic-pcied/scripts/pcied
@@ -6,9 +6,11 @@
 """
 
 try:
+    import os
     import sys
     import signal
     import threading
+    import subprocess
     from sonic_daemon_base.daemon_base import Logger
     from sonic_daemon_base.daemon_base import DaemonBase
 except ImportError, e:
@@ -20,17 +22,7 @@ except ImportError, e:
 
 SYSLOG_IDENTIFIER = "pcied"
 
-PLATFORM_SPECIFIC_MODULE_NAME = "pcieutil"
-PLATFORM_SPECIFIC_CLASS_NAME = "PcieUtil"
-
 PCIED_MAIN_THREAD_SLEEP_SECS = 60
-
-PCIEUTIL_LOAD_ERROR = 1
-
-# Global platform specific pcieutil class instance
-platform_pcieutil = None
-# Global chassis object based on new platform api
-platform_chassis = None
 
 # Global logger class instance
 logger = Logger(SYSLOG_IDENTIFIER)
@@ -39,14 +31,19 @@ logger = Logger(SYSLOG_IDENTIFIER)
 # Helper functions =============================================================
 #
 def check_pcie_devices():
-    resultInfo = platform_pcieutil.get_pcie_check()
-    for item in resultInfo:
-        if item["result"] != "Passed":
-            logger.log_warning("PCIe Device: " +  item["name"] + " Not Found")
-            err+=1
-    if err:
-        logger.log_error("PCIe Device Checking All Test ----------->>> FAILED")
-
+    cmd = [ 'sudo', 'pcieutil', 'pcie-check' ]
+    tmp_pcie_status_f="/tmp/pcie_status_f"
+    p = subprocess.Popen(cmd, stdout=subprocess.PIPE,
+                              stderr=subprocess.PIPE)
+    resultInfo, err = p.communicate()
+    if not err:
+        os.system('redis-cli -n 6 GET "PCIE_STATE|system" > %s'%tmp_pcie_status_f)
+        f=open(tmp_pcie_status_f, "r")
+        pcie_status = f.readline()
+        if "1" in pcie_status:
+            logger.log_info("PCIe device status check : PASSED")
+        else:
+            logger.log_info("PCIe device status check : FAILED")
 
 #
 # Daemon =======================================================================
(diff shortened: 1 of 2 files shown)


sonic-pcied/scripts/pcied
```diff
@@ -0,0 +1,156 @@
...
+REDIS_HOSTIP = "127.0.0.1"
+
+# Global logger class instance
+logger = Logger(SYSLOG_IDENTIFIER)
+
+#
+# Helper Function ==============================================================
+#
+
+def get_platform_and_hwsku():
```
jleveque, line 44: This is a common function. Can you please import from the UtilHelper class?

https://github.com/Azure/sonic-utilities/blob/master/utilities_common/util_base.py#L73

sonic-pcied/scripts/pcied
```diff
@@ -0,0 +1,156 @@
...
+
+    def check_pcie_devices(self):
+        platform, hwsku = get_platform_and_hwsku()
+        pciefilePath = "/".join([PLATFORM_ROOT_PATH, platform, "plugins", PCIE_CONF_FILE])
+        sys.path.append(os.path.abspath(pciefilePath))
+        if not os.path.exists(pciefilePath):
+            logger.log_error("Platform pcie configuration file doesn't exist! exit pcied")
+            sys.exit("Platform PCIe Configuration file doesn't exist!")
+
+        tmp_pcie_status_f="/tmp/pcie_status_f"
```
jleveque, line 89: Please add spaces around `=`. There are other instances of this throughout the file.

Tip: Please pass all Python code through http://pep8online.com/ and fix style errors accordingly (don't worry about the "line too long" errors, unless the line is over ~120 characters).

sonic-pcied/scripts/pcied
```diff
@@ -0,0 +1,156 @@
...
+        os.system('sudo pcieutil pcie-check | grep "%s" > %s'%(PCIE_RESULT_REGIX, tmp_pcie_status_f))
+        f=open(tmp_pcie_status_f, "r")
```
paavaanan, line 91: Instead of using /tmp/ file to query results. We should use STATE_DB as a place-holder.

sonic-pcied/scripts/pcied
```diff
@@ -0,0 +1,156 @@
...
+
+        self.timeout = PCIED_MAIN_THREAD_SLEEP_SECS
+        self.stop_event = threading.Event()
+
+        self.state_db = swsssdk.SonicV2Connector(host=REDIS_HOSTIP)
+        self.state_db.connect("STATE_DB")
+
+    def check_pcie_devices(self):
+        platform, hwsku = get_platform_and_hwsku()
+        pciefilePath = "/".join([PLATFORM_ROOT_PATH, platform, "plugins", PCIE_CONF_FILE])
```
keboliu, line 84: is PCIE_CONF_FILE a plugin or a config file? If it's not a plugin I would suggest putting it under platform folder.

sonic-pcied/scripts/pcied
```diff
@@ -0,0 +1,156 @@
...
+        sys.path.append(os.path.abspath(pciefilePath))
+        if not os.path.exists(pciefilePath):
+            logger.log_error("Platform pcie configuration file doesn't exist! exit pcied")
+            sys.exit("Platform PCIe Configuration file doesn't exist!")
```
keboliu, line 88: I would suggest moving this check outside this function and do it before entering the loop, or it will periodically print out error msg on the platform which doesn't have this supported?

sonic-pcied/scripts/pcied
```diff
@@ -0,0 +1,156 @@
...
+
+        tmp_pcie_status_f = "/tmp/pcie_status_f"
+        os.system('sudo pcieutil pcie-check | grep "%s" > %s' % (PCIE_RESULT_REGIX, tmp_pcie_status_f))
+        f = open(tmp_pcie_status_f, "r")
+        pcie_status = f.readline()
```
keboliu, line 93: Is it possible to use subprocess.popen and parsing the command output directly?

sonic-pcied/scripts/pcied
```diff
@@ -0,0 +1,156 @@
...
+
+# Global logger class instance
+logger = Logger(SYSLOG_IDENTIFIER)
+
+#
+# Helper Function ==============================================================
+#
+
+
+def get_platform_and_hwsku():
```
keboliu, line 45: this function already implemented here: https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-daemon-base/sonic_daemon_base/daemon_base.py#L111

sonic-pcied/scripts/pcied
```diff
@@ -0,0 +1,156 @@
...
+            logger.log_error("Platform pcie configuration file doesn't exist! exit pcied")
+            sys.exit("Platform PCIe Configuration file doesn't exist!")
+
+        tmp_pcie_status_f = "/tmp/pcie_status_f"
+        os.system('sudo pcieutil pcie-check | grep "%s" > %s' % (PCIE_RESULT_REGIX, tmp_pcie_status_f))
+        f = open(tmp_pcie_status_f, "r")
+        pcie_status = f.readline()
+        if "PASSED" in pcie_status:
+            self.update_state_db("PCIE_STATUS|", "PCIE_DEVICES", "PASSED")
+            logger.log_info("PCIe device status check : PASSED")
```
keboliu, line 96: do we want it to be printed out each 60s?

sujinmkang: I thought it's ok to indicate the pcie status every 60 seconds. Do you have any other suggestion? I was also thinking about logging only when its status changes.

sonic-pcied/scripts/pcied
sujinmkang: Good point, I'm planing to work on additional feature on sonic-pcie again. I will fix this with that.

sonic-pcied/scripts/pcied
sujinmkang: I tried but the output of pcieutil wasn't return as command output

sonic-pcied/scripts/pcied
keboliu: IMO logging only on when status change is better.

sonic-pcied/scripts/pcied
```diff
@@ -0,0 +1,132 @@
...
+    import swsssdk
+except ImportError, e:
+    raise ImportError(str(e) + " - required module not found")
+
+#
+# Constants ====================================================================
+#
+SYSLOG_IDENTIFIER = "pcied"
+
+PCIE_RESULT_REGIX = "PCIe Device Checking All Test"
```
jleveque, line 24: s/PCIE_RESULT_REGIX/PCIE_RESULT_REGEX/

jleveque: You also need to update it on line 63 where it is used.

sonic-pcied/scripts/pcied
```diff
@@ -0,0 +1,132 @@
...
+
+        self.timeout = PCIED_MAIN_THREAD_SLEEP_SECS
+        self.stop_event = threading.Event()
+
+        self.state_db = swsssdk.SonicV2Connector(host=REDIS_HOSTIP)
+        self.state_db.connect("STATE_DB")
+
+    def check_pcie_devices(self):
+        tmp_pcie_status_f = "/tmp/pcie_status_f"
+        os.system('sudo pcieutil pcie-check | grep "%s" > %s' % (PCIE_RESULT_REGIX, tmp_pcie_status_f))
```
jleveque, line 63: s/PCIE_RESULT_REGIX/PCIE_RESULT_REGEX/ here as well.

sonic-pcied/scripts/pcied
jleveque: That's strange. Did you check both stdout and stderr?

sonic-pcied/scripts/pcied
```diff
@@ -10,6 +10,7 @@ try:
     import sys
     import signal
     import threading
+    import subprocess
```
jleveque, line 13: Please add this import in alphabetical order.

sonic-pcied/scripts/pcied
```diff
@@ -0,0 +1,136 @@
...
+            sys.exit("Platform PCIe Configuration file doesn't exist!")
+
+        self.timeout = PCIED_MAIN_THREAD_SLEEP_SECS
+        self.stop_event = threading.Event()
+
+        self.state_db = swsssdk.SonicV2Connector(host=REDIS_HOSTIP)
+        self.state_db.connect("STATE_DB")
+
+    def check_pcie_devices(self):
+        cmd = [ 'sudo', 'pcieutil', 'pcie-check' ]
```
bratashX, line 64: pcieutil is absent in pmon container. Please view on these links:  https://github.com/Azure/sonic-buildimage/issues/5236

PR #61, opened 2020-06-23
[psud] Store PSU temperature and voltage information to database

Junchao-Mellanox:
Why I did this?
 
System health feature requires PSU temperature and voltage information, need store them to database for host side use.

How I did?

Get temperature and voltage information from platform API and store them to database

How I verify?

Manual test and check redis database.

### DIFF
--- a/sonic-psud/scripts/psud
+++ b/sonic-psud/scripts/psud
@@ -38,6 +38,11 @@ PSU_INFO_TABLE = 'PSU_INFO'
 PSU_INFO_KEY_TEMPLATE = 'PSU {}'
 PSU_INFO_PRESENCE_FIELD = 'presence'
 PSU_INFO_STATUS_FIELD = 'status'
+PSU_INFO_TEMP_FIELD = 'temp'
+PSU_INFO_TEMP_TH_FIELD = 'temp_th'
+PSU_INFO_VOLTAGE_FIELD = 'voltage'
+PSU_INFO_VOLTAGE_MAX_TH_FIELD = 'voltage_max_th'
+PSU_INFO_VOLTAGE_MIN_TH_FIELD = 'voltage_min_th'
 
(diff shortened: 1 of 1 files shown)


sonic-psud/scripts/psud
```diff
@@ -38,6 +38,11 @@ PSU_INFO_TABLE = 'PSU_INFO'
 PSU_INFO_KEY_TEMPLATE = 'PSU {}'
 PSU_INFO_PRESENCE_FIELD = 'presence'
 PSU_INFO_STATUS_FIELD = 'status'
+PSU_INFO_TEMP_FIELD = 'temp'
+PSU_INFO_TEMP_TH_FIELD = 'temp_th'
+PSU_INFO_VOLTAGE_FIELD = 'voltage'
+PSU_INFO_VOLTAGE_MAX_TH_FIELD = 'voltage_max_th'
+PSU_INFO_VOLTAGE_MIN_TH_FIELD = 'voltage_min_th'
```
jleveque, line 45: Wondering if we should use the full word "threshold" in the threshold-related DB keys for clarity?

Junchao-Mellanox: Fixed

jleveque:
@Junchao-Mellanox: Can we backport this to 201911?

Junchao-Mellanox:
@jleveque for Mellanox platform, we have PSU temperature information in TEMPERATURE_INFO table of state DB. But I am not sure if there is such information for other vendors. So we add some entries to PSU_INFO table.

jleveque:
@Junchao-Mellanox: Thanks for the info. While I would like to see this ported back to 201911, it would involve backporting all the temperature and threshould APIs to the 201911 branch in sonic-platform-common. That would be a lot of work, and we would need to make sure we don't break anything. For now, I say we don't backport this to 201911.

PR #67, opened 2020-07-14
[xcvrd] Return non-zero error code on SFP error

vdahiya12:
Currently when there is an error event in xcvrd within SFP state machine, it returns with 0 exit code. With this change the SFP if it has an error event with return with non-zero exit code.

### DIFF
--- a/sonic-xcvrd/scripts/xcvrd
+++ b/sonic-xcvrd/scripts/xcvrd
@@ -1162,7 +1162,6 @@ class DaemonXcvrd(DaemonBase):
         logger.log_info("Shutting down...")
 
         if self.sfp_error_event.is_set():
-            logger.log_error("SFP error event occured in SFP monitoring loop")
             sys.exit(SFP_SYSTEM_ERROR)
 
 #
--- a/sonic-xcvrd/scripts/xcvrd
+++ b/sonic-xcvrd/scripts/xcvrd
@@ -986,7 +986,7 @@ class sfp_state_update_task:
         if self.task_stopping_event.is_set():
             return
 
-        self.task_process = multiprocessing.Process(target=self.task_worker,args=(self.task_stopping_event,sfp_error_event))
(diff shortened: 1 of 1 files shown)


sonic-xcvrd/scripts/xcvrd
```diff
@@ -1157,6 +1161,10 @@ class DaemonXcvrd(DaemonBase):
 
         logger.log_info("Shutting down...")
 
+        if self.sfp_error_event.is_set():
+            logger.log_error("SFP error event occured in SFP monitoring loop")
```
jleveque, line 1165: I think this log message is unnecessary, since error messages are already being logged at each point that `sfp_error_event.set()`is being called. Suggest to remove.

vdahiya12: fixed

sonic-xcvrd/scripts/xcvrd
```diff
@@ -979,11 +982,11 @@ class sfp_state_update_task:
 
         logger.log_info("Stop SFP monitoring loop")
 
-    def task_run(self):
+    def task_run(self, sfp_error_event):
         if self.task_stopping_event.is_set():
             return
 
-        self.task_process = multiprocessing.Process(target=self.task_worker,args=(self.task_stopping_event,))
+        self.task_process = multiprocessing.Process(target=self.task_worker,args=(self.task_stopping_event,sfp_error_event))
```
jleveque, line 989: Add a space between `,` and `sfp_error_event`

vdahiya12: fixed

jleveque:
This will not cherry-pick cleanly to the 201811 branch, as the code has diverged quite a bit. Will need a separate PR for 201811.

jleveque:
201811 branch PR here: https://github.com/Azure/sonic-platform-daemons/pull/68

PR #74, opened 2020-07-27
Migrate from sonic-daemon-base package to sonic-py-common package

jleveque:
As part of consolidating all common Python-based functionality into the new sonic-py-common package, this pull request migrates from importing the sonic-daemon-base package to importing the sonic-py-common package. This is the next step toward resolving https://github.com/Azure/sonic-buildimage/issues/4999.

- Also reorganize imports for consistency

### DIFF
--- a/sonic-ledd/scripts/ledd
+++ b/sonic-ledd/scripts/ledd
@@ -10,7 +10,6 @@ try:
     import sys
 
     from sonic_py_common import daemon_base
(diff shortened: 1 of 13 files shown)


jleveque:
Removed "Request for 201911 Branch" label, as backporting this change and the related changes will most likely be better performed by creating new PRs against the 201911 branch.

PR #94, opened 2020-09-18
[thermalctld] Fix issue: thermalctld should be auto restarted when being killed

Junchao-Mellanox:
**- Why I did it**

Part of thermalctld function is to handle user space thermal policies for events like fan/PSU removing, it works together with kernel thermal algorithm to make sure the switch won't be overheat.

Recently, we found that commit https://github.com/Azure/sonic-buildimage/commit/cbc75fe4c85fd01aded0bacb283c8bbf62adf2f8 changes its autorestart configuration in supervisord, and it won't be auto restarted after being killed. This PR is to make sure that thermalctld will be always restarted when it is killed.

**- How I did it**

1. Catch signal and assign it to a exit_code member data
2. Add a line "exit(self.exit_code)" to make sure it won't always exit with code 0

**- How to verify it**

Manual test

**- Which release branch to backport (provide reason below if selected)**

Depends on where https://github.com/Azure/sonic-buildimage/commit/cbc75fe4c85fd01aded0bacb283c8bbf62adf2f8 is going to merge to. 

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -632,7 +632,7 @@ class ThermalControlDaemon(daemon_base.DaemonBase):
             self.log_info("Caught SIGHUP - ignoring...")
         elif sig == signal.SIGINT or sig == signal.SIGTERM:
             self.log_info("Caught signal {} - exiting...".format(sig))
-            self.exit_code = sig
+            self.exit_code = sig + 128
             self.stop_event.set()
         else:
             self.log_warning("Caught unhandled signal '" + sig + "'")
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -618,7 +618,11 @@ class ThermalControlDaemon(daemon_base.DaemonBase):
         """
         super(ThermalControlDaemon, self).__init__(log_identifier)
         self.stop_event = threading.Event()
-        self.exit_code = -1
+
+        # Thermal control daemon is designed to never exit, it must always
+        # return non zero exit code when exiting and so that supervisord will 
+        # restart it automatically.
+        self.exit_code = 1
 
     # Signal handler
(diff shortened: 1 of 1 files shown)


sonic-thermalctld/scripts/thermalctld
```diff
@@ -629,11 +630,9 @@ class ThermalControlDaemon(daemon_base.DaemonBase):
...
         if sig == signal.SIGHUP:
             self.log_info("Caught SIGHUP - ignoring...")
-        elif sig == signal.SIGINT:
-            self.log_info("Caught SIGINT - exiting...")
-            self.stop_event.set()
-        elif sig == signal.SIGTERM:
-            self.log_info("Caught SIGTERM - exiting...")
+        elif sig == signal.SIGINT or sig == signal.SIGTERM:
+            self.log_info("Caught signal {} - exiting...".format(sig))
+            self.exit_code = sig
```
jleveque, line 635: Suggest changing exit code to `(128 + sig)`, as this is a standard convention in Linux. This was it will also remain consistent with the default signal handler in DaemonBase: https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-py-common/sonic_py_common/daemon_base.py#L53

sonic-thermalctld/scripts/thermalctld
```diff
@@ -618,6 +618,7 @@ class ThermalControlDaemon(daemon_base.DaemonBase):
         """
         super(ThermalControlDaemon, self).__init__(log_identifier)
         self.stop_event = threading.Event()
+        self.exit_code = -1
```
jleveque, line 621: Exit codes in Linux are non-negative values. `-1` will get converted to `255`. I suggest using a positive value here between 1 and 127 so that it's easy to search for in the code.

jleveque (changes requested):

jleveque (approved):

PR #97, opened 2020-10-01
Introduce chassisd to monitor status of cards on chassis

mprabhu-nokia:
Introducing chassisd to monitor status of cards on a modular chassis

HLD: https://github.com/Azure/SONiC/pull/646

**-What I did**
Introducing a new process to monitor status of control, line and fabric cards.

**-How I did it**
Support of monitoring of line-cards and fabric-cards. This runs in the main thread periodically.
It updates the STATE_DB with the status information. 'show platform chassis-modules' will read from the STATE_DB

Support of handling configuration of moving the cards toadministratively up/down state. The handling happens as part
of a separate thread that waits on select() for config event from a CHASSIS_MODULE table in CONFIG_DB. 

**-How I verified it**

### DIFF
--- a/sonic-chassisd/pytest.ini
+++ b/sonic-chassisd/pytest.ini
@@ -1,0 +1,3 @@
+[pytest]
+filterwarnings =
+    ignore::DeprecationWarning
--- a/sonic-chassisd/pytest.ini
+++ b/sonic-chassisd/pytest.ini
@@ -1,3 +1,2 @@
 [pytest]
-filterwarnings =
-    ignore::DeprecationWarning
+addopts = --cov=scripts --cov-report html --cov-report term --cov-report xml
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -183,10 +183,6 @@ class ChassisdDaemon(daemon_base.DaemonBase):
         chassis_tbl = swsscommon.Table(state_db, CHASSIS_INFO_TABLE)
         module_tbl = swsscommon.Table(state_db, CHASSIS_MODULE_INFO_TABLE)
 
-        # Connect to CONFIG_DB and check with chassis cfg table
-        config_db = daemon_base.db_connect("CONFIG_DB")
-        config_tbl = swsscommon.Table(config_db, CHASSIS_CFG_TABLE)
-
         # Post number-of-modules info to STATE_DB
         num_modules = platform_chassis.get_num_modules()
         fvs = swsscommon.FieldValuePairs([(CHASSIS_INFO_CARD_NUM_FIELD, str(num_modules))])
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -108,8 +108,9 @@ class ModuleConfigUpdater(logger.Logger):
         if not key.startswith(ModuleBase.MODULE_TYPE_SUPERVISOR) and \
            not key.startswith(ModuleBase.MODULE_TYPE_LINE) and \
            not key.startswith(ModuleBase.MODULE_TYPE_FABRIC):
-            (self.log_error("Incorrect module-name {}. Should start with {} or {} or {}".format(key,\
-             ModuleBase.MODULE_TYPE_SUPERVISOR, ModuleBase.MODULE_TYPE_LINE, ModuleBase.MODULE_TYPE_FABRIC)))
+            self.log_error("Incorrect module-name {}. Should start with {} or {} or {}".format(key,
+                           ModuleBase.MODULE_TYPE_SUPERVISOR, ModuleBase.MODULE_TYPE_LINE,
+                           ModuleBase.MODULE_TYPE_FABRIC))
             return
 
         module_index = try_get(self.chassis.get_module_index, key, default=INVALID_MODULE_INDEX)
@@ -121,8 +122,8 @@ class ModuleConfigUpdater(logger.Logger):
 
         if (admin_state == MODULE_ADMIN_DOWN) or (admin_state == MODULE_ADMIN_UP):
             #Setting the module to administratively up/down state
-            (self.log_info("Changing module {} to admin {} state".format(key,
-             'DOWN' if admin_state == MODULE_ADMIN_DOWN else 'UP')))
+            self.log_info("Changing module {} to admin {} state".format(key,
+                          'DOWN' if admin_state == MODULE_ADMIN_DOWN else 'UP'))
             try_get(self.chassis.get_module(module_index).set_admin_state, admin_state, default=False)
 
 #
@@ -181,8 +182,9 @@ class ModuleUpdater(logger.Logger):
                 if not key.startswith(ModuleBase.MODULE_TYPE_SUPERVISOR) and \
                    not key.startswith(ModuleBase.MODULE_TYPE_LINE) and \
                    not key.startswith(ModuleBase.MODULE_TYPE_FABRIC):
-                    (self.log_error("Incorrect module-name {}. Should start with {} or {} or {}".format(key,
-                     ModuleBase.MODULE_TYPE_SUPERVISOR, ModuleBase.MODULE_TYPE_LINE, ModuleBase.MODULE_TYPE_FABRIC)))
+                    self.log_error("Incorrect module-name {}. Should start with {} or {} or {}".format(key,
+                                   ModuleBase.MODULE_TYPE_SUPERVISOR, ModuleBase.MODULE_TYPE_LINE,
+                                   ModuleBase.MODULE_TYPE_FABRIC))
                     continue
 
                 fvs = swsscommon.FieldValuePairs([(CHASSIS_MODULE_INFO_DESC_FIELD, module_info_dict[CHASSIS_MODULE_INFO_DESC_FIELD]),
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -115,13 +115,13 @@ class ModuleConfigUpdater(logger.Logger):
 
         module_index = try_get(self.chassis.get_module_index, key, default=INVALID_MODULE_INDEX)
 
-        #Continue if the index is invalid
+        # Continue if the index is invalid
         if module_index < 0:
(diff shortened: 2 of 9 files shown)


sonic-chassisd/scripts/chassisd
```diff
@@ -0,0 +1,232 @@
...
+        self.log_info("Stop daemon main loop")
+
+        # Delete all the information from DB and then exit
+        for card_index in range(1, num_cards + 1):
+            card_tbl._del(CHASSIS_MODULE_INFO_KEY_TEMPLATE.format(card_index))
+
+        chassis_tbl._del(CHASSIS_INFO_KEY_TEMPLATE.format(1))
+
+        self.log_info("Shutting down...")
+
```
mprabhu-nokia, line 222: get_change_event() for async notification

lguohan:
can we start with python3, also what is the unit test that can be used to test it?

jleveque (changes requested):
Please also add unit tests for chassisd

mprabhu-nokia:
> 
> 
> Please also add unit tests for chassisd

Added UT interfacing with Mock classes.

sonic-chassisd/scripts/chassisd
mprabhu-nokia: Will be added to future-items list

sonic-chassisd/scripts/chassisd
```diff
@@ -0,0 +1,327 @@
+#!/usr/bin/env python2
```
jleveque, line 1: Please convert to Python 3, as Python 2 is deprecated, and we are in the process of converting all Python code to Python 3

jleveque (changes requested):

sonic-chassisd/scripts/chassisd
mprabhu-nokia: Moved to Python3.

sonic-chassisd/pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+filterwarnings =
+    ignore::DeprecationWarning
```
jleveque, line 3: I believe you can remove this warning filter, and instead, enable pytest coverage reporting. You can use this PR as an example: https://github.com/Azure/sonic-platform-common/pull/139/files. Make sure to add `'pytest-cov'` to `tests_require` and also update the .gitignore file in this repo to ignore the generated coverage files.

jleveque (changes requested):

mprabhu-nokia:
plugins: cov-2.6.0
collected 8 items

tests/test_chassisd.py ........                                                                                                                                                        [100%]

Name               Stmts   Miss  Cover
--------------------------------------
scripts/chassisd     177     64    64%
Coverage HTML written to dir htmlcov
Coverage XML written to file coverage.xml

The missing coverage is mainly around the area of:
* init, signal handling of the main loop
* async subscription handling of the config DB. select() handling etc.

However, the functional area has been separated to classes and fully covered.

sonic-chassisd/pytest.ini
mprabhu-nokia: Done. Please review. Attached report below.

jleveque:
@liat-grozovik: Can someone from your team please review?

@Staphylo: Can you also review, please?

sonic-chassisd/scripts/chassisd
```diff
@@ -0,0 +1,332 @@
...
+            self.stop.set()
+        elif sig == signal.SIGTERM:
+            self.log_info("Caught SIGTERM - exiting...")
+            self.stop.set()
+        else:
+            self.log_warning("Caught unhandled signal '" + sig + "'")
+
+    # Run daemon
+    def run(self):
+        global platform_chassis
```
Staphylo, line 279: I feel it's usually better to avoid globals when possible but this is mostly nitpicking.

sonic-chassisd/scripts/chassisd
```diff
@@ -0,0 +1,332 @@
...
+            if state == swsscommon.Select.TIMEOUT:
+                # Do not flood log when select times out
+                continue
+            if state != swsscommon.Select.OBJECT:
+                self.logger.log_warning("sel.select() did not return swsscommon.Select.OBJECT")
+                continue
+
+            (key, op, fvp) = sst.pop()
+
+            if op == 'SET':
```
Staphylo, line 245: Don't we need to check any key here like an `admin_status` one?
Also I'm not clear about bringing the card down when an entry is added to CONFIG_DB and bringing it up when the key is removed. Might be a misunderstanding on my end.

sonic-chassisd/scripts/chassisd
```diff
@@ -0,0 +1,332 @@
...
+    def module_db_update(self):
+        for module_index in range(0, self.num_modules):
+            module_info_dict = self._get_module_info(module_index)
+            if module_info_dict is not None:
+                key = module_info_dict[CHASSIS_MODULE_INFO_NAME_FIELD]
+
+                if not key.startswith(ModuleBase.MODULE_TYPE_SUPERVISOR) and \
+                   not key.startswith(ModuleBase.MODULE_TYPE_LINE) and \
+                   not key.startswith(ModuleBase.MODULE_TYPE_FABRIC):
+                    (self.log_error("Incorrect module-name {}. Should start with {} or {} or {}".format(key,
```
Staphylo, line 184: Is there a reason for the whole expression to be in parenthesis?
I've seen other occurrences of this pattern in this review.

sonic-chassisd/scripts/chassisd
```diff
@@ -0,0 +1,332 @@
...
+            return
+
+        module_index = try_get(self.chassis.get_module_index, key, default=INVALID_MODULE_INDEX)
+
+        #Continue if the index is invalid
+        if module_index < 0:
+            self.log_error("Unable to get module-index for key {} to set admin-state {}". format(key, admin_state))
+            return
+
+        if (admin_state == MODULE_ADMIN_DOWN) or (admin_state == MODULE_ADMIN_UP):
```
Staphylo, line 122: Should we assert/ValueError or print a log message if the admin_state parameter is invalid?

sonic-chassisd/scripts/chassisd
mprabhu-nokia: The key validation happens in 2 places. One, in the CLI context itself. Two, in module_config_update() function in this file. The validation was moved there since I found it easier to write UT for it.
In the HDL review, we had decided that there is no default configuration for modules unlike Ethernet interfaces. So, shutdown will add the configuration and startup with remove the configuration.

sonic-chassisd/scripts/chassisd
mprabhu-nokia: This is similar to how other pmon processes like thermalctld, psud etc have it.

sonic-chassisd/scripts/chassisd
mprabhu-nokia: The log line was too long. So, just for the new line I put it in the parenthesis. What is the recommended approach. Using '\'?

sonic-chassisd/scripts/chassisd
mprabhu-nokia: I will add an error log here. Thanks.

mprabhu-nokia: Actually, in the while loop where this routine is invoked, we are passing either MODULE_ADMIN_UP or MODULE_ADMIN_DOWN. So, no other value is possible.

sonic-chassisd/scripts/chassisd
Staphylo: I'm not sure what's the policy in this package but I would just not go to the line.
The `\` is only needed in a few corner cases, it doesn't apply here.

mprabhu-nokia: You are right. Cleaned up all instances in this file.

Staphylo (approved):

sonic-chassisd/scripts/chassisd
```diff
@@ -0,0 +1,334 @@
...
+    chassisd
+    Module information update daemon for SONiC
+    This daemon will loop to collect all modules related information and then write the information to state DB.
+    The loop interval is CHASSIS_INFO_UPDATE_PERIOD_SECS in seconds.
+"""
+
+try:
+    import sys
+    import signal
+    import threading
```
jleveque, line 13: Please rearrange alphabetically:

```
    import signal
    import sys
    import threading
```

jleveque (changes requested):

sonic-chassisd/scripts/chassisd
```diff
@@ -0,0 +1,334 @@
...
+           not key.startswith(ModuleBase.MODULE_TYPE_LINE) and \
+           not key.startswith(ModuleBase.MODULE_TYPE_FABRIC):
+            self.log_error("Incorrect module-name {}. Should start with {} or {} or {}".format(key,
+                           ModuleBase.MODULE_TYPE_SUPERVISOR, ModuleBase.MODULE_TYPE_LINE,
+                           ModuleBase.MODULE_TYPE_FABRIC))
+            return
+
+        module_index = try_get(self.chassis.get_module_index, key, default=INVALID_MODULE_INDEX)
+
+        #Continue if the index is invalid
```
jleveque, line 118: Please add a space after `#` for all comments through out the PR per PEP8 standards.

jleveque (changes requested):

sonic-chassisd/scripts/chassisd
mprabhu-nokia: Fixed.

sonic-chassisd/scripts/chassisd
mprabhu-nokia: Fixed.

jleveque (approved):

keboliu:
I think this daemon should only for the chassis system, so how to prevent it from running on a 1U system?

mprabhu-nokia:
> 
> 
> I think this daemon should only for the chassis system, so how to prevent it from running on a 1U system?

With the current implementation, the process will start but gracefully exit if not running for modular chassis. This is not part of critical processes, so should not affect PMON container. 
Going forward, one option would be for vendors to add skip_chassisd in their pmon_daemon_control.json file for 1 RU SKUs.

PR #98, opened 2020-10-05
[psud] Fix psud logging

vboykox:
Fix `Failed to update PSU data - global name 'self' is not defined` 

Signed-off-by: Volodymyr Boyko <volodymyrx.boiko@intel.com>


vboykox:
@jleveque

jleveque (approved):
Thanks for the fix!

PR #103, opened 2020-10-16
[thermalctld] Print exception using repr(e) to get more information

Junchao-Mellanox:
Why I did this?

The existing code print exception like:

```python
self.log_error('Caught exception while initializing thermal manager - {}'.format(e))
```

And it would print empty string if the exception class doesn't implement __str__. And the empty string is useless for debugging.

How to fix it?

Always print the exception using repr(e), like:

```python
self.log_error('Caught exception while initializing thermal manager - {}'.format(repr(e)))
```

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -14,7 +14,7 @@ try:
     from sonic_py_common import daemon_base, logger
     from sonic_py_common.task_base import ProcessTaskBase
 except ImportError as e:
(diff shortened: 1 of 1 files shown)


liat-grozovik (approved):

jleveque (approved):

abdosi:
@liat-grozovik Cherry-pick has conflict.

Junchao-Mellanox:
@liat-grozovik Created internal PR for 201911 https://github.com/Junchao-Mellanox/sonic-platform-daemons/pull/17

PR #112, opened 2020-11-14
[chassisd][thermalctld] Import of mock libraries determined by environment variable

jleveque:
Previously, chassisd and thermalctld assumed that the swsscommon library would not be installed in the unit testing environment. This is not a valid assumption, and would cause unit tests to fail if swsscommon was available in the unit test environement, because it would get imported, but there would be no Redis DB to communicate with.

This PR uses environment variables, which are set by the unit tests themselves, to determine whether to load the real or mock libraries. This solution is similar to what is done in sonic-utilities.


jleveque:
@mprabhu-nokia, @Junchao-Mellanox: FYI.

PR #123, opened 2020-11-19
[xcvrd] Fix xcvrd crash on other port prefixes

zzhiyuan:
Implement a better check for logical_port_name_to_physical_port_list function.

Currently it will only accept names that start with "Ethernet". If we try to add other ports it will crash.

In the future Arista will implement other port types. For example "Recirc0".

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -107,14 +107,14 @@ EMPTY_NAMESPACE = ''
 
 # Find out the underneath physical port list by logical name
 def logical_port_name_to_physical_port_list(port_name):
-    if port_name.startswith("Ethernet"):
+    try:
+        return [int(port_name)]
+    except ValueError:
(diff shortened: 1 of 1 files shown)


jleveque:
@zzhiyuan: Can you provide an example in the description?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -107,14 +107,14 @@
 
 # Find out the underneath physical port list by logical name
 def logical_port_name_to_physical_port_list(port_name):
-    if port_name.startswith("Ethernet"):
+    try:
+        return [int(port_name)]
```
jleveque, line 111: I'm not sure we even need to handle this case. All platforms should provide a port_config.ini (or the new platform.json) file which should specify a logical port name.

@keboliu: This has been present since your initial commit of xcvrd. Can you explain your reasoning for trying to cast the logical port name to an int if the port name doesn't start with "Ethernet"?

jleveque (approved):

PR #128, opened 2020-11-25
Align style with PEP8 standards

jleveque:
Align style with slightly modified PEP8 standards (extend maximum line length to 120 chars). This will also help in the transition to Python 3, where it is more strict about whitespace.

Done using `autopep8 --in-place --max-line-length 120` and some manual tweaks.


vdahiya12 (approved):

vdahiya12:
I would suggest being cautious towards pmon container, when we have an image for this change and everything is working in a sane manner. (submodule update) Autopep8 can sometimes mess up indentation, although i could not catch anything with my naked eyes.

lguohan:
why is this one not merged?

PR #134, opened 2020-12-11
[xcvrd] removing the delete notification message logic for command probe in mux cable driver

vdahiya12:
Summary:
This PR provides removes the delete logic on command probe message received from linkmgr after processing the message
### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Approach
added the changes in y_cable_helper.py by replacing the API

#### What is the motivation for this PR?
the delete message tends to create an error scenario if many probe messages come and redis-api fails to retrieve the message contents  

#### How did you do it?
removed the call for deletion of the message 

#### How did you verify/test it?
Ran the pmon docker with the changed file, and redis-cli to obtain the events and check the functionality is working

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
+++ b/sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
@@ -154,7 +154,7 @@ def update_tor_active_side(read_side, state, logical_port_name):
         return -1
 
 
-def update_appdb_port_mux_cable_response_table(logical_port_name, asic_index, appl_db):
+def update_appdb_port_mux_cable_response_table(logical_port_name, asic_index, appl_db, read_side):
 
     status = None
     y_cable_response_tbl = {}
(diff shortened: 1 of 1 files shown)


tahmed-dev (approved):

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -595,8 +594,14 @@ def task_worker(self):
                     probe_identifier = fvp_dict["command"]
 
                     if probe_identifier == "probe":
-                        update_appdb_port_mux_cable_response_table(port_m, asic_index, appl_db)
-                    delete_port_from_y_cable_command_table(port_m, y_cable_command_tbl[asic_index])
```
jleveque, line 599: Looks like this was the only place the function was getting called. Should you also delete the function definition now that it is unused?

vdahiya12: fixed

jleveque (approved):

PR #137, opened 2020-12-28
[psud] Fix issue where PSU Fan info is not updated in State DB

aravindmani-1:
**Issue:**

- The PSU FAN information is not updated in the redis-db.

**Steps to reproduce:**

- Load latest master image and check redis-db PSU Fan table.

**To fix this issue:**

- Initialize self.presence and other variables in PsuStatus dunder init to False instead of True.
- Import datetime module.
- Discussions related to this issue can be seen in https://github.com/Azure/sonic-platform-daemons/issues/136

UT:
[psu daemon fix UT.txt](https://github.com/Azure/sonic-platform-daemons/files/5746605/psu.daemon.fix.UT.txt)

### DIFF
--- a/sonic-psud/scripts/psud
+++ b/sonic-psud/scripts/psud
@@ -14,7 +14,7 @@ try:
     import signal
     import sys
     import threading
-
+    from datetime import datetime
     from sonic_py_common import daemon_base, logger
 
     # If unit testing is occurring, mock swsscommon and module_base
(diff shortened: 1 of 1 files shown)


sonic-psud/scripts/psud
```diff
@@ -14,7 +14,7 @@ try:
     import signal
     import sys
     import threading
-
+    from datetime import datetime
```
Junchao-Mellanox, line 17: Why we need this?

aravindmani-1: In the following piece of code, we update datetime.
https://github.com/Azure/sonic-platform-daemons/blob/e6c786bd02e6a253fcf67dec895e7c24e940c788/sonic-psud/scripts/psud#L557 
Since the datetime module is not imported, all the Fan fields will not be updated to DB.

Junchao-Mellanox (approved):

jleveque (approved):

PR #138, opened 2021-01-06
[multi-asic] Xcvrd process to exit if there are no Front End ASIC's

judyjoseph:
Closes issue: https://github.com/Azure/sonic-buildimage/issues/6097

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -74,6 +74,7 @@ SFPUTIL_LOAD_ERROR = 1
 PORT_CONFIG_LOAD_ERROR = 2
 NOT_IMPLEMENTED_ERROR = 3
 SFP_SYSTEM_ERROR = 4
+NO_FRONTEND_ASICS = 5
 
 RETRY_TIMES_FOR_SYSTEM_READY = 24
 RETRY_PERIOD_FOR_SYSTEM_READY_MSECS = 5000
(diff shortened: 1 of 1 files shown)


anamehra:
Is this change really required? pmon_daemon_control.json should be good enough to block spawning of xcvrd if required. We may have future use for this process on a distributed system.

judyjoseph:
> Is this change really required? pmon_daemon_control.json should be good enough to block spawning of xcvrd if required. We may have future use for this process on a distributed system.

The intend I had was to prevent XCRVd from crashing in case a user didn't know about the skip pmon daemon control pmon_daemon_control.json. There are places in the xcvrd code where we assume the front-end asic's are present by default -- which may not be the case in for eg: fabric cards in Chassis. 
Also Could you specify the future use case which you were referring here ?

anamehra:
> > Is this change really required? pmon_daemon_control.json should be good enough to block spawning of xcvrd if required. We may have future use for this process on a distributed system.
> 
> The intend I had was to prevent XCRVd from crashing in case a user didn't know about the skip pmon daemon control pmon_daemon_control.json. There are places in the xcvrd code where we assume the front-end asic's are present by default -- which may not be the case in for eg: fabric cards in Chassis.
> Also Could you specify the future use case which you were referring here ?
I was thinking about if xcrvd on Controller card should handle LC transreceivers as pseudo devices to handle OIR during LC OIR. But I think that could be part of chassis fault handling design. I think for now, it should be fine to disable this.

anamehra (approved):

judyjoseph:
If the xcvrd exit in between, it will cause the pmon docker to exit as it is a critical process. So I will not put this fix, would stick to the skip_flags in the pmon_daemon_control.json file itself to control whether we need to have xcvrd in pmon or not. Closing this PR.

PR #142, opened 2021-01-16
[pcied] Fix continuous warning message and config file location

sujinmkang:
- Remove the unnecessary warning message
- Change the configuration file (pcie.yaml) location to platform directory from plugin directory
- Change the deamon to python3


sujinmkang:
need to create a new PR.

PR #144, opened 2021-01-17
[pcied] Remove unnecessary message and move the configuration path

sujinmkang:
[pcied] Remove unnecessary message and move the configuration path from plugin directory to platform directory

This PR fixes https://github.com/Azure/sonic-buildimage/issues/5819 and fixes https://github.com/Azure/sonic-buildimage/issues/6437.

- Remove unnecessary repeated message
- Move the configuration path from plugin directory to platform directory

### DIFF
--- a/sonic-pcied/scripts/pcied
+++ b/sonic-pcied/scripts/pcied
@@ -42,7 +42,6 @@ class DaemonPcied(DaemonBase):
 
         (platform_path, _) = device_info.get_paths_to_platform_and_hwsku_dirs()
         pciefilePath = os.path.join(platform_path, PCIE_CONF_FILE)
-        sys.path.append(os.path.abspath(pciefilePath))
         if not os.path.exists(pciefilePath):
             self.log_error("Platform pcie configuration file doesn't exist! Exiting ...")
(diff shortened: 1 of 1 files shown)


sonic-pcied/scripts/pcied
```diff
@@ -57,17 +57,12 @@ class DaemonPcied(DaemonBase):
     def check_pcie_devices(self):
         try:
             platform_path, _ = device_info.get_paths_to_platform_and_hwsku_dirs()
-            platform_plugins_path = os.path.join(platform_path, "plugins")
-            sys.path.append(os.path.abspath(platform_plugins_path))
-            from pcieutil import PcieUtil
+            sys.path.append(os.path.abspath(platform_path))
```
jleveque, line 60: Why do we need to append `platform_path` to `sys.path`? We're not loading Python modules from here, correct? Only the pcie.yaml file resides there, right?

sujinmkang: I will change it.

jleveque (approved):

dprital:
@sujinmkang , we need this fix also for 202012. Can you please merge it to 2012 as well ?

PR #147, opened 2021-01-27
[ledd] Refactor to allow for more thorough unit testing; Increase unit test coverage

jleveque:
- Refactor ledd to allow for more thorough unit testing
    - Remove infinite loop from `run()` method to allow unit testing of the method. Instead, call the `run()` method in an infinite loop in `main()`
- Increase overall ledd unit test coverage from 48% to 93%.

Previous unit test coverage:

```
Name           Stmts   Miss  Cover
----------------------------------
scripts/ledd      66     34    48%
Coverage HTML written to dir htmlcov
Coverage XML written to file coverage.xml
```

Unit test coverage with this patch:

```
Name           Stmts   Miss  Cover
----------------------------------
scripts/ledd      72      5    93%
Coverage HTML written to dir htmlcov
Coverage XML written to file coverage.xml
```

### DIFF
--- a/sonic-ledd/scripts/ledd
+++ b/sonic-ledd/scripts/ledd
@@ -6,6 +6,7 @@
 """
 
 import getopt
+import os
 import sys
 
 from sonic_py_common import daemon_base
@@ -37,11 +38,12 @@ LEDUTIL_LOAD_ERROR = 1
 
 class DaemonLedd(daemon_base.DaemonBase):
 
-    # Run daemon
-    def run(self):
+    def __init__(self):
+        daemon_base.DaemonBase.__init__(self, SYSLOG_IDENTIFIER)
+
         # Load platform-specific LedControl module
         try:
-            led_control = self.load_platform_util(LED_MODULE_NAME, LED_CLASS_NAME)
+            self.led_control = self.load_platform_util(LED_MODULE_NAME, LED_CLASS_NAME)
         except Exception as e:
             self.log_error("Failed to load ledutil: %s" % (str(e)), True)
             sys.exit(LEDUTIL_LOAD_ERROR)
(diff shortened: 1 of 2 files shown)


jleveque:
@Staphylo: Can you please test this patch out on your platform(s) to confirm you don't see any regressions?

As discussed, we may move the ledd functionality in the future and hook it into the new platform API, but in the meantime, we want to ensure >= 90% unit test coverage of all code in the SONiC codebase, thus this PR.

Staphylo:
@jleveque I'll try testing your change today

Staphylo:
I tried your change and it's behaving similarly to the old one.
I couldn't verify by looking at a dut but I enabled some debug traces in our plugin to verify the calls.

Something that I realized is that changing a port from `up` to `down` or vice versa generates 10+ calls to `port_link_status_change`.
From up to down I saw 9 up events and 3 down events.
From down to up I saw 11 down events and 1 up event.

jleveque:
> I tried your change and it's behaving similarly to the old one.
> I couldn't verify by looking at a dut but I enabled some debug traces in our plugin to verify the calls.
> 
> Something that I realized is that changing a port from `up` to `down` or vice versa generates 10+ calls to `port_link_status_change`.
> From up to down I saw 9 up events and 3 down events.
> From down to up I saw 11 down events and 1 up event.

Thanks, @Staphylo! Regarding the 10+ calls to `port_link_status_change` -- you're saying this happens both in the current checked-in code as well as with this patch, correct? As long as the behavior is the same we can merge this and look into the reason why later. I just want to make sure that behavior isn't introduced by this PR.

Staphylo:
@jleveque, yes the same behavior can be observed before and after your change.
This change LGTM.

judyjoseph:
The changes Looks ok in multi-asic too Joe. 
The multiple triggers coming to port_link_status_change() - I had seen earlier too, had done a brief investigation at that time to find the event was remaining in queue in swsscommon for some time before cleared...need to check further.

judyjoseph (approved):

lguohan (approved):

PR #149, opened 2021-02-02
[ci] Test and build packages using Azure Pipelines

jleveque:
Configure Azure Pipelines to run unit tests on all packages, build all Python wheels and publish the test results, test coverage and resulting wheels.

Note: Azure Pipelines currently has a limitation such that it does not properly support publishing multiple test coverage reports in the same pipeline. If you publish multiple test coverage reports in the same pipeline (which I am doing here), the summary and report is shown for the last task only. Any previously uploaded data is ignored. Hopefully Azure Pipelines will add support for multiple test coverage reports in the near future. See [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-code-coverage-results?view=azure-devops#is-code-coverage-data-merged-when-multiple-files-are-provided-as-input-to-the-task-or-multiple-tasks-are-used-in-the-pipeline) for more info.


azure-pipelines.yml
```diff
@@ -4,16 +4,101 @@
...
+    python2 setup.py test
+    python2 setup.py bdist_wheel
+    python3 setup.py test
+    python3 setup.py bdist_wheel
+  workingDirectory: sonic-thermalctld
+  displayName: 'Test and build sonic-thermalctld'
+
+- publish: $(System.DefaultWorkingDirectory)/
+    artifact: sonic-platform-daemons
+    displayName: "Archive artifacts"
```
lguohan, line 104: add publish test results.

https://github.com/Azure/sonic-mgmt/blob/master/.azure-pipelines/run-test-template.yml#L88

jleveque: Working on it right now, actually :)

azure-pipelines.yml
```diff
@@ -4,16 +4,368 @@
...
+
+- script: |
+    set -e
+    python3 setup.py bdist_wheel
+  workingDirectory: sonic-thermalctld
+  displayName: 'Build sonic-thermalctld (Python 3)'
+
+- publish: '$(System.DefaultWorkingDirectory)/sonic-thermalctld/dist/'
+  artifact: sonic-thermalctld
+  displayName: "Publish sonic-thermalctld wheels"
```
lguohan, line 371: suggest to use template, I can see lots of duplication. check the sonic-mgmt/azure-pipelines.yml

jleveque: @lguohan: I templatized my solution, but it appears that Azure Pipelines runs the templates in the host OS rather than the specified container, so the tasks fail. I can't seem to find anything in the documentation regarding this. Do you have any pointers?

lguohan:
how about splitting into 5 jobs?

jleveque:
> how about splitting into 5 jobs to resolve the coverage report issue.

@lguohan: I just split it into 7 jobs (one per daemon). However, this does not resolve the coverage report issue, because Azure Pipelines currently reports code coverage at the Pipeline level, not the job level. Therefore, the issue still exists; the coverage report only reflects the last tests to finish, which now with parallel jobs can potentially vary.

lguohan (approved):

lguohan:
not sure why there is test failure.

jleveque:
> not sure why there is test failure.

I injected it on purpose for testing. I just reverted the commit. All test should pass now. Can you please review again?

lguohan:
i like such negative test

PR #157, opened 2021-02-25
[thermalctld] Refactor to allow for greater unit test coverage; Add more unit tests

jleveque:
#### Description
- Refactor thermalctld to reduce the amount of code in infinite loops, thus allowing us better unit test coverage
- Refactor mock_platform.py such that it inherits from sonic-platform-common in order to ensure it is aligned with the current API definitions (this introduces a test-time dependency on the sonic-platform-common package)
- Increase pytest verbosity to prevent truncation of error messages

- Miscellaneous cleanup:
    - Fixes to grammar
    - Remove unnecessary punctuation from log messages

- Increase overall unit test unit test coverage from 73% to 93%

Previous unit test coverage:

```
Name                  Stmts   Miss  Cover
-----------------------------------------
scripts/thermalctld     424    113    73%
```

Unit test coverage with this patch:

```
Name                  Stmts   Miss  Cover
-----------------------------------------
scripts/thermalctld     434     29    93%
```

#### Motivation and Context

As part of ongoing efforts to increase unit test coverage >= 90% across the SONiC codebase in order to prevent regressions

The refactored thermalctld has been tested on a physical DUT and the unit test results can be examined via the Azure Pipelines check builds in this PR.

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -254,8 +254,8 @@ class FanUpdater(logger.Logger):
 
         bad_fan_count = FanStatus.get_bad_fan_count()
         if bad_fan_count > 0 and old_bad_fan_count != bad_fan_count:
-            self.log_warning("Insufficient number of working fans warning: {} fans are not working.".format(
-                bad_fan_count
+            self.log_warning("Insufficient number of working fans warning: {} fan{} not working.".format(
+                bad_fan_count, " is" if bad_fan_count == 1 else "s are"
             ))
         elif old_bad_fan_count > 0 and bad_fan_count == 0:
             self.log_notice("Insufficient number of working fans warning cleared: all fans are back to normal.")
@@ -627,7 +627,7 @@ class TemperatureUpdater(logger.Logger):
         warning = False
         if temperature != NOT_AVAILABLE and temperature_status.set_over_temperature(temperature, high_threshold):
             self._log_on_status_changed(not temperature_status.over_temperature,
-                                        'High temperature warning cleared: {} temperature restore to {}C, high threshold {}C.'.
+                                        'High temperature warning cleared: {} temperature restored to {}C, high threshold {}C.'.
                                         format(name, temperature, high_threshold),
                                         'High temperature warning: {} current temperature {}C, high threshold {}C'.
                                         format(name, temperature, high_threshold)
@@ -636,7 +636,7 @@ class TemperatureUpdater(logger.Logger):
 
         if temperature != NOT_AVAILABLE and temperature_status.set_under_temperature(temperature, low_threshold):
             self._log_on_status_changed(not temperature_status.under_temperature,
-                                        'Low temperature warning cleared: {} temperature restore to {}C, low threshold {}C.'.
+                                        'Low temperature warning cleared: {} temperature restored to {}C, low threshold {}C.'.
                                         format(name, temperature, low_threshold),
                                         'Low temperature warning: {} current temperature {}C, low threshold {}C'.
                                         format(name, temperature, low_threshold)
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -1,28 +1,31 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 
 """
     thermalctld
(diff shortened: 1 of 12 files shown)


jleveque:
@Junchao-Mellanox: Can you please review and test these changes? Thanks in advance!

jleveque:
@Junchao-Mellanox: Can you please review and test these changes?

sonic-thermalctld/scripts/thermalctld
```diff
@@ -718,97 +731,118 @@ class ThermalControlDaemon(daemon_base.DaemonBase):
...
-                thermal_manager.initialize()
-                thermal_manager.load(ThermalControlDaemon.POLICY_FILE)
-                thermal_manager.init_thermal_algorithm(chassis)
-        except NotImplementedError:
-            self.log_warning('Thermal manager is not supported on this platform.')
+            if self.thermal_manager:
+                self.thermal_manager.run_policy(chassis)
         except Exception as e:
-            self.log_error('Caught exception while initializing thermal manager - {}'.format(repr(e)))
+            self.log_error('Caught exception while running thermal policy - {}'.format(e))
```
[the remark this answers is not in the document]
jleveque: The problem is `repr(e)` produces slightly different output between Python 2 and Python 3 (one ends with a comma, the other doesn't), whereas the output of `e` is identical. If we didn't have to continue supporting both versions, it wouldn't be a problem. Without this change, I will need to add ugly version checks in the unit tests. I can do it if you insist, and we can simply clean it all up once we no longer support Python 2.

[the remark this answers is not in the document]
Junchao-Mellanox: Understood. Then repr(e) is not necessary. As long as the exception is reproduceable, we can manually add it for debug.

Junchao-Mellanox (dismissed):

Junchao-Mellanox (dismissed):

sonic-thermalctld/scripts/thermalctld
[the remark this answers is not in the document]
jleveque: @Junchao-Mellanox: I went back and changed all log statements to print `repr(e)`, seeing as I'd rather make the unit tests temporarily messy than to potentially make debugging more difficult. Since we'll soon remove Python 2 support, this will be a short-lived inconvenience, and only if modifying the unit tests. Please review again.

sonic-thermalctld/tests/test_thermalctld.py
```diff
@@ -180,7 +180,11 @@ def test_update_fan_with_exception(self):
         assert fan.get_status_led() == MockFan.STATUS_LED_COLOR_RED
         assert fan_updater.log_warning.call_count == 1
 
-        fan_updater.log_warning.assert_called_with("Failed to update fan status - Failed to get speed")
+        # TODO: Clean this up once we no longer need to support Python 2
+        if sys.version_info.major == 3:
```
Junchao-Mellanox, line 184: How about we add a helper function like below:
```python
def get_exception_message(msg):
    if sys.version_info.major == 3:
        return 'Exception(\'{},\')'.format(msg)
    else:
        return 'Exception(\'{}\')'.format(msg)
```

And we don't have to change everywhere when removing python2 support.

jleveque: Thank you for the suggestion. However, I believe this unit test code will not change much, if at all before we remove Python 2 support. And with things the way they are, all we have to do is delete lines. If we make the change you suggested, there will actually be a bit more work needed when removing Python 2 support. Therefore, I prefer we just leave things as-is for the time being.

Junchao-Mellanox (approved):

sonic-thermalctld/setup.py
```diff
@@ -23,7 +23,8 @@
     tests_require=[
         'mock>=2.0.0; python_version < "3.3"',
         'pytest',
-        'pytest-cov'
+        'pytest-cov',
+        'sonic-platform-common'
```
vdahiya12, line 27: @jleveque  this dependency is causing https://github.com/Azure/sonic-buildimage/pull/7143 sonic-buildimage to fail. 
Do we need to remove it or add something ?

jleveque: PR to add dependency in sonic-buildimage [here](https://github.com/Azure/sonic-buildimage/pull/7181)

PR #162, opened 2021-03-07
[xcvrd] Fix crash: If 'dom_capability' not in port_info_dict, insert 'N/A'

jleveque:
#### Description

If 'dom_capability' not in port_info_dict, insert 'N/A'

#### Motivation and Context

Currently, some vendors are using custom transceiver info parsers which do not yet provide the `dom_capability` field in the results of `get_transceiver_info()`. However, PR https://github.com/Azure/sonic-platform-daemons/pull/72 introduced storing this value to State DB under the assumption that it would always be present. On platforms where this value is not present, it would cause xcvrd to crash (see issue: https://github.com/Azure/sonic-buildimage/issues/6978).

This change will prevent a crash if it is not present, and will in turn save `'N/A'` as the `dom_capability` value in State DB.

Tested on one affected platform.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -299,7 +299,8 @@ def post_port_sfp_info_to_db(logical_port_name, table, transceiver_dict,
                      ('application_advertisement', port_info_dict['application_advertisement']
                       if 'application_advertisement' in port_info_dict else 'N/A'),
                      ('is_replaceable', str(is_replaceable)),
-                     ('dom_capability',port_info_dict['dom_capability']),
+                     ('dom_capability', port_info_dict['dom_capability']
+                      if 'dom_capability' in port_info_dict else 'N/A'),
                      ])
                 table.set(port_name, fvs)
             else:


sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -299,7 +299,8 @@ def post_port_sfp_info_to_db(logical_port_name, table, transceiver_dict,
                      ('application_advertisement', port_info_dict['application_advertisement']
                       if 'application_advertisement' in port_info_dict else 'N/A'),
                      ('is_replaceable', str(is_replaceable)),
-                     ('dom_capability',port_info_dict['dom_capability']),
+                     ('dom_capability', port_info_dict['dom_capability']
+                      if 'dom_capability' in port_info_dict else 'N/A'),
```
vdahiya12, line 303: Do you think we should add this check for all the fields and raise a seperate PR for that ?
It will save us from future regressions

vdahiya12 (approved):

sonic-xcvrd/xcvrd/xcvrd.py
jleveque: That's tricky, because it's a slippery slope, as it would allow new platforms to *not* provide support for a field which all other platforms currently support. I think we should only apply it to new fields until we have refactored sonic-platform-common and all vendors have migrated to the shared parser implementations.

PR #164, opened 2021-03-11
[thermalctld] Use interval set in thermal policy if available

andywongarista:
#### Description
Update thermalctld to retrieve interval from thermal manager (which in turn loads an interval from thermal_policy.json) instead of using a constant.

#### Motivation and Context

https://github.com/Azure/sonic-platform-common/pull/178

This will allow platform vendors to specify an alternate interval for running thermal policies, e.g. 15 seconds instead of the current 60.

Verified that thermalctld runs without exiting.
Also ran test_thermalctld.py


jleveque:
Please fix new conflicts

jleveque:
@Junchao-Mellanox: Please review.

jleveque:
@Junchao-Mellanox: Please review.

jleveque:
It looks like something went wrong with your latest push. Can you please rebase against the latest master?

sujinmkang (approved):

sujinmkang:
lgtm

PR #165, opened 2021-03-14
[xcvrd] change firmware information fields name inside MUX_CABLE_INFO table for Y cable

vdahiya12:
This PR fixes the naming convention for firmware related fields for Y cable.
In particular all the fields are now named as tor_self and tor_peer in place of tor1 and tor 2 

#### Description
This PR fixes the naming convention for firmware related fields for Y cable.
In particular all the fields are now named as tor_self and tor_peer in place of tor1 and tor 2 

#### Motivation and Context
Required by telemetry team as part of their initial schema

Ran the fix on a 7050 testbed


sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -897,26 +889,26 @@ def post_port_mux_info_to_db(logical_port_name, table):
...
+                 ('build_slot2_tor_peer', str(mux_info_dict["build_slot2_tor_peer"])),
+                 ('version_slot1_tor_peer', str(mux_info_dict["version_slot1_tor_peer"])),
+                 ('version_slot2_tor_peer', str(mux_info_dict["version_slot2_tor_peer"])),
+                 ('run_slot1_tor_peer', str(mux_info_dict["run_slot1_tor_peer"])),
+                 ('run_slot2_tori_peer', str(mux_info_dict["run_slot2_tor_peer"])),
+                 ('commit_slot1_tor_peer', str(mux_info_dict["commit_slot1_tor_peer"])),
+                 ('commit_slot2_tor_peer', str(mux_info_dict["commit_slot2_tor_peer"])),
+                 ('empty_slot1_tor_peer', str(mux_info_dict["empty_slot1_tor_peer"])),
+                 ('empty_slot2_tor_peer', str(mux_info_dict["empty_slot2_tor_peer"]))
                  ])
```
jleveque, line 912: I believe you're adding all elements from `mux_info_dict` into the DB and using the key names from the dict as the key names for the DB. Therefore, you should be able to simplify this to one line:

```python
fvs = swsscommon.FieldValuePairs(list(mux_info_dict.items()))
```

Here, `list(mux_info_dict.items())` creates a list of tuples. However, I notice you're converting a lot of the values to strings. If this is necessary, then you might need to create the list of tuples then walk the list and convert the second element of each tuple to a string, but it would be a lot less code and maintenance than this, as long as we intend to push every item from `mux_info_dict` to the DB.

We can save this for a future cleanup PR.

jleveque:
Check build failing because unit tests need to be updated to reflect these changes.

vdahiya12:
> Check build failing because unit tests need to be updated to reflect these changes.

fixed

jleveque (approved):

PR #166, opened 2021-03-17
[thermalctld] Update threshold values

d-dashkov:
Signed-off-by: d-dashkov <Dmytro_Dashkov@Jabil.com>

#### Description
The thermal daemon has been modified, before setting the thresholds values in Redis, compares them from Redis with the platform API values, if they differ, the update process sets these new values for the platform.

#### Motivation and Context
Associated with PR: https://github.com/Azure/sonic-platform-common/pull/180

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -506,6 +506,7 @@ class TemperatureStatus(logger.Logger):
 class TemperatureUpdater(logger.Logger):
     # Temperature information table name in database
     TEMPER_INFO_TABLE_NAME = 'TEMPERATURE_INFO'
+    THRESH_NAME_LIST = ['high_threshold', 'low_threshold', 'critical_high_threshold', 'critical_low_threshold']
 
     def __init__(self, log_identifier, chassis):
         """
@@ -517,7 +518,9 @@ class TemperatureUpdater(logger.Logger):
         self.chassis = chassis
         self.temperature_status_dict = {}
         state_db = daemon_base.db_connect("STATE_DB")
+        config_db = daemon_base.db_connect("CONFIG_DB")
         self.table = swsscommon.Table(state_db, TemperatureUpdater.TEMPER_INFO_TABLE_NAME)
+        self.config_table = swsscommon.Table(config_db, TemperatureUpdater.TEMPER_INFO_TABLE_NAME)
         self.phy_entity_table = swsscommon.Table(state_db, PHYSICAL_ENTITY_INFO_TABLE)
         self.chassis_table = None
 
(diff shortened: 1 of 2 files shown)


jleveque (changes requested):
Please fix unit test failures

sonic-thermalctld/scripts/thermalctld
```diff
@@ -640,6 +640,36 @@ class TemperatureUpdater(logger.Logger):
                                         )
         warning = warning | temperature_status.under_temperature
 
+        THRESH_NAME_LIST = ['high_threshold', 'low_threshold', 'critical_high_threshold', 'critical_low_threshold']
```
Junchao-Mellanox, line 643: The following 3 variables can be class level

sonic-thermalctld/scripts/thermalctld
```diff
@@ -640,6 +640,36 @@ class TemperatureUpdater(logger.Logger):
...
+            if second == NOT_AVAILABLE:
+                continue
+
+            th_type = 0
+            if first in THRESH_NAME_LIST:
+                th_type = THRESH_NAME_LIST.index(first)
+            else:
+                continue
+
+            if second != str(THRESH_COMP_VAL_LIST[th_type]):
```
Junchao-Mellanox, line 659: "second" here is the value of a threshold, while THRESH_COMP_VAL_LIST[th_type] is a element in THRESH_COMP_VAL_LIST, i suppose they never equal.

sonic-thermalctld/scripts/thermalctld
```diff
@@ -640,6 +640,36 @@ class TemperatureUpdater(logger.Logger):
...
+                try:
+                    if THRESH_FUNC_LIST[th_type](float(second)) is True:
```
Junchao-Mellanox, line 661: Why you get a value from STATE_DB and set it to the platform?

Junchao-Mellanox (changes requested):

Junchao-Mellanox:
I don't quite understand the change here. If you want to allow user to change the temperature threshold, you probably need:
1. A CLI to change the CONFIG_DB 
2. A service/daemon to read the CONFIG_DB changes and do action

STATE_DB is used to reflect the "state" of the switch.

sonic-thermalctld/scripts/thermalctld
d-dashkov: Second is new value in string and THRESH_COMP_VAL_LIST[th_type] is current value in float, which I cast to string, so they should be equal.
```
second: <class 'str'>111.0
THRESH_COMP_VAL_LIST[th_type]: <class 'float'>111.0
```

sonic-thermalctld/scripts/thermalctld
d-dashkov: This was misunderstood by the SONiC DB architecture and now I have moved the new values to CONFIG_DB as you suggested.

PR #167, opened 2021-03-22
[thermalctld] Initialize fan led in thermalctld for the first run

Junchao-Mellanox:
#### Description
Initialize fan led in thermalcltd for the first run. Add a flag "led_initialized" in FanStatus and set it as False on __init__ function. FanUpdater will use this flag to determine if fan led should be set even if no fan event detected.

#### Motivation and Context
Make sure fan led is initialized to correct color.

Manual test


liat-grozovik:
@Junchao-Mellanox  could it be an additional change to https://github.com/Azure/sonic-buildimage/pull/7114

liat-grozovik (approved):

jleveque (approved):

PR #169, opened 2021-03-30
[xcvrd] Gracefully handle improper 'specification_compliance' field; also fix other potential bugs

jleveque:
#### Description

Gracefully handle improper 'specification_compliance' field; also fix other potential bugs and adjust some style

#### Motivation and Context
The 'specification_compliance' field of transceiver info is expected to be a string representation of a dictionary. However, there is a chance, upon some kind of platform issue that a vendor's platform API returns something like 'N/A'. In this case, xcrvd would crash. Rather than crash, xcvrd should handle this gracefully and log a warning message instead.

Also fixed a couple potential bugs where `default_dict` was being compared to `0`, when it should have been `len(default_dict)`

Also rename some constants using uppercase naming convention.

Tested on one platform; also added a unit test

### DIFF
--- a/sonic-xcvrd/tests/test_xcvrd.py
+++ b/sonic-xcvrd/tests/test_xcvrd.py
@@ -317,3 +317,25 @@ class TestXcvrdScript(object):
         mux_tbl = Table("STATE_DB", y_cable_helper.MUX_CABLE_STATIC_INFO_TABLE)
         rc = post_port_mux_static_info_to_db(logical_port_name, mux_tbl)
         assert(rc != -1)
+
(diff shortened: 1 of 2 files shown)


jleveque:
@dgsudharsan: Can you please review?

dgsudharsan (dismissed):

jleveque:
@dgsudharsan: I made one more small change to ensure the function doesn't return `None`. Can you please review again?

dgsudharsan (dismissed):

dgsudharsan (approved):

PR #170, opened 2021-04-06
[thermalctld] Fix 'NameError("name 'chassis' is not defined")' error in log

noaOrMlnx:
#### Description

Changed the chassis definition to be a part of the class do the chassis will be found in run() function
#### Motivation and Context

Without the change, an error appears in syslog. 
```  ERR pmon#thermalctld: Caught exception while running thermal policy - NameError("name 'chassis' is not defined") ```

N/A


shlomibitton (approved):

keboliu:
@Junchao-Mellanox, please also review

Junchao-Mellanox (approved):

jleveque (approved):

PR #172, opened 2021-04-08
[thermalctld] No need exit thermalcltd when loading invalid policy file

Junchao-Mellanox:
#### Description
Currently, when thermalctld loading an invalid policy file, it catches the exception and call sys.exit. However, there is a sub process created and sys.exit cause the daemon hang forever. Actually, it is not necessary to exit thermalctld here because:

1. Other than running thermal policy, there is a sub process which monitoring thermal status. So even if there is an invalid policy file, thermal monitoring can still work.
2. Even if we exit here, supervisord will restart thermalctld and it fall into the same exception again and again. 

#### Motivation and Context
The issue failed a few test cases in test_platform_info.py

test_platform_info.py

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -761,7 +761,6 @@ class ThermalControlDaemon(daemon_base.DaemonBase):
             self.log_warning('Thermal manager is not supported on this platform')
         except Exception as e:
             self.log_error('Caught exception while initializing thermal manager - {}'.format(repr(e)))
-            sys.exit(ERR_INIT_FAILED)
 
     def deinit(self):
         """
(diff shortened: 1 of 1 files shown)


sonic-thermalctld/scripts/thermalctld
```diff
@@ -761,7 +761,6 @@ class ThermalControlDaemon(daemon_base.DaemonBase):
             self.log_warning('Thermal manager is not supported on this platform')
         except Exception as e:
             self.log_error('Caught exception while initializing thermal manager - {}'.format(repr(e)))
-            sys.exit(ERR_INIT_FAILED)
```
jleveque, line 764: Since this is the only place `ERR_INIT_FAILED` is used, you should also remove the definition at the top of the file.

jleveque (changes requested):

sonic-thermalctld/scripts/thermalctld
Junchao-Mellanox: Sure

jleveque (approved):

PR #173, opened 2021-04-16
[muxcable] Fix Redis Selectable Processing

tahmed-dev:
Selectable object can return multiple fields, however current code
process only the first filed and wait till the next selectable
is triggered. This PR consumes all fields that were returned by
a selectable object. This results in reducing processing time from
order of second to order of millisecond.

singed-off-by: Tamer Ahmed <tamer.ahmed@microsoft.com>

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
+++ b/sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
@@ -1054,66 +1054,71 @@ class YCableTableUpdateTask(object):
             namespace = redisSelectObj.getDbConnector().getNamespace()
             asic_index = multi_asic.get_asic_index_from_namespace(namespace)
 
-            (port, op, fvp) = status_tbl[asic_index].pop()
-            if fvp:
-                # This check might be redundant, to check, the presence of this Port in keys
-                # in logical_port_list but keep for now for coherency
-                # also skip checking in logical_port_list inside sfp_util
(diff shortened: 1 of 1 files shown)


lguohan (approved):
:shipit:

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -1054,66 +1054,71 @@ def task_worker(self):
...
-
-                if "state" in fvp_dict:
-                    # got a state change
-                    new_status = fvp_dict["state"]
-                    (status, fvs) = y_cable_tbl[asic_index].get(port)
-                    if status is False:
-                        helper_logger.log_warning("Could not retreive fieldvalue pairs for {}, inside state_db table {}".format(
-                            port, y_cable_tbl[asic_index]))
+            while True:
+                (port, op, fvp) = status_tbl[asic_index].pop()
```
lguohan, line 1058: i always think swss-common should provide pops that allows user to pop everything. @qiluo-msft, we really should fix that.

jleveque (approved):

vdahiya12 (approved):

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
qiluo-msft: Yes, I will fix it.

PR #175, opened 2021-04-26
Collect asic info and store in CHASSIS_STATE_DB

ngoc-do:
Signed-off-by: ngocdo <ngocdo@arista.com>

chassisd inside pmon queries asic information from platform modules and stores it into CHASSIS_STATE_DB of redis_chassis.

The reason we keep information in CHASSIS_STATE_DB because it's accessible by swss containers.

Asic information includes asic pci address, module name, and asic id in the module. For example, CHASSIS_STATE_DB will look like below:

```
127.0.0.1:6379[6]> HGETALL "CHASSIS_ASIC_TABLE|ASIC0"
1) "asic_pci_address"
2) "0000:09:00.0"
3) "module_name"
4) "FABRIC-CARD0"
5) "asic_id_in_module"
6) "0"
127.0.0.1:6379[6]> HGETALL "CHASSIS_ASIC_TABLE|ASIC1"
1) "asic_pci_address"
2) "0000:0b:00.0"
3) "module_name"
4) "FABRIC-CARD0"
5) "asic_id_in_module"
6) "1"
127.0.0.1:6379[6]> HGETALL "CHASSIS_ASIC_TABLE|ASIC2"
1) "asic_pci_address"
2) "0000:13:00.0"
3) "module_name"
4) "FABRIC-CARD1"
5) "asic_id_in_module"
6) "0"
```

This PR will merge only after https://github.com/Azure/sonic-platform-common/pull/185 merges since this PR calls `get_all_asics` that is implemented in the other one.

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -44,6 +44,13 @@ CHASSIS_MODULE_INFO_NAME_FIELD = 'name'
 CHASSIS_MODULE_INFO_DESC_FIELD = 'desc'
 CHASSIS_MODULE_INFO_SLOT_FIELD = 'slot'
 CHASSIS_MODULE_INFO_OPERSTATUS_FIELD = 'oper_status'
+CHASSIS_MODULE_INFO_NUM_ASICS_FIELD = 'num_asics'
+CHASSIS_MODULE_INFO_ASICS = 'asics'
+
+CHASSIS_ASIC_INFO_TABLE = 'CHASSIS_ASIC_TABLE'
+CHASSIS_ASIC = 'asic'
+CHASSIS_ASIC_PCI_ADDRESS_FIELD = 'asic_pci_address'
+CHASSIS_ASIC_ID_IN_MODULE_FIELD = 'asic_id_in_module'
 
 CHASSIS_MIDPLANE_INFO_TABLE = 'CHASSIS_MIDPLANE_TABLE'
 CHASSIS_MIDPLANE_INFO_KEY_TEMPLATE = 'CHASSIS_MIDPLANE {}'
@@ -157,6 +164,7 @@ class ModuleUpdater(logger.Logger):
         state_db = daemon_base.db_connect("STATE_DB")
         self.chassis_table = swsscommon.Table(state_db, CHASSIS_INFO_TABLE)
         self.module_table = swsscommon.Table(state_db, CHASSIS_MODULE_INFO_TABLE)
+        self.asic_table = swsscommon.Table(state_db, CHASSIS_ASIC_INFO_TABLE)
         self.midplane_table = swsscommon.Table(state_db, CHASSIS_MIDPLANE_INFO_TABLE)
         self.info_dict_keys = [CHASSIS_MODULE_INFO_NAME_FIELD,
                                CHASSIS_MODULE_INFO_DESC_FIELD,
@@ -211,8 +219,17 @@ class ModuleUpdater(logger.Logger):
                 fvs = swsscommon.FieldValuePairs([(CHASSIS_MODULE_INFO_DESC_FIELD, module_info_dict[CHASSIS_MODULE_INFO_DESC_FIELD]),
                                                   (CHASSIS_MODULE_INFO_SLOT_FIELD,
                                                    module_info_dict[CHASSIS_MODULE_INFO_SLOT_FIELD]),
-                                                  (CHASSIS_MODULE_INFO_OPERSTATUS_FIELD, module_info_dict[CHASSIS_MODULE_INFO_OPERSTATUS_FIELD])])
+                                                  (CHASSIS_MODULE_INFO_OPERSTATUS_FIELD, module_info_dict[CHASSIS_MODULE_INFO_OPERSTATUS_FIELD]),
+                                                  (CHASSIS_MODULE_INFO_NUM_ASICS_FIELD, str(len(module_info_dict[CHASSIS_MODULE_INFO_ASICS])))])
                 self.module_table.set(key, fvs)
+
+                for asic_id, asic in enumerate(module_info_dict[CHASSIS_MODULE_INFO_ASICS]):
+                   asic_global_id, asic_pci_addr = asic
+                   asic_key = "%s%s" % (CHASSIS_ASIC, asic_global_id)
+                   asic_fvs = swsscommon.FieldValuePairs([(CHASSIS_ASIC_PCI_ADDRESS_FIELD, asic_pci_addr),
+                                                          (CHASSIS_MODULE_INFO_NAME_FIELD, key),
+                                                          (CHASSIS_ASIC_ID_IN_MODULE_FIELD, str(asic_id))])
+                   self.asic_table.set(asic_key, asic_fvs)
 
     def _get_module_info(self, module_index):
         """
@@ -225,11 +242,14 @@ class ModuleUpdater(logger.Logger):
         slot = try_get(self.chassis.get_module(module_index).get_slot, default=INVALID_SLOT)
         status = try_get(self.chassis.get_module(module_index).get_oper_status,
                          default=ModuleBase.MODULE_STATUS_OFFLINE)
+        asics = try_get(self.chassis.get_module(module_index).get_all_asics,
+                        default=[])
 
         module_info_dict[CHASSIS_MODULE_INFO_NAME_FIELD] = name
         module_info_dict[CHASSIS_MODULE_INFO_DESC_FIELD] = str(desc)
         module_info_dict[CHASSIS_MODULE_INFO_SLOT_FIELD] = str(slot)
         module_info_dict[CHASSIS_MODULE_INFO_OPERSTATUS_FIELD] = str(status)
+        module_info_dict[CHASSIS_MODULE_INFO_ASICS] = asics
 
         return module_info_dict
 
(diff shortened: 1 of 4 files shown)


sonic-chassisd/scripts/chassisd
```diff
@@ -211,9 +219,18 @@ class ModuleUpdater(logger.Logger):
...
                 self.module_table.set(key, fvs)
 
+                for asic_id, asic in enumerate(module_info_dict[CHASSIS_MODULE_INFO_ASICS]):
+                   asic_global_id, asic_pci_addr = asic
+                   asic_key = "%s%s" % (CHASSIS_ASIC, asic_global_id)
+                   asic_fvs = swsscommon.FieldValuePairs([(CHASSIS_ASIC_PCI_ADDRESS_FIELD, asic_pci_addr),
+                                                          (CHASSIS_MODULE_INFO_NAME_FIELD, key),
+                                                          (CHASSIS_ASIC_ID_IN_MODULE_FIELD, str(asic_id))])
+                   self.asic_table.set(asic_key, asic_fvs)
+
```
mprabhu-nokia, line 233: We need to add pytest testcases for UT and coverage.

mprabhu-nokia (changes requested):

sonic-chassisd/scripts/chassisd
```diff
@@ -211,9 +219,18 @@ class ModuleUpdater(logger.Logger):
...
+                                                  (CHASSIS_MODULE_INFO_NUM_ASICS_FIELD, str(len(module_info_dict[CHASSIS_MODULE_INFO_ASICS])))])
                 self.module_table.set(key, fvs)
 
+                for asic_id, asic in enumerate(module_info_dict[CHASSIS_MODULE_INFO_ASICS]):
+                   asic_global_id, asic_pci_addr = asic
+                   asic_key = "%s%s" % (CHASSIS_ASIC, asic_global_id)
+                   asic_fvs = swsscommon.FieldValuePairs([(CHASSIS_ASIC_PCI_ADDRESS_FIELD, asic_pci_addr),
+                                                          (CHASSIS_MODULE_INFO_NAME_FIELD, key),
+                                                          (CHASSIS_ASIC_ID_IN_MODULE_FIELD, str(asic_id))])
+                   self.asic_table.set(asic_key, asic_fvs)
```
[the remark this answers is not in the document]
ngoc-do: Thanks. Updated.

sonic-chassisd/scripts/chassisd
ngoc-do: I will add tests in the next commit.

ngoc-do: Test was added.

sonic-chassisd/scripts/chassisd
```diff
@@ -163,6 +170,9 @@ class ModuleUpdater(logger.Logger):
                                CHASSIS_MODULE_INFO_SLOT_FIELD,
                                CHASSIS_MODULE_INFO_OPERSTATUS_FIELD]
 
+        chassis_state_db = daemon_base.db_connect("CHASSIS_STATE_DB")
```
judyjoseph, line 173: This new table is kept in CHASSIS_STATE_DB, unlike the other tables like MIDPLANE_TABLE etc that is kept in STATE_DB ?

judyjoseph: See this comment in PR description that the info is kept in CHASSIS_STATE_DB because it's accessible by swss containers. Does the swss use this asic present info ?

judyjoseph (approved):

anshuv-mfst:
@judyjoseph - can you please merge asap please, needed for 202106 release

sonic-chassisd/scripts/chassisd
ngoc-do: There are a couple of places we can keep asic info like STATE_DB (redis instance) or CHASSIS_STATE_DB (redis chassis instance). 

CHASSIS_STATE_DB is accessible from every where in swss like`/usr/bin/swss.sh or /usr/bin/local/swss.sh` - script creating swss container` and `docker_init.sh` inside swss container. Inside swss container, STATE_DB is not accessible (asic database is mapped, instead of database).  So CHASSIS_STATE_DB is a safer choice.

sonic-chassisd/tests/test_chassisd.py
```diff
@@ -366,3 +368,76 @@ def test_midplane_presence_supervisor():
...
+    module_updater.module_db_update()
+    module_updater.check_midplane_reachability()
+
+    #Asic presence on fabric module
+    fabric.set_oper_status(ModuleBase.MODULE_STATUS_ONLINE)
+    module_updater.module_db_update()
+    fabric_asic_table = module_updater.asic_table
+    assert len(fabric_asic_table.getKeys()) == 2
+
+    def test_fabric_asic(asic_name, asic_pci_address, module_name, asic_id_in_module):
```
judyjoseph, line 419: Seems this new test test_fabric_asic is not being run. To correct the function definition format here.
```
tests/test_chassis_db_init.py::test_provision_db PASSED                  [  8%]
tests/test_chassisd.py::test_moduleupdater_check_valid_fields PASSED     [ 16%]
tests/test_chassisd.py::test_moduleupdater_check_invalid_name PASSED     [ 25%]
tests/test_chassisd.py::test_moduleupdater_check_status_update PASSED    [ 33%]
tests/test_chassisd.py::test_moduleupdater_check_deinit PASSED           [ 41%]
tests/test_chassisd.py::test_configupdater_check_valid_names PASSED      [ 50%]
tests/test_chassisd.py::test_configupdater_check_valid_index PASSED      [ 58%]
tests/test_chassisd.py::test_configupdater_check_admin_state PASSED      [ 66%]
tests/test_chassisd.py::test_configupdater_check_num_modules PASSED      [ 75%]
tests/test_chassisd.py::test_midplane_presence_modules PASSED            [ 83%]
tests/test_chassisd.py::test_midplane_presence_supervisor PASSED         [ 91%]
tests/test_chassisd.py::test_asic_presence PASSED                        [100%]
```

ngoc-do: No, `test_fabric_asic` is not supposed to run as an unit test. 

It's part of `test_asic_presence`

`tests/test_chassisd.py::test_asic_presence PASSED                        [100%]`

judyjoseph: As a practice, we keep the function names starting with test_* as the testcases selected by pytest during the collection process. So we could remove the "test_" prefix for this function and define it outside the function def test_asic_presence ?

ngoc-do: Sure, I'll remove test_ prefix, but I think keeping it inside is clearer. Does it make sense to you?

judyjoseph (approved):

mprabhu-nokia (approved):

mprabhu-nokia:
@judyjoseph I don't think I have permissions to merge. Could you please help with that? I have approved.

sonic-chassisd/scripts/chassisd
```diff
@@ -211,9 +228,29 @@ class ModuleUpdater(logger.Logger):
...
+                   asic_fvs = swsscommon.FieldValuePairs([(CHASSIS_ASIC_PCI_ADDRESS_FIELD, asic_pci_addr),
+                                                          (CHASSIS_MODULE_INFO_NAME_FIELD, key),
+                                                          (CHASSIS_ASIC_ID_IN_MODULE_FIELD, str(asic_id))])
+                   self.asic_table.set(asic_key, asic_fvs)
+
+        # Asics that are on the "not online" modules need to be cleaned up
+        asics = list(self.asic_table.getKeys())
+        for asic in asics:
+           fvs = self.asic_table.get(asic)
+           if fvs[CHASSIS_MODULE_INFO_NAME_FIELD] in notOnlineModules:
```
mprabhu-nokia, line 251: @ngoc-do 
Looks like there is a bug here? Are you seeing this as well? fvs is in the format below:
```
 [True, (('asic_pci_address', 'n/a'), ('name', 'FABRIC-CARD6'), ('asic_id_in_module', '1'))]
```

However, if the code is changed to below, it works:
```
fvp = dict(fvs[1])
if fvp[CHASSIS_MODULE_INFO_NAME_FIELD] in notOnlineModules:
```

ngoc-do: Hmm I don't see that. But if this happens, the unit test will fail, I think?

mprabhu-nokia: unit-tests use mock_swsscommon.py, where it uses a simple dictionary. So, unit-tests are passing. However, I am seeing the behavior I mentioned previously in the real hw environment. Are you testing on the near latest github? Will try and investigate. Will help if you can confirm/investigate as well.

ngoc-do: Ah OK. 

Yes,  I tested on our chassis when committing the change. But it could be that there were new changes at the upstream, and I didn't have latest upstream in our local repos by the time I tested. I will double check, but I'm afraid that our local repos hasn't been updated at this moment. So if you see it needs to be fixed, free to do that.

ngoc-do: OK, I found it happened on our latest update.

I have a fix in https://github.com/Azure/sonic-platform-daemons/pull/203. Please review.

judyjoseph:
This commit is already present in 202106 branch

PR #176, opened 2021-04-26
[xcvrd] Change the y_cable presence logic to use "mux_cable" table as identifier from Config DB

vdahiya12:

This PR changes the logic to use "mux_cable" table as identifier, and only those ports which have this table inside config DB and key pair of "state:auto/active" will be processed as having a Y-Cable port.
The earlier logic was using the PORT table tag with a mux_cable:true key-value pair. 

#### Description
change the logic to initiate Y-Cable logic

#### Motivation and Context
Required for changing the logic, so that this can be integrated with images which don't support "mux_cable" tag inside config DB

Ran the change on Arista 7050cx3 platform by restarting the container

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
+++ b/sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
@@ -283,35 +283,37 @@ def check_identifier_presence_and_update_mux_table_entry(state_db, port_tbl, y_c
         # Convert list of tuples to a dictionary
         mux_table_dict = dict(fvs)
         if "state" in mux_table_dict:
+            val = mux_table_dict.get("state", None)
+            if val in ["active", "auto"]:
 
-            y_cable_asic_table = y_cable_tbl.get(asic_index, None)
-            mux_asic_table = mux_tbl.get(asic_index, None)
-            static_mux_asic_table = static_tbl.get(asic_index, None)
-            if y_cable_presence[0] is True and y_cable_asic_table is not None and mux_asic_table is not None and static_mux_asic_table is not None:
-                # fill in the newly found entry
-                read_y_cable_and_update_statedb_port_tbl(
-                    logical_port_name, y_cable_tbl[asic_index])
-                post_port_mux_info_to_db(logical_port_name,  mux_tbl[asic_index])
-                post_port_mux_static_info_to_db(logical_port_name,  static_tbl[asic_index])
+                y_cable_asic_table = y_cable_tbl.get(asic_index, None)
+                mux_asic_table = mux_tbl.get(asic_index, None)
+                static_mux_asic_table = static_tbl.get(asic_index, None)
+                if y_cable_presence[0] is True and y_cable_asic_table is not None and mux_asic_table is not None and static_mux_asic_table is not None:
+                    # fill in the newly found entry
+                    read_y_cable_and_update_statedb_port_tbl(
+                        logical_port_name, y_cable_tbl[asic_index])
+                    post_port_mux_info_to_db(logical_port_name,  mux_tbl[asic_index])
+                    post_port_mux_static_info_to_db(logical_port_name,  static_tbl[asic_index])
 
-            else:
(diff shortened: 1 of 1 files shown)


sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -282,9 +282,9 @@ def check_identifier_presence_and_update_mux_table_entry(state_db, port_tbl, y_c
     else:
         # Convert list of tuples to a dictionary
         mux_table_dict = dict(fvs)
-        if "mux_cable" in mux_table_dict:
-            val = mux_table_dict.get("mux_cable", None)
-            if val == "true":
+        if "state" in mux_table_dict:
+            val = mux_table_dict.get("state", None)
+            if val == "active" or val == "auto":
```
theasianpianist, line 287: This seems unnecessary, `val` will always be either `active` or `auto`

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -518,9 +518,9 @@ def check_identifier_presence_and_update_mux_info_entry(state_db, mux_tbl, asic_
...
```
theasianpianist, line 523: same as obev

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
jleveque: It can't hurt to confirm. However, I suggest simplifying and making more Pythonic:

```python
if val in ["active", "auto"]:
```

theasianpianist (approved):

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
vdahiya12: fixed

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
vdahiya12: fixed according to Joe's suggestion

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -282,9 +282,9 @@ def check_identifier_presence_and_update_mux_table_entry(state_db, port_tbl, y_c
     else:
         # Convert list of tuples to a dictionary
         mux_table_dict = dict(fvs)
-        if "mux_cable" in mux_table_dict:
-            val = mux_table_dict.get("mux_cable", None)
-            if val == "true":
+        if "state" in mux_table_dict:
+            val = mux_table_dict.get("state", None)
+            if val in ["active", "auto"]:
```
jleveque, line 287: Since you added the check back, should we do anything in the `else` case here? E.g., log a warning/info message? Otherwise it might be difficult to track down the issue if val is not one of the expected values.

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -518,9 +518,9 @@ def check_identifier_presence_and_update_mux_info_entry(state_db, mux_tbl, asic_
...
```
jleveque, line 523: Same question as above.

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
vdahiya12: added a log_warning; fixed

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
vdahiya12: fixed

jleveque (approved):

PR #179, opened 2021-05-06
[psud] Add PSU Hardware Revision to Redis STATE_DB

robocoder99:
#### Description
Added "hardware revision" field to list of platform fields to sync to STATE_DB for the PSU. Also updated relevant unit tests. 

#### Motivation and Context
Now that hardware revision exists as a platform 2.0 field for all devices, it is appropriate to synchronize this field to STATE_DB for PSUs as is done with all other fields. This will allow this field to be exposed to CLI tools through psushow in the future which reads state from STATE_DB. 

This has been tested on all mellanox platforms aside from MSN2010. Tested by deploying change and verifying that "revision" was present in STATE_DB for all platforms that support it.

### DIFF
--- a/sonic-psud/scripts/psud
+++ b/sonic-psud/scripts/psud
@@ -48,6 +48,7 @@ PSU_INFO_KEY_TEMPLATE = 'PSU {}'
 PSU_INFO_PRESENCE_FIELD = 'presence'
 PSU_INFO_MODEL_FIELD = 'model'
 PSU_INFO_SERIAL_FIELD = 'serial'
+PSU_INFO_REV_FIELD = 'revision'
 PSU_INFO_STATUS_FIELD = 'status'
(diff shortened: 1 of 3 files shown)


liat-grozovik (approved):

liat-grozovik:
@robocoder99 is this OR pending on other PRs to be merged first? if so, can you please add?

robocoder99:
This PR can safely be merged without the Mellanox implementation (as it has to be tolerant of other vendor non-implemented status) it just won't do anything.

jleveque:
@robocoder99: Can you please also add a regression test in https://github.com/Azure/sonic-mgmt/blob/master/tests/platform_tests/api/test_psu.py to ensure vendors are aware they need to implement the new method for their platforms?

robocoder99:
Yes, that is in progress (done but in Mellanox internal review)

PR #183, opened 2021-05-12
[chassisd] Add script to initialize chassis info in STATE_DB

robocoder99:
#### Description
I added a new script `chassis_db_init` that uploads chassis hardware information such as serial number, model number and hardware revision to the STATE_DB under the CHASSIS_INFO table. 

#### Motivation and Context
I made this change in order to expose the chassis hardware information to SONiC user space and allow CLI utilities to access it in order to expose it to the user. 

Unit tests were added and was deployed and tested on Mellanox platform MSN2700.

### DIFF
--- a/sonic-chassisd/tests/test_chassis_db_init.py
+++ b/sonic-chassisd/tests/test_chassis_db_init.py
@@ -36,5 +36,3 @@ def test_provision_db():
     assert serial == fvs[CHASSIS_INFO_SERIAL_FIELD]
     assert model == fvs[CHASSIS_INFO_MODEL_FIELD]
     assert revision == fvs[CHASSIS_INFO_REV_FIELD]
-
-
--- a/sonic-chassisd/tests/test_chassis_db_init.py
+++ b/sonic-chassisd/tests/test_chassis_db_init.py
@@ -1,0 +1,40 @@
+import os
+import sys
+from imp import load_source
+
+from mock import Mock, MagicMock, patch
+from sonic_py_common import daemon_base
+
+from .mock_platform import MockChassis, MockModule
+from .mock_module_base import ModuleBase
+
+SYSLOG_IDENTIFIER = 'chassis_db_init_test'
+NOT_AVAILABLE = 'N/A'
+
+daemon_base.db_connect = MagicMock()
+
+test_path = os.path.dirname(os.path.abspath(__file__))
+modules_path = os.path.dirname(test_path)
+scripts_path = os.path.join(modules_path, "scripts")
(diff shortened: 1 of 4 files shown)


sonic-chassisd/tests/test_chassis_db_init.py
```diff
@@ -0,0 +1,40 @@
...
+    revision = "Rev C"
+
+    chassis_table = provision_db(chassis, log)
+
+    fvs = chassis_table.get(CHASSIS_INFO_KEY_TEMPLATE.format(1))
+    assert serial == fvs[CHASSIS_INFO_SERIAL_FIELD]
+    assert model == fvs[CHASSIS_INFO_MODEL_FIELD]
+    assert revision == fvs[CHASSIS_INFO_REV_FIELD]
+
+
```
jleveque, line 40: Delete trailing blank lines

jleveque (changes requested):

jleveque:
Also please update PR description, as it references `chassisshow`, which is not part of this PR.

jleveque:
Just curious why a separate script is necessary for this. Why can we not simply add this logic to chassisd when it starts up?

robocoder99:
We decided on a separate script because chassisd is primarily for handling modular chassis management (and thus is only run on modular chassis platforms), a separate task from uploading basic information such as serial, model, and hardware revision that is applicable to all platforms and also one-shot.

jleveque (approved):

jleveque:
Since this script is a one-shot script, and not a daemon per se, I have reservations about adding it in this repo. We could simply add it in the buildimage repo as part of https://github.com/Azure/sonic-buildimage/pull/7596. However, it would not be built into a package, so we would not be able to run unit tests against it. I'm on the fence, but I guess I can live with having it as part of the chassisd package.

robocoder99:
I would argue that while the name `sonic-platform-daemons` does indicate a one-shot task wouldn't belong, architecturally it does belong here as it performs the same fundamental task as the chassisd daemon (uploading chassis information to the redis db). As such anyone looking for this code would likely expect to find it here and I think that is valuable. 

If there exists now or in the future a collection of one-shot tasks for the purpose of provisioning redis then I would argue this should be refactored at that point.

jleveque:
> architecturally it does belong here as it performs the same fundamental task as the chassisd daemon (uploading chassis information to the redis db).

I agree with this, which is why I am on the fence.

> If there exists now or in the future a collection of one-shot tasks for the purpose of provisioning redis then I would argue this should be refactored at that point.

Agreed. We can proceed with this PR.

PR #189, opened 2021-05-31
Refactor Pcied and add unittest

sujinmkang:
#### Description

Refactor the pcied and add the unit test
#### Motivation and Context

Added unit test to increase the pmon unit test coverage.

Build with unit test enabled and run manually on a dut to verify the pcied.


sonic-pcied/scripts/pcied
```diff
@@ -5,31 +5,66 @@
...
+platform_pcieutil = None
+
+exit_code = 0
+
+# wrapper functions to call the platform api
+def _wrapper_get_pcie_check():
+    return platform_pcieutil.get_pcie_check()
+
+def _wrapper_get_pcie_aer_stats(bus, dev, func):
+    return platform_pcieutil.get_pcie_aer_stats(bus=bus, device=dev, func=func)
```
jleveque, line 47: It seems like these wrappers are unnecessary because you aren't checking for old or new API. The same function is being called for both, so these seem superfluous.

jleveque (changes requested):

sonic-pcied/scripts/pcied
```diff
@@ -39,142 +71,145 @@ class DaemonPcied(daemon_base.DaemonBase):
...
-        for item in resultInfo:
-            if item["result"] == "Failed":
-                self.log_warning("PCIe Device: " + item["name"] + " Not Found")
-                err += 1
+        Id = read_id_file(self.device_name)
 
+        self.aer_stats = {}
+        if Id is not None:
+            self.device_table.set(self.device_name, [('id', Id)])
+            self.aer_stats = platform_pcieutil.get_pcie_aer_stats(bus=Bus, dev=Dev, func=Fn)
```
ArunSaravananBalachandran, line 133: Please update the argument names as specified in https://github.com/Azure/sonic-platform-common/blob/master/sonic_platform_base/sonic_pcie/pcie_common.py#L102
```suggestion
            self.aer_stats = platform_pcieutil.get_pcie_aer_stats(bus=Bus, device=Dev, func=Fn)
```

sonic-pcied/scripts/pcied
```diff
@@ -134,7 +130,7 @@ class DaemonPcied(daemon_base.DaemonBase):
         self.aer_stats = {}
         if Id is not None:
             self.device_table.set(self.device_name, [('id', Id)])
-            self.aer_stats = _wrapper_get_pcie_aer_stats(bus=Bus, dev=Dev, func=Fn)
+            self.aer_stats = platform_pcieutil.get_pcie_aer_stats(bus=Bus, device=Dev, func=Fn)
```
jleveque, line 133: Why change from `dev` to `device`? It looks like the parameter is called `dev`: https://github.com/Azure/sonic-platform-common/blob/master/sonic_platform_base/sonic_pcie/pcie_base.py#L44

ArunSaravananBalachandran: It is observed that the parameter name is `dev` in the base class method defined in `pcie_base.py`, while it is overridden as `device` by the derived class method in `pcie_common.py` used by pcied (https://github.com/Azure/sonic-platform-common/blob/master/sonic_platform_base/sonic_pcie/pcie_common.py#L104) unifying them both would resolve the issue

sujinmkang: Updated the pcie_common.py with this PR : https://github.com/Azure/sonic-platform-common/pull/197
And revert the change in pcied.

sujinmkang:
@jleveque Can you please review and approve this PR? PR build passed now. Thanks!

jleveque (approved):

qiluo-msft:
This commit could not be cleanly cherry-pick to 202012. Please submit another PR.

qiluo-msft:
This PR could not be cleanly cherry-picked to 202012. Please submit another PR.

PR #193, opened 2021-06-10
[xcvrd] Force cleanup of chassis global variable on deinit

robocoder99:
#### Description
I added a `del` directive for the global `platform_chassis` object in `xcvrd` on deinit to ensure that the pointer is properly cleaned up such that the `__del__()` method is successfully called on the chassis object which closes its connection to SAI. 

#### Motivation and Context
On Mellanox platforms on 202012 we are seeing that in some cases xcvrd is maintaining the socket to the API past its deconstruction causing an error from SAI. This was traced by reproducing the issue and identifying that the pmon container was causing this issue where xcvrd is the only daemon within pmon that opens a SDK socket on Mellanox. 

Error in log...
```
syncd#SDK: [SX_API_INTERNAL.ERR] Failed command read at communication channel: Connection reset by peer
```

Change was uploaded to xcvrd on running pmon container on a switch running the latest 202012 build. xcvrd was then restarted with supervisord and we verified that we could no longer reproduce the bug. 

#### Backporting

**Backport to 202012** while this is a necessary fix in general we are seeing symptoms specifically on 202012

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1336,7 +1336,6 @@ class DaemonXcvrd(daemon_base.DaemonBase):
 
     # Deinitialize daemon
     def deinit(self):
-        global platform_chassis
         self.log_info("Start daemon deinit...")
 
         # Delete all the information from DB and then exit
@@ -1354,7 +1353,7 @@ class DaemonXcvrd(daemon_base.DaemonBase):
         if self.y_cable_presence[0] is True:
             y_cable_helper.delete_ports_status_for_y_cable()
 
-        del platform_chassis
+        del globals()['platform_chassis']
 
     # Run daemon
 
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1336,6 +1336,7 @@ class DaemonXcvrd(daemon_base.DaemonBase):
 
     # Deinitialize daemon
     def deinit(self):
+        global platform_chassis
         self.log_info("Start daemon deinit...")
 
         # Delete all the information from DB and then exit
@@ -1352,6 +1353,8 @@ class DaemonXcvrd(daemon_base.DaemonBase):
 
         if self.y_cable_presence[0] is True:
             y_cable_helper.delete_ports_status_for_y_cable()
+
+        del platform_chassis
 
     # Run daemon
 


liat-grozovik:
@robocoder99 can you please handle LGTM alerts?

liat-grozovik (approved):

liat-grozovik:
regarding the new alert from LGTM @robocoder99 as the fix is to delete the pointers, not sure why LGTM raise a concern about this specific line which is the whole reason for this PR. Maybe something is missing. Can you please check it out?

liat-grozovik:
@keboliu  and @jleveque kindly reminder to review this enhancement .

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1353,6 +1354,8 @@ def deinit(self):
         if self.y_cable_presence[0] is True:
             y_cable_helper.delete_ports_status_for_y_cable()
 
+        del platform_chassis
```
jleveque, line 1357: To fix LGTM alert, which seems to be incorrect (it thinks this is a local variable, when it is not), suggest trying to replace with:

```python
del globals()['platform_chassis']

```

jleveque (changes requested):

sonic-xcvrd/xcvrd/xcvrd.py
robocoder99: I am hesitant to move to a less readable form of the code due to a bug in the static analyzer. Is there no way to either accept that it is wrong in this case and merge anyway or ask it to ignore this specific case?

jleveque: I'm not a big fan of adding comments for LGTM to ignore lines, but you can look into what they have available and see if you can find something that works, then we can discuss.

robocoder99: I understand, I wouldn't want to add a line to the codebase for it either. Can we not accept the warning and merge nonetheless? 

I feel like that is the most elegant solution.

jleveque: While it may be the most elegant solution, it will degrade the LGTM code quality grade, which is currently an A+. I'd rather not introduce false alarms into the code quality score.

While my initial suggestion my not be as readable, I think it is the best compromise. I prefer it to adding comments purely for the LGTM analyzer.

robocoder99: Implemented and tested your suggestion @jleveque

jleveque (approved):

dprital:
@daall Can you please merge to 202012 ?

PR #194, opened 2021-06-15
[CI] sonic-config-engine now depends on SONiC YANG packages

jleveque:
The Python 3 version of sonic-config-engine now depends on sonic-yang-mgmt and sonic-yang-models, so we now need to install them as part of continuous integration to get the CI working again


qiluo-msft (approved):

qiluo-msft:
Only the python3 version of "sonic-config-engine now depends on sonic-yang-mgmt and sonic-yang-models". Could you update the PR description?

PR #197, opened 2021-06-28
[xcvrd] initial support for integrating vendor specfic class objects for calling Y-Cable API's inside xcvrd

vdahiya12:

#### Description
This PR integrates vendor specific class objects inside xcvrd for Y-Cable API's to be called. 
Detailed designed document can be found https://github.com/Azure/SONiC/pull/757

 

#### Motivation and Context
Basically xcvrd now has  an interface for Y-Cable to interact with PMON docker and HOST
which can be uniform across all vendors. As part of this refactor, we will be moving towards a model where only xcvrd interacts with the cables/transceivers,  and host-side processes will communicate with xcvrd rather than with the devices directly with Y-Cable. 

Ran the changes on Arista7050cx3 switch, making changes inside the container.

### DIFF
--- a/sonic-xcvrd/tests/test_xcvrd.py
+++ b/sonic-xcvrd/tests/test_xcvrd.py
@@ -354,7 +354,9 @@ class TestXcvrdScript(object):
         mux_info_dict['version_nic_active'] = '0.8'
         mux_info_dict['version_nic_inactive'] = '0.7'
         mux_info_dict['version_nic_next'] = '0.7'
-        set_show_firmware_fields("Ethernet0", mux_info_dict, xcvrd_show_fw_res_tbl)
+        rc = set_show_firmware_fields("Ethernet0", mux_info_dict, xcvrd_show_fw_res_tbl)
+
+        assert(rc == 0)
 
     def test_get_media_settings_key(self):
         xcvr_info_dict = {
--- a/sonic-xcvrd/tests/test_xcvrd.py
+++ b/sonic-xcvrd/tests/test_xcvrd.py
@@ -16,6 +16,10 @@ from .mock_swsscommon import Table
 daemon_base.db_connect = MagicMock()
 swsscommon.Table = MagicMock()
 swsscommon.ProducerStateTable = MagicMock()
+swsscommon.SubscriberStateTable = MagicMock()
+swsscommon.SonicDBConfig = MagicMock()
+#swsscommon.Select = MagicMock()
+
 sys.modules['sonic_y_cable'] = MagicMock()
 sys.modules['sonic_y_cable.y_cable'] = MagicMock()
 
@@ -217,7 +221,7 @@ class TestXcvrdScript(object):
 
     @patch('xcvrd.xcvrd_utilities.y_cable_helper.y_cable_platform_sfputil', MagicMock(return_value=[0]))
     @patch('xcvrd.xcvrd_utilities.y_cable_helper.logical_port_name_to_physical_port_list', MagicMock(return_value=[0]))
-    @patch('xcvrd.xcvrd_utilities.y_cable_helper._wrapper_get_presence', MagicMock(return_value=True))
+    @patch('xcvrd.xcvrd_utilities.y_cable_helper.y_cable_wrapper_get_presence', MagicMock(return_value=True))
     @patch('xcvrd.xcvrd_utilities.y_cable_helper.get_muxcable_info', MagicMock(return_value={'tor_active': 'self',
                                                                        'mux_direction': 'self',
                                                                        'manual_switch_count': '7',
@@ -254,9 +258,10 @@ class TestXcvrdScript(object):
         rc = post_port_mux_info_to_db(logical_port_name, mux_tbl)
         assert(rc != -1)
 
+
     @patch('xcvrd.xcvrd_utilities.y_cable_helper.y_cable_platform_sfputil', MagicMock(return_value=[0]))
     @patch('xcvrd.xcvrd_utilities.y_cable_helper.logical_port_name_to_physical_port_list', MagicMock(return_value=[0]))
-    @patch('xcvrd.xcvrd_utilities.y_cable_helper._wrapper_get_presence', MagicMock(return_value=True))
+    @patch('xcvrd.xcvrd_utilities.y_cable_helper.y_cable_wrapper_get_presence', MagicMock(return_value=True))
     @patch('xcvrd.xcvrd_utilities.y_cable_helper.get_muxcable_static_info', MagicMock(return_value={'read_side': 'self',
                                                                               'nic_lane1_precursor1': '1',
                                                                               'nic_lane1_precursor2': '-7',
@@ -293,6 +298,63 @@ class TestXcvrdScript(object):
         mux_tbl = Table("STATE_DB", y_cable_helper.MUX_CABLE_STATIC_INFO_TABLE)
         rc = post_port_mux_static_info_to_db(logical_port_name, mux_tbl)
         assert(rc != -1)
+
+    def test_y_cable_helper_format_mapping_identifier1(self):
+        rc = format_mapping_identifier("ABC        ")
+        assert(rc == "abc")
+
+    def test_y_cable_wrapper_get_transceiver_info(self):
+        with patch('xcvrd.xcvrd_utilities.y_cable_helper.y_cable_platform_sfputil') as patched_util:
+            patched_util.get_transceiver_info_dict.return_value = {'manufacturer': 'Microsoft',
+                                                                              'model': 'model1'}
+
+            transceiver_dict = y_cable_wrapper_get_transceiver_info(1)
+            vendor = transceiver_dict.get('manufacturer')
+            model = transceiver_dict.get('model')
+
+        assert(vendor == "Microsoft")
+        assert(model == "model1")
+
+    def test_y_cable_wrapper_get_presence(self):
+        with patch('xcvrd.xcvrd_utilities.y_cable_helper.y_cable_platform_sfputil') as patched_util:
+            patched_util.get_presence.return_value = True
+
+            presence = y_cable_wrapper_get_presence(1)
+
+        assert(presence == True)
+
+    @patch('xcvrd.xcvrd_utilities.y_cable_helper.logical_port_name_to_physical_port_list', MagicMock(return_value=[0]))
+    @patch('xcvrd.xcvrd_utilities.y_cable_helper.y_cable_wrapper_get_presence', MagicMock(return_value=True))
+    def test_get_ycable_physical_port_from_logical_port(self):
+
+        instance = get_ycable_physical_port_from_logical_port("Ethernet0")
+
+        assert(instance == 0)
+
+    @patch('xcvrd.xcvrd_utilities.y_cable_helper.logical_port_name_to_physical_port_list', MagicMock(return_value=[0]))
+    @patch('xcvrd.xcvrd_utilities.y_cable_helper.y_cable_wrapper_get_presence', MagicMock(return_value=True))
+    def test_get_ycable_port_instance_from_logical_port(self):
+
+        with patch('xcvrd.xcvrd_utilities.y_cable_helper.y_cable_port_instances') as patched_util:
+            patched_util.get.return_value = 0
+            instance = get_ycable_port_instance_from_logical_port("Ethernet0")
+
+        assert(instance == 0)
+
+    def test_set_show_firmware_fields(self):
+
+        mux_info_dict = {}
+        xcvrd_show_fw_res_tbl = Table("STATE_DB", "XCVRD_SHOW_FW_RES")
+        mux_info_dict['version_self_active'] = '0.8'
+        mux_info_dict['version_self_inactive'] = '0.7'
+        mux_info_dict['version_self_next'] = '0.7'
+        mux_info_dict['version_peer_active'] = '0.8'
+        mux_info_dict['version_peer_inactive'] = '0.7'
+        mux_info_dict['version_peer_next'] = '0.7'
+        mux_info_dict['version_nic_active'] = '0.8'
+        mux_info_dict['version_nic_inactive'] = '0.7'
+        mux_info_dict['version_nic_next'] = '0.7'
+        set_show_firmware_fields("Ethernet0", mux_info_dict, xcvrd_show_fw_res_tbl)
 
     def test_get_media_settings_key(self):
         xcvr_info_dict = {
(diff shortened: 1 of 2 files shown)


vdahiya12:
dependent on https://github.com/Azure/sonic-platform-common/pull/203

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -271,37 +323,93 @@ def check_identifier_presence_and_update_mux_table_entry(state_db, port_tbl, y_c
...
+
+                            module = module_dir.get(model)
+                            if module is None:
+                                helper_logger.log_warning(
+                                    "Error: Unable to find module name from model {}".format(logical_port_name))
+                                return
+
+                            attr_name = 'sonic_y_cable.' + module
+                            yc = getattr(import_module(attr_name), 'YCable')
+                            y_cable_port_instances[physical_port] = yc(physical_port, "y-cable_%d" % physical_port, helper_logger)
```
jleveque, line 359: Why pass the helper_logger? I believe we decided that the YCableBase class now inherits from the Logger class.

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -271,37 +323,93 @@ def check_identifier_presence_and_update_mux_table_entry(state_db, port_tbl, y_c
...
+
+                if len(physical_port_list) == 1:
+
+                    physical_port = physical_port_list[0]
+                    if _wrapper_get_presence(physical_port):
+                        port_info_dict = _wrapper_get_transceiver_info(
+                            physical_port)
+                        if port_info_dict is not None:
+                            vendor = port_info_dict['manufacturer'].rstrip()
+                            model = port_info_dict['model'].rstrip()
```
jleveque, line 343: Do we just plan to use these strings as-is, or do we plan to do any standardization (e.g., change case, replace spaces with underscores, etc.)?

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -1103,7 +1268,7 @@ def task_worker(self):
                             new_status = 'unknown'
 
                         fvs_updated = swsscommon.FieldValuePairs([('state', new_status),
-                                                                  ('read_side', read_side),
+                                                                  ('read_side',read_side),
```
jleveque, line 1271: Add space back.

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -493,12 +610,20 @@ def delete_ports_status_for_y_cable():
...
-            delete_port_from_y_cable_table(
-                logical_port_name, mux_tbl[asic_index])
+            delete_port_from_y_cable_table(logical_port_name, y_cable_tbl[asic_index])
+            delete_port_from_y_cable_table(logical_port_name, static_tbl[asic_index])
+            delete_port_from_y_cable_table(logical_port_name, mux_tbl[asic_index])
+            # delete the y_cable port instance
+            physical_port_list = logical_port_name_to_physical_port_list(logical_port_name)
+
+            if len(physical_port_list) == 1:
+
```
jleveque, line 620: Suggest removing this blank line to match with the `else` case.

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -325,32 +433,38 @@ def check_identifier_presence_and_delete_mux_table_entry(state_db, port_tbl, asi
...
-                    logical_port_name, mux_tbl[asic_index])
+                delete_port_from_y_cable_table(logical_port_name, y_cable_tbl[asic_index])
+                delete_port_from_y_cable_table(logical_port_name, static_tbl[asic_index])
+                delete_port_from_y_cable_table(logical_port_name, mux_tbl[asic_index])
                 delete_change_event[:] = [True]
+                # delete the y_cable instance
+                physical_port_list = logical_port_name_to_physical_port_list(logical_port_name)
+
+                if len(physical_port_list) == 1:
+
```
jleveque, line 455: Suggest removing this blank line to match with the `else` case.

jleveque (changes requested):

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -42,6 +43,25 @@
 MUX_CABLE_STATIC_INFO_TABLE = "MUX_CABLE_STATIC_INFO"
 MUX_CABLE_INFO_TABLE = "MUX_CABLE_INFO"
 
+def format_the_string(string):
```
jleveque, line 46: Suggest a more descriptive name, like `format_mapping_identifier()`

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -42,6 +43,25 @@
...
+    """
+
+    # create a working copy (and make it lowercase, while we're at it)
+    s = string.lower()
+
+    # remove leading and trailing whitespace
+    s = s.strip()
+
+    # Make spaces into underscores
+    s = re.sub('[\\s\\t\\n]+', '_', s)
```
jleveque, line 61: Do you need `\t` and `\n`? I believe `\s` should cover all whitespace.

Also suggest using a raw string for the pattern, which eliminates the need for escaping the backslashes. E.g., `s = re.sub(r'\s+', '_', s)`

Also suggest changing comment to `# Replace whitespace with underscores`

jleveque (changes requested):

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
vdahiya12: fixed

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
vdahiya12: fixed

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
vdahiya12: fixed

jleveque (approved):

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -271,37 +426,110 @@ def check_identifier_presence_and_update_mux_table_entry(state_db, port_tbl, y_c
...
+                        if port_info_dict is not None:
+                            vendor = port_info_dict.get('manufacturer')
+
+                            if vendor is None:
+                                helper_logger.log_warning(
+                                    "Error: Unable to find Vendor name for Transceiver for Y-Cable initiation {}".format(logical_port_name))
+
+                            model = port_info_dict.get('model')
+
+                            if vendor is None:
```
wangxin, line 453: `if model is None:` ?

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -58,6 +100,85 @@ def _wrapper_get_presence(physical_port):
     return y_cable_platform_sfputil.get_presence(physical_port)
 
 
+def _wrapper_get_transceiver_info(physical_port):
+    if y_cable_platform_chassis is not None:
+        try:
+            return y_cable_platform_chassis.get_sfp(physical_port).get_transceiver_info()
```
wangxin, line 106: For simulated y-cable, the physical cable connected to dual ToR is none-y-cable. But we need xcvrd to initialize instances of simulated y-cable class. To workaround this, we may need to patch this file for testbeds using simulated y-cable. Assume we have simulated y cable at location `sonic-y-cable/microsoft/y_cable_simulated.py`, then can we patch here like below before running tests:
```suggestion
            xcvr_info = y_cable_platform_chassis.get_sfp(physical_port).get_transceiver_info()
            xcvr_info['vendor'] = 'microsoft'
            xcvr_info['model'] = 'simulated'
            return xcvr_info
```

Will this cause any negative impact to the existing xcvr daemon?

vdahiya12: Would this patch be enabled when running tests or would it be for all cases. 
I think this patch should only be enabled when their is mux_simulator injection code is enabled.

lguohan:
can you add unit test accordingly?

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
wangxin: This patch will only be injected for dualtor testbed using mux-simulator.

vdahiya12: Yeah, i think this should work then.

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
vdahiya12: fixed

sonic-xcvrd/tests/test_xcvrd.py
```diff
@@ -294,6 +299,63 @@ def test_post_port_mux_static_info_to_db(self):
...
+        mux_info_dict['version_self_active'] = '0.8'
+        mux_info_dict['version_self_inactive'] = '0.7'
+        mux_info_dict['version_self_next'] = '0.7'
+        mux_info_dict['version_peer_active'] = '0.8'
+        mux_info_dict['version_peer_inactive'] = '0.7'
+        mux_info_dict['version_peer_next'] = '0.7'
+        mux_info_dict['version_nic_active'] = '0.8'
+        mux_info_dict['version_nic_inactive'] = '0.7'
+        mux_info_dict['version_nic_next'] = '0.7'
+        set_show_firmware_fields("Ethernet0", mux_info_dict, xcvrd_show_fw_res_tbl)
```
tahmed-dev, line 357: What is being tested in this test case?

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -22,6 +26,10 @@
 
 helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
 
+y_cable_port_instances = {}
+y_cable_port_locks = {}
```
tahmed-dev, line 30: are those locks shared between CLI and applications?

vdahiya12: that's correct, moving to the abstract class model, there is a lock for each port. Each api to y_cable driver will lock it for that port and only release when the API is returned. Basically all y_cable api's will be called by xcvrd only. cli will call the y_cable api through redis and invoke it per port.

sonic-xcvrd/tests/test_xcvrd.py
vdahiya12: added an assert for return code
fixed

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
tahmed-dev: This means that presence of cli activity such show, could potentially affect our switching time, right?

vdahiya12: For the same port, Yes it will. Cli activities such as show hwmode, download firmware etc would slow these, but in rela production we dont anticipate to use these as such.

tahmed-dev (approved):

vdahiya12:
azp run

sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -58,6 +100,87 @@ def _wrapper_get_presence(physical_port):
     return y_cable_platform_sfputil.get_presence(physical_port)
 
 
+def y_cable_wrapper_get_transceiver_info(physical_port):
```
wangxin, line 103: @vdahiya12 I got a better idea of hooking the simulated y_cable driver.
For the simulated y_cable driver to work, we need to inject a configuration file like `/etc/sonic/mux_simulator.json`. The config file needs to supply information like IP/port of the mux simulator.

So, we can decorate this function like this:
```suggestion
def hook_y_cable_simulated(target):
    """Decorator to add hook for using the simulated y_cable driver.

    This decorator checks existence of the configuration file required by the simulated y_cable driver. If the
    configuration file is found, then override the "manufacturer" and "model" fields with value "microsoft" and 
    "simulated" in the collected transceiver info dict. Consequently, instance of the simulated y_cable driver 
    class will be initialized.

    When the configuration file is not found on system, then just return the original transceiver info to initialize
    instance of y_cable driver class of whatever actually plugged physical y_cable.

    For test systems using simulated y_cable, we can just inject the simulated y_cable driver config file then
    restart the pmon service before testing starts.

    Args:
        target (function): The function collecting transceiver info.
    """    
    MUX_SIMULATOR_CONFIG_FILE = "/etc/sonic/mux_simulator.json"
    VENDOR = "microsoft"
    MODEL = "simulated"
    def wrapper(*args, **kwargs):
        res = target(*args, **kwargs)
        if os.path.exists(MUX_SIMULATOR_CONFIG_FILE):
            res["manufacturer"] = VENDOR
            res["model"] = 
        return res
    wrapper.__name__ = target.__name__
    return wrapper


@hook_y_cable_simulated
def y_cable_wrapper_get_transceiver_info(physical_port):
```

When config file of simulated y_cable driver is found on system, then update the "manufacturer" and "model" in the collected transceiver info dict to "microsoft" and "simulated.

vdahiya12: added to code, fixed

vdahiya12: added in xcvrd, fixed

tahmed-dev (approved):

qiluo-msft:
This commit could not be cleanly cherry-pick to 202012. Please submit another PR.

PR #198, opened 2021-07-04
[pcied] Fix pcied failure to load due to 'pcied NameError: name 'self' is not defined'

DavidZagury:
#### Description

Fix https://github.com/Azure/sonic-buildimage/issues/7993

- Fixes wrong log line when Pcie module does not exists that causes the pcied daemon to enter FATAL state.
- Change log level of line "Failed to load platform Pcie module" to notice since this is not an error flow, this is part of the normal loading flow we expect when a vendor didn't supply a Pcie class.

Load image, check pcied daemon status.

### DIFF
--- a/sonic-pcied/scripts/pcied
+++ b/sonic-pcied/scripts/pcied
@@ -10,7 +10,7 @@ import signal
 import sys
 import threading
 
-from sonic_py_common import daemon_base, device_info
+from sonic_py_common import daemon_base, device_info, logger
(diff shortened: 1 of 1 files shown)


BasimShalata:
@sujinmkang can you please review.

liat-grozovik (approved):

lguohan (approved):
:shipit:

lguohan (approved):

PR #200, opened 2021-07-08
Add str conversion for numeric xcvr info fields to xcvrd

andywongarista:
#### Description

Converted nominal_bit_rate and cable_length fields from the sfp transceiver info dict into str in post_port_sfp_info_to_db

#### Motivation and Context

The documentation for get_transceiver_info in SfpBase indicates that these field values should be integers. However, only string values can be passed into swsscommon.FieldValuePairs. Therefore, a string conversion is needed. 

Tested on an Arista platform that xcvrd runs without errors when nominal_bit_rate and cable_length are integers.


lguohan:
is this a bug fix?

andywongarista:
> is this a bug fix?

Yes. I believe it's gone unnoticed because most platforms end up returning these fields as string (rather than int) and the tests don't verify the types of these fields.

prgeor (approved):

PR #201, opened 2021-07-09
Introduce mgmtinit delay after transceiver reset

prgeor:
Signed-off-by: Prince George <prgeor@microsoft.com>

#### Description

Added delay of 2 seconds for QSFP's MgmtInit after module presence is detected

#### Motivation and Context
As per CMIS spec : http://www.qsfp-dd.com/wp-content/uploads/2021/05/CMIS5p0.pdf (see D.1.3 Software Configuration and Initialization) module's resetL must be de-asserted and a delay of 2 seconds is required for MgmtInit to complete.

TBD
Verify the eeprom dump and link comes UP under following scenarios:

1. Multiple reboots with optics from different vendors across various ports
2. OIR same optics multiple times on same port
3. Swap optics from different vendors on same port
4. Repeat 2 and 3 across multiple ports

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -39,6 +39,9 @@ TRANSCEIVER_STATUS_TABLE = 'TRANSCEIVER_STATUS'
 
 SELECT_TIMEOUT_MSECS = 1000
 
+# Mgminit time required as per CMIS spec
+MGMT_INIT_TIME_DELAY_SECS = 2
+
 DOM_INFO_UPDATE_PERIOD_SECS = 60
 TIME_FOR_SFP_READY_SECS = 1
 XCVRD_MAIN_THREAD_SLEEP_SECS = 60
@@ -201,8 +204,16 @@ def _wrapper_get_sfp_error_description(physical_port):
         except NotImplementedError:
             pass
     return None
-# Remove unnecessary unit from the raw data
 
+def _wrapper_reset_transceiver(physical_port):
+    if platform_chassis:
+        try:
+            return platform_chassis.get_sfp(physical_port).reset()
+        except NotImplementedError:
+            pass
+    return platform_sfputil.reset(physical_port)
+
+# Remove unnecessary unit from the raw data
 
 def beautify_dom_info_dict(dom_info_dict, physical_port):
     dom_info_dict['temperature'] = strip_unit_and_beautify(dom_info_dict['temperature'], TEMP_UNIT)
@@ -260,8 +271,19 @@ def beautify_dom_threshold_info_dict(dom_info_dict):
     dom_info_dict['txbiashighwarning'] = strip_unit_and_beautify(dom_info_dict['txbiashighwarning'], BIAS_UNIT)
     dom_info_dict['txbiaslowwarning'] = strip_unit_and_beautify(dom_info_dict['txbiaslowwarning'], BIAS_UNIT)
 
-# Update port sfp info in db
+# Reset/re-initialize the transciever to its default state
 
+def reset_transceiver(logical_port_name):
+    physical_port_list = logical_port_name_to_physical_port_list(logical_port_name)
+    if physical_port_list is None:
+        helper_logger.log_error("No physical ports found for logical port '{}'".format(logical_port_name))
+        return
+
+    for physical_port in physical_port_list:
+        if _wrapper_reset_transceiver(physical_port):
+            time.sleep(MGMT_INIT_TIME_DELAY_SECS)
+
+# Update port sfp info in db
 
 def post_port_sfp_info_to_db(logical_port_name, table, transceiver_dict,
                              stop_event=threading.Event()):
@@ -1082,6 +1104,7 @@ class SfpStateUpdateTask(object):
 
                             if value == sfp_status_helper.SFP_STATUS_INSERTED:
                                 helper_logger.log_info("Got SFP inserted event")
+                                reset_transceiver(logical_port)
                                 # A plugin event will clear the error state.
                                 update_port_transceiver_status_table(
                                     logical_port, status_tbl[asic_index], sfp_status_helper.SFP_STATUS_INSERTED)
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -905,6 +905,7 @@ class SfpStateUpdateTask(object):
     def __init__(self):
         self.task_process = None
         self.task_stopping_event = multiprocessing.Event()
+        self.sfp_insert_events = {}
 
     def _mapping_event_from_change_event(self, status, port_dict):
         """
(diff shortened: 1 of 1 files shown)


sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -260,8 +271,19 @@ def beautify_dom_threshold_info_dict(dom_info_dict):
...
-# Update port sfp info in db
+# Reset/re-initialize the transciever to its default state
 
+def reset_transceiver(logical_port_name):
+    physical_port_list = logical_port_name_to_physical_port_list(logical_port_name)
+    if physical_port_list is None:
+        helper_logger.log_error("No physical ports found for logical port '{}'".format(logical_port_name))
+        return
+
+    for physical_port in physical_port_list:
```
aravindmani-1, line 282: Module reset is not required upon module insertion. Can we use sleep instead?.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -201,8 +204,16 @@ def _wrapper_get_sfp_error_description(physical_port):
...
     return None
-# Remove unnecessary unit from the raw data
 
+def _wrapper_reset_transceiver(physical_port):
+    if platform_chassis:
+        try:
+            return platform_chassis.get_sfp(physical_port).reset()
+        except NotImplementedError:
+            pass
+    return platform_sfputil.reset(physical_port)
```
lguohan, line 214: why do we need to reset the transceiver after insersion?

prgeor: This is required for all QSFP28/QSFP-DD modules that follow CMIS spec so that host can deterministically start accessing the I2C interface after 2 seconds delay of de-asserting the resetL.

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: can you check CMIS spec : http://www.qsfp-dd.com/wp-content/uploads/2021/05/CMIS5p0.pdf (see D.1.3 Software Configuration and Initialization)

sonic-xcvrd/xcvrd/xcvrd.py
lguohan: resetL means put the module out of reset, I thought that is the default behavior, this reset api is to first set the reset pin to high, and then to low. therefore, i am not sure you need this operation.

prgeor: as per spec, its the host which is responsible for bringing the module out of reset (even though it *may be* done automatically in hw), hence I think its better host software take explicit control of the module initialization rather than believing it will be done by hw. Do you see any problem?

lguohan: if you really want to make sure the module is out of reset, i suggest to add additional platform API to set resetL. In that case, if the module is already out of reset, then the api will do nothing. otherwise, the api will bring the module out of reset. current reset API is doing resetH, and then resetL which is not what the spec says.

sonic-xcvrd/xcvrd/xcvrd.py
aravindmani-1: Also, this logic would introduce 8 seconds sleep for a single 4x100G.
For e.g. If Ethernet0,2,4,6 is part of single physical port(1), delay will be executed for each port(0/2/4/6).

prgeor: yes, you are right

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: Ok, if platform is taking care, removing the reset

lguohan (approved):

sonic-xcvrd/xcvrd/xcvrd.py
mikeberesford: On top of this, the insertion events are processed serially, so even if there was only one sleep done per physical port on a 32-port system this would result in over a minute of delay at startup, even though the ports were all "inserted" at the same time, right?

lguohan: @prgeor , can you check the boot up sequence, will this introduce cumulative delay?

prgeor: yes, if there are 32 optics populated in a 32-ports system then the total delay will be 32 * 2 secs = 64 secs.

lguohan: can you then fix this properly?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1082,6 +1084,8 @@ def task_worker(self, stopping_event, sfp_error_event, y_cable_presence):
 
                             if value == sfp_status_helper.SFP_STATUS_INSERTED:
                                 helper_logger.log_info("Got SFP inserted event")
+                                # Delay for I2C MgmtInit to complete
+                                time.sleep(MGMT_INIT_TIME_DELAY_SECS)
```
lguohan, line 1088: as we found in the other thread, let's not sleep here. let's put them into a list, and later look the list to get the data. let's fix this properly.

lguohan (changes requested):
please fix this properly.

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: reset_transceiver() is already removed

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: I didn't understand

prgeor: Removed the cumulative delay across port insertions

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: Removed the cumulative delay across port insertions

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1082,6 +1106,8 @@ def task_worker(self, stopping_event, sfp_error_event, y_cable_presence):
...
```
prgeor, line 1110: Not required. To be removed.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1020,7 +1039,12 @@ def task_worker(self, stopping_event, sfp_error_event, y_cable_presence):
         while not stopping_event.is_set():
             next_state = state
             time_start = time.time()
+            # Ensure not to block for any event if sfp insert event is pending
+            if self.sfp_insert_events:
+                timeout = 1
```
prgeor, line 1044: check timeout in msec or sec. its too small

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -928,6 +931,22 @@ def _mapping_event_from_change_event(self, status, port_dict):
         helper_logger.log_debug("mapping from {} {} to {}".format(status, port_dict, event))
         return event
 
+    # Soak SFP insert event until management init completes
+    def _soak_sfp_insert_event(self, port_dict):
```
prgeor, line 935: can this be moved to a wrapper function?

sonic-xcvrd/xcvrd/xcvrd.py
aravindmani-1: As per https://github.com/Azure/sonic-platform-common/blob/4533f829781f9d3706406f33d82a58adef74d294/sonic_platform_base/chassis_base.py#L553 definition, timeout should be in milliseconds.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -928,6 +931,22 @@ def _mapping_event_from_change_event(self, status, port_dict):
...
+        self.sfp_insert_events = {}
```
aravindmani-1, line 936: can you please check get_change_event for all the vendors?. Incase empty port dict is returned, timeout will not work properly. In this case, do we have to re-initialize sfp_insert_events?.

prgeor: insertion time is updated only once when module is inserted (the event is then deleted from port_dict)

aravindmani-1 (approved):

lguohan (approved):

lguohan:
@prgeor , please continue to add unit test.

PR #203, opened 2021-07-14
Fix a bug at removing asics on not-online cards

ngoc-do:
Fix a bug coming from https://github.com/Azure/sonic-platform-daemons/pull/175 when asic is removed due to card getting offline (before it was online).

The change works on real hw + unit test.

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -248,6 +248,8 @@ class ModuleUpdater(logger.Logger):
         asics = list(self.asic_table.getKeys())
         for asic in asics:
            fvs = self.asic_table.get(asic)
+           if isinstance(fvs, list):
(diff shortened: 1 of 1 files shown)


ngoc-do:
Add @mprabhu-nokia for review.

mprabhu-nokia (approved):
Did you get to check if some underlying implementation changed because of which the table.get() returns a list instead of tuple?

mprabhu-nokia:
@judyjoseph Could you please review?

ngoc-do:
> Did you get to check if some underlying implementation changed because of which the table.get() returns a list instead of tuple?

I didn't have chance to look at if it's because of table implementation change since I fixed on our code base that recently got upgraded.

judyjoseph (approved):

PR #208, opened 2021-07-27
[y-cable] fix for logging the xcvrd metrics before writing the state to the State-DB

vdahiya12:

#### Description

This PR fixes the logging for mux_metrics for writing anything to the state DB for a state transition. Basically previously xcvrd would first post the query result to the DB and only then update its own metric for state transition, but now we do the reverse.

#### Motivation and Context

Second change was motivated by the sequence of events which happen inside state transition for mux-metrics table. Previously orchagent reports finishing the transition before xcvrd which is not the case. Ideally xcvrd does the transition first followed by orchagent followed by linkmgr

Ran the change on Arista7050cx3 testbed.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
+++ b/sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
@@ -15,8 +15,6 @@ from . import sfp_status_helper
 
 SELECT_TIMEOUT = 1000
 
-Y_CABLE_DIRECTION_SYNC = 3
-
 y_cable_platform_sfputil = None
 y_cable_platform_chassis = None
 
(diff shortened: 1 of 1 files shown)


sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -122,6 +124,15 @@ def update_tor_active_side(read_side, state, logical_port_name):
                     return y_cable_toggle_mux_torA(physical_port)
 
             # TODO: Should we confirm that the mux was indeed toggled?
+            # attempt to read the cable direction in 3 attempts and if still the direction is not in sync
+            # return unknown
+            for i in range(Y_CABLE_DIRECTION_SYNC):
```
tahmed-dev, line 129: how would that work if the above calls are returning already?

vdahiya12: fix

tahmed-dev (approved):

PR #216, opened 2021-09-28
[xcvrd][y_cable] refactor xcvrd to listen to port probe without locks; fix the get_firmware_version API to sync with download_firmware

vdahiya12:

#### Description
This PR refactors the listening of port probes from linkmgr to xcvrd as so that xcvrd does not see timeout messages like these below

`Sep 27 22:24:37.490921 SONIC WARNING mux#linkmgrd: link_manager/LinkManagerStateMachine.cpp:815 handleMuxWaitTimeout: Ethernet120: xcvrd timed out responding to linkmgrd, current state: (P: Unknown, M: Wait, L: Down)`

Refactoring to a separate listener thread mode as well as removing the lock(safe because read/write transaction via eeprom for port probe) seems to handle this, and does not have the logs like this one above.
as the read write eeprom also has a mutex lock in the platform api call, optoe implementation.
https://github.com/Azure/sonic-linux-kernel/blob/889d76e36b3f012d3782a1c5e1587c32e4d1ed11/patch/driver-support-optoe.patch#L717

#### Motivation and Context
refactor xcvrd for listening to port probes, disable calling the get_firmware_api() while a download is in progress. 

Ran the changes on Arista7050cx3-32s testbed

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
+++ b/sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
@@ -238,17 +238,16 @@ def y_cable_toggle_mux_torA(physical_port):
     port_instance = y_cable_port_instances.get(physical_port)
     if port_instance is None:
         helper_logger.log_error(
-            "Error: Could not get port instance for read side for  Y cable port {}".format(physical_port))
+            "Error: Could not get port instance for read side for  Y cable port {} {}".format(physical_port, threading.currentThread().getName()))
         return -1
 
-    with y_cable_port_locks[physical_port]:
-        try:
-            update_status = port_instance.toggle_mux_to_tor_a()
-        except Exception as e:
-            update_status = -1
-            helper_logger.log_warning("Failed to execute the toggle mux ToR A API for port {} due to {}".format(physical_port,repr(e)))
+    try:
+        update_status = port_instance.toggle_mux_to_tor_a()
+    except Exception as e:
+        update_status = -1
+        helper_logger.log_warning("Failed to execute the toggle mux ToR A API for port {} due to {} {}".format(physical_port, repr(e) , threading.currentThread().getName()))
 
-    helper_logger.log_debug("Y_CABLE_DEBUG: Status of toggling mux to ToR A for port {} {}".format(physical_port, update_status))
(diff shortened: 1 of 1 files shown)


sonic-xcvrd/xcvrd/xcvrd_utilities/y_cable_helper.py
```diff
@@ -238,17 +238,16 @@ def y_cable_toggle_mux_torA(physical_port):
     port_instance = y_cable_port_instances.get(physical_port)
     if port_instance is None:
         helper_logger.log_error(
-            "Error: Could not get port instance for read side for  Y cable port {}".format(physical_port))
+            "Error: Could not get port instance for read side for  Y cable port {} {}".format(physical_port, threading.currentThread().getName()))
         return -1
 
-    with y_cable_port_locks[physical_port]:
-        try:
-            update_status = port_instance.toggle_mux_to_tor_a()
```
theasianpianist, line 246: toggling the mux is a write operation, would we still need the lock for this?

vdahiya12: This should be fine, as long as no firmware API's are getting called in sync we are good. 
as the read write eeprom also has a mutex lock in the platform api call, optoe implementation. 
https://github.com/Azure/sonic-linux-kernel/blob/889d76e36b3f012d3782a1c5e1587c32e4d1ed11/patch/driver-support-optoe.patch#L717

So theoretically we can get rid of all the locking. 
the only problem is firmware API's have a parallelism problem(in the firmware confirmed with vendor)
so that's why we have the locks still in place.

theasianpianist: Will this vary between different cable vendors?

vdahiya12: Both the vendors have the same ask currently.

theasianpianist: But would it be possible for a vendor to not include the mutex in their own implementation, which means we would need it here?

vdahiya12: mutex is provided by optoe layer, all vendors go through that read/write eeprom API

theasianpianist (approved):

vdahiya12:
@qiluo-msft please cherry-pick into 202012

PR #221, opened 2021-11-03
Skip loading media config from media_settings.json for fast-reboot

tjchadaga:
#### Description

Changes to skip loading media_settings.json in case of fast-reboot

#### Motivation and Context

During fast-reboot, delayed serdes programming done after transceiver detection results in additional interface flap and traffic loss exceeding the required 30s (Issue: https://github.com/Azure/sonic-buildimage/issues/9165). This is addressed by making a backup of media settings from appdb during fast-reboot (https://github.com/Azure/sonic-utilities/pull/1910) and loading this to redis after the switch boots up (https://github.com/Azure/sonic-buildimage/pull/9166). Configuration of serdes triggered by reading media_settings.json file is skipped in this PR

Verified by running advanced_reboot/test_advanced_reboot.py::test_fast_reboot and ensured that the test passes.
Verified TRANSCEIVER_INFO in stateDB, SERDES objects in asicdb after bootup
Also tested coldboot and ensured no side effects.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1464,8 +1464,8 @@ class DaemonXcvrd(daemon_base.DaemonBase):
         # Initialize xcvr table helper
         xcvr_table_helper = XcvrTableHelper()
 
-        output = subprocess.check_output("cat /proc/cmdline", shell=True, universal_newlines=True)
-        if "fast-reboot" in output:
+        output = subprocess.check_output("sonic-db-cli STATE_DB GET 'FAST_REBOOT|system'", shell=True, universal_newlines=True)
+        if "1" in output:
             self.log_info("Skip loading media settings for fast-reboot case")
         else:
             self.load_media_settings()
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -827,6 +827,18 @@ def init_port_sfp_status_tbl(port_mapping, stop_event=threading.Event()):
             else:
                 update_port_transceiver_status_table(logical_port_name, xcvr_table_helper.get_status_tbl(asic_index), sfp_status_helper.SFP_STATUS_INSERTED)
 
+def is_fast_reboot_enabled():
+    fastboot_enabled = False
+    state_db_host =  daemon_base.db_connect("STATE_DB")
+    fastboot_tbl = swsscommon.Table(state_db_host, 'FAST_REBOOT')
+    keys = fastboot_tbl.getKeys()
+
+    if "system" in keys:
+        output = subprocess.check_output('sonic-db-cli STATE_DB get "FAST_REBOOT|system"', shell=True, universal_newlines=True)
+        if "1" in output:
+            fastboot_enabled = True
+
+    return fastboot_enabled
 #
 # Helper classes ===============================================================
 #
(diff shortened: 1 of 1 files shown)


sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1463,7 +1464,12 @@ def init(self):
         # Initialize xcvr table helper
         xcvr_table_helper = XcvrTableHelper()
 
-        self.load_media_settings()
+        output = subprocess.check_output("cat /proc/cmdline", shell=True, universal_newlines=True)
```
yxieca, line 1467: Can you check the db entry instead?

cmdline will never change, so if a device booted up with fast-reboot, cmdline will always have that mark. If we do config reload or load minigraph, we need to take cold reboot procedure.

yxieca (changes requested):
as commented.

sonic-xcvrd/xcvrd/xcvrd.py
yxieca: Please take reference from syncd.sh

tjchadaga: @yxieca - made the requested changes, please review

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1463,7 +1463,14 @@ def init(self):
         # Initialize xcvr table helper
         xcvr_table_helper = XcvrTableHelper()
 
-        self.load_media_settings()
+        fastboot_tbl = swsscommon.Table(xcvr_table_helper.state_db_host, 'FAST_REBOOT')
+        keys = fastboot_tbl.getKeys()
+
+        if "system" in keys:
+            self.log_info("Skip loading media_settings.json in case of fast-reboot")
```
yxieca, line 1470: You should also check if the value is true to be safe. you can use following syntax:

if fastboot.tbl.get('system', False):
    # skip

yxieca: please disregard my example. But please check the value.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1463,7 +1463,14 @@ def init(self):
         # Initialize xcvr table helper
         xcvr_table_helper = XcvrTableHelper()
 
-        self.load_media_settings()
+        fastboot_tbl = swsscommon.Table(xcvr_table_helper.state_db_host, 'FAST_REBOOT')
+        keys = fastboot_tbl.getKeys()
+
+        if "system" in keys:
```
prgeor, line 1469: if "FAST_REBOOT" key has a expiry then we should think about following two scenarios where Xcvrd restart happens after the key has expired:

1. Xcvrd is crashing continiously
2. DRI restarts the PMON/xcvrd

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1463,7 +1463,14 @@ def init(self):
...
+            self.log_info("Skip loading media_settings.json in case of fast-reboot")
+        else:
+            self.load_media_settings()
```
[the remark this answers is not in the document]
tjchadaga: the changes made as part of https://github.com/Azure/sonic-utilities/pull/1910 only makes a backup of media settings  file and not transceiver info.  Skipping notification altogether will require us to backup transceiver info as well. Since transceiver info notification is not impacting fast-reboot or causing a flap, is there a need for this change?

sonic-xcvrd/xcvrd/xcvrd.py
tjchadaga: In error cases like these, wouldn't it be better to not do this optimization and take the cold-reboot route?

prgeor: yes, i agree these fall under cold boot case.

prgeor (approved):

yxieca (approved):

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1463,7 +1464,20 @@ def init(self):
         # Initialize xcvr table helper
         xcvr_table_helper = XcvrTableHelper()
 
-        self.load_media_settings()
+        fastboot_enabled = False
+        fastboot_tbl = swsscommon.Table(xcvr_table_helper.state_db_host, 'FAST_REBOOT')
```
prgeor, line 1468: can we combine changes in line 1468 and 1472 in one function like here: [https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-py-common/sonic_py_common/device_info.py#L575](https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-py-common/sonic_py_common/device_info.py#L575) and use it to determine fast-reboot?

prgeor (approved):

PR #223, opened 2021-11-08
Failed to set Pre-emphasis settings with new xcvr package

aravindmani-1:
#### Description
With new xcvr package, xcvrd fails to set pre-emphasis settings 
#### Motivation and Context
Return types has to be handled properly in xcvrd while handling media settings.

Tested in DellEMC Z9332f platform.
[cable_len_UT.txt](https://github.com/Azure/sonic-platform-daemons/files/7494484/cable_len_UT.txt)

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -668,11 +668,11 @@ def get_media_settings_key(physical_port, transceiver_dict):
         media_key += '-' + media_compliance_code
         if _wrapper_get_sfp_type(physical_port) == 'QSFP_DD':
             if media_compliance_code == "passive_copper_media_interface":
-                if len(media_len) != 0:
-                    media_key += '-' + media_len + 'M'
+                if media_len != 0:
+                    media_key += '-' + str(media_len) + 'M'
         else:
-            if len(media_len) != 0:
-                media_key += '-' + media_len + 'M'
+            if media_len != 0:
+                media_key += '-' + str(media_len) + 'M'
     else:
         media_key += '-' + '*'
 


sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -668,11 +668,11 @@ def get_media_settings_key(physical_port, transceiver_dict):
         media_key += '-' + media_compliance_code
         if _wrapper_get_sfp_type(physical_port) == 'QSFP_DD':
             if media_compliance_code == "passive_copper_media_interface":
-                if len(media_len) != 0:
-                    media_key += '-' + media_len + 'M'
+                if media_len != 0:
+                    media_key += '-' + str(media_len) + 'M'
```
prgeor, line 672: will this change affect platforms that have not migrated to sfp-refactor?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -668,11 +668,11 @@ def get_media_settings_key(physical_port, transceiver_dict):
...
         else:
-            if len(media_len) != 0:
-                media_key += '-' + media_len + 'M'
+            if media_len != 0:
+                media_key += '-' + str(media_len) + 'M'
```
prgeor, line 675: please verify, this change where Dell platform is running with old xcvr package

sonic-xcvrd/xcvrd/xcvrd.py
aravindmani-1: No. Unless cable length is 0 which is very unlikely, this won't affect other platforms.

sonic-xcvrd/xcvrd/xcvrd.py
aravindmani-1: Verified in DellEMC S5232f platform and it is working fine.

prgeor (approved):

PR #226, opened 2021-11-17
Moving pcied service into sonic-host-services, part #1

MaratGubaiev:
fixing https://github.com/Azure/sonic-buildimage/issues/7282

#### Description
pcied removed from sonic-platform-daemons (pmon container)
WARNING it should be merged with coupled changes in sonic-buildimage!


sujinmkang:
@MaratGubaiev please fix the error. I think you need to remove the sonic_pcied from https://github.com/Azure/sonic-platform-daemons/blob/master/azure-pipelines.yml

MaratGubaiev:
> @MaratGubaiev please fix the error. I think you need to remove the sonic_pcied from https://github.com/Azure/sonic-platform-daemons/blob/master/azure-pipelines.yml

@sujinmkang Done. Maybe you know how to fix this problem: https://github.com/Azure/sonic-buildimage/pull/9285#issuecomment-972533456

sujinmkang:
@MaratGubaiev What is the intention to move pcied to host service?

PR #228, opened 2021-11-19
Clear all data from DB table when the daemon stops

sujinmkang:
#### Description
Clear all data from DB table when the daemon stops

#### Motivation and Context
Clean up when the daemon stops

Check Thermal and physical info data if they are cleared after stopping the thermalctld.

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -214,6 +214,20 @@ class FanUpdater(logger.Logger):
         for name in self.fan_status_dict.keys():
             self.table._del(name)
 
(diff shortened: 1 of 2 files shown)


sonic-thermalctld/scripts/thermalctld
```diff
@@ -214,6 +214,20 @@ class FanUpdater(logger.Logger):
         for name in self.fan_status_dict.keys():
```
Junchao-Mellanox, line 214: should we remove this line?

Junchao-Mellanox (approved):

prgeor (approved):

PR #230, opened 2021-11-25
[xcvrd] Implement to read sfp info from eeprom several times while de…

dennis0113:
#### Description
It is known that some sfp modules will need more time to power up its
data path and sometimes it caused that the application takes the
incorrect or uninitialized data to display.

The original way is to read eeprom as soon as the sfp is inserted. And
the latest modification in PR(https://github.com/Azure/sonic-platform-daemons/pull/201)
is to add a 2 seconds delay after sfp is detected.

Our modification is to keep the original way which will read eeprom as
soon as the sfp is detected. It is expected that the eeprom data in
some sfp modules which have shorter warm up time can be obtained
immediately. For the other modules which has longer init time, xcvrd
will have a second try to read eeprom when sfp is inserted for 2 seconds
and the third try will be applied when sfp is inserted for 5 seconds.
Record to syslog when the checksum of sfp info read from eeprom is
different from the last try.

1.Build sonic_xcvrd python .whl
2.Install the python wheel on switch and insert sfp transceiver

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -9,6 +9,7 @@ try:
     import ast
     import copy
     import functools
+    import hashlib
     import json
     import multiprocessing
     import os
@@ -17,7 +18,8 @@ try:
     import threading
     import time
     import subprocess
- 
+
+    from enum import Enum
     from sonic_py_common import daemon_base, device_info, logger
     from sonic_py_common import multi_asic
     from swsscommon import swsscommon
@@ -46,6 +48,16 @@ MGMT_INIT_TIME_DELAY_SECS = 2
 
 # SFP insert event poll duration
 SFP_INSERT_EVENT_POLL_PERIOD_MSECS = 1000
+
+# state of read sfp eeprom
+READ_EEPROM_STATE_INSERTED = 0
+READ_EEPROM_STATE_SECOND_TRY = 1
+READ_EEPROM_STATE_LAST_TRY = 2
+
+class insert_event_field(Enum):
+    STATE = 0
+    TIMESTAMP = 1
+    CHECKSUM = 2
 
 DOM_INFO_UPDATE_PERIOD_SECS = 60
 STATE_MACHINE_UPDATE_PERIOD_MSECS = 60000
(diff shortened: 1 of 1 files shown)


prgeor:
Please note, datapath init is different from management init. Management init can take at max 2 seconds for any QSFP28,QSFP+,QSFP-DD. 

1. What is the motivation of this PR?
2. What issue are you facing? Which spec does your transceiver follow ?

dennis0113:
1.What is the motivation of this PR?
   > For some kinds of QSFP which has faster warm up time, we can read its eeprom data
      as soon as it was detected as presence by xcvrd.
      For some kinds of QSFP which has slower warm up time, 
      we can have more try to read its eeprom data.
      
2.What issue are you facing? Which spec does your transceiver follow ?
   > We found that some QSFP's eeprom data could not be updated to redis-db correctly by xcvrd.
       When this situation occurred, we compared the hexdump of sys node where xcvrd extracted eeprom data.
       and find that the data in sys node is correct, but xcvrd just could not update it to db.
       We considered that this kinds of QSFP may have long warm up time to prepare the data in eeprom.
       Therefore, have this implement to let xcvrd can read eeprom data more times once the transceiver is detected.

prgeor:
> 1.What is the motivation of this PR?
> 
> > For some kinds of QSFP which has faster warm up time, we can read its eeprom data
> > directly as soon as it was detected as presence by xcvrd.

What spec does these kind of QSFP follow? Who is the vendor of such QSFP?

> > For some kinds of QSFP which has slower warm up time,
> > we can have more tries to read its eeprom data.
> 

What spec does these kind of QSFP follow? Who is the vendor of such QSFP?


> 
> > We found that some QSFP's eeprom data could not be updated to redis-db correctly by xcvrd.
> > When this situation occurred, we compared the hexdump of sys node where xcvrd extracted eeprom data.
> > and find that the data in sys node is correct, but xcvrd just could not update it to db.
> > We considered that this kinds of QSFP may have long warm up time to prepare the data in eeprom.
> > Therefore, have this implement to let xcvrd can read eeprom data more times once the transceiver is detected.

From SFF-8679 QSFP28-hw spec the initialization time for i2c to be ready is 2000 msec MAX. Looks like the QSFP module in your case doesn't seem to be compliant with the spec, if so please check with the QSFP vendor. If the module i2c is ready before 2000 msec timeout, then Xcvrd will wait at max 2000 msec before it does the first i2c transaction. Why do you think 2 seconds wait time before reading and updating the DB is a concern? Which platform are you testing?

image

prgeor:
@dennis0113 are you still looking for this PR?

dennis0113:
Sorry... I will keep working on this PR and update the reply for the question you had mentioned.

PR #233, opened 2021-12-11
[voq][pmon]xcvrd process show backtrace in the syslog on  VOQ linecard.

mlok-nokia:
#### Description
xcvrd_utilities/port_mapping class returns the port_mapping_data info which includes internal ports. Internal port doesn't have transceiver. It should not be in the list for the xcvrd.
Issue -- https://github.com/Azure/sonic-buildimage/issues/9505

This commit adds code to check and skip internal port if a port is either Backplane or Inband port.  

#### Motivation and Context
This change will skip the internal port so that will trigger the exception in the XCVRD.
 https://github.com/Azure/sonic-buildimage/issues/9505

Boot up the VOQ linecard.  Verify there is no such exception in syslog

### DIFF
--- a/sonic-xcvrd/tests/test_xcvrd.py
+++ b/sonic-xcvrd/tests/test_xcvrd.py
@@ -341,8 +341,8 @@ class TestXcvrdScript(object):
     @patch('swsscommon.swsscommon.Table')
     def test_get_port_mapping(self, mock_swsscommon_table):
         mock_table = MagicMock()
-        mock_table.getKeys = MagicMock(return_value=['Ethernet0', 'Ethernet4'])
-        mock_table.get = MagicMock(side_effect=[(True, (('index', 1), )), (True, (('index', 2), ))])
+        mock_table.getKeys = MagicMock(return_value=['Ethernet0', 'Ethernet4', 'Ethernet-IB0'])
+        mock_table.get = MagicMock(side_effect=[(True, (('index', 1), )), (True, (('index', 2), )), (True, (('index', 3), ))])
         mock_swsscommon_table.return_value = mock_table
         port_mapping = get_port_mapping()
         assert port_mapping.logical_port_list.count('Ethernet0')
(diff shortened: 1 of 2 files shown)


mlok-nokia:
@abdosi @judyjoseph @minionatwork Please review this PR which address the backtrace on the Xcvrd process in PMON in the latest code.

sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
```diff
@@ -138,6 +139,8 @@ def handle_port_update_event(sel, asic_context, stop_event, logger, port_change_
                 (key, op, fvp) = port_tbl.pop()
                 if not key:
                     break
+                if key.startswith((backplane_prefix(), inband_prefix(), recirc_prefix())):
```
judyjoseph, line 142: Looks good, suggest to have a function defined eg: validate_port which will have this logic.

mlok-nokia: Ok.  I have created validate_port function to handle it as suggest. Also add the modify the test case to test the change

sonic-xcvrd/tests/test_xcvrd.py
```diff
@@ -341,8 +341,8 @@ def test_handle_port_config_change(self, mock_select, mock_sub_table):
     @patch('swsscommon.swsscommon.Table')
     def test_get_port_mapping(self, mock_swsscommon_table):
         mock_table = MagicMock()
-        mock_table.getKeys = MagicMock(return_value=['Ethernet0', 'Ethernet4'])
-        mock_table.get = MagicMock(side_effect=[(True, (('index', 1), )), (True, (('index', 2), ))])
+        mock_table.getKeys = MagicMock(return_value=['Ethernet0', 'Ethernet4', 'Ethernet-IB0'])
+        mock_table.get = MagicMock(side_effect=[(True, (('index', 1), )), (True, (('index', 2), )), (True, (('index', 3), ))])
```
judyjoseph, line 345: Thanks for adding this check, can you add the assert fail check too for this.

mlok-nokia: Thanks Judy,  I have added assert check as your suggestion.

judyjoseph (approved):

abdosi (approved):

judyjoseph:
@mlok-nokia  Please raise a separate PR for 202111 branch as there are conflicts.

PR #236, opened 2022-01-25
[pmon][xcvrd] Porting PR233 -- xcvrd process show backtrace on the internal port

mlok-nokia:
#### Description
(Porting PR233 to branch 202111)
xcvrd_utilities/port_mapping class returns the port_mapping_data info which includes internal ports. Internal port doesn't have transceiver. It should not be in the list for the xcvrd.

#### Motivation and Context
(Due to the code base is different, cherry-pick cannot be done) 
This change will skip the internal port so that will trigger the exception in the XCVRD.
Boot up the VOQ linecard. Verify there is no such exception in syslog


mlok-nokia:
@judyjoseph Per your request, I have created this PR for porting the PR 233. Please review it. Thanks

prgeor (approved):

judyjoseph (approved):

PR #246, opened 2022-02-28
Fix UT failed cause by change pycommon to use swsscommon

liuh-80:
Fix UT failed cause by change sonic_py_common to use swss_common

#### Description
    Fix UT failed cause by change sonic_py_common to use swss_common:

        In current UT, we implement swsscommon mock class under sonic-pcied/tests/mocked_libs/swsscommon and load these mock libs by add the mock path in front of real swsscommon lib path with following code:

              # Add mocked_libs path so that the file under test can load mocked modules from there
              mocked_libs_path = os.path.join(tests_path, "mocked_libs")
              sys.path.insert(0, mocked_libs_path)

        However, because we change sonic_py_common to use swsscommon, so real version of swsscommon been load before we load mock lib. then the UT break with following error message:
              
              ______________ TestDaemonPcied.test_update_pcie_devices_status_db ______________
              self = <tests.test_DaemonPcied.TestDaemonPcied object at 0x7f29542655d0>
                  @mock.patch('pcied.load_platform_pcieutil', mock.MagicMock())
                  def test_update_pcie_devices_status_db(self):
               >       daemon_pcied = pcied.DaemonPcied(SYSLOG_IDENTIFIER)
              tests/test_DaemonPcied.py:160: 
              _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
              scripts/pcied:93: in __init__
                  self.device_table = swsscommon.Table(self.state_db, PCIE_DEVICE_TABLE_NAME)
              _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
              self = <swsscommon.swsscommon.Table;  >
              args = (<MagicMock name='mock()' id='139815470171344'>, 'PCIE_DEVICE')
                  def __init__(self, *args):
               >       this = _swsscommon.new_Table(*args)
               E       NotImplementedError: Wrong number or type of arguments for overloaded function 'new_Table'.
               E         Possible C/C++ prototypes are:
               E           swss::Table::Table(swss::DBConnector const *,std::string const &)
               E           swss::Table::Table(swss::RedisPipeline *,std::string const &,bool)

#### Motivation and Context
    pyswss will be deprecate, so sonic_py_common will changed to use swss_common.
    Some UT in this project failed because this change.

    Pass all UT and sonic-buildimage E2E test.


liuh-80:
Code in this PR will be test in another PR:
    https://github.com/Azure/sonic-buildimage/pull/10099

qiluo-msft:
Could you add "UT failed" message into PR description?

liuh-80:
> Could you add "UT failed" message into PR description?

Fixed, Add UT error message to PR description.

liuh-80:
Will push a update later, the UT issue is:
In current UT, we implement swsscommon mock class under sonic-pcied/tests/mocked_libs/swsscommon and load these mock libs by add the mock path in front of real swsscommon lib path with following code: 

    # Add mocked_libs path so that the file under test can load mocked modules from there
    mocked_libs_path = os.path.join(tests_path, "mocked_libs")
    sys.path.insert(0, mocked_libs_path)

However, because we change sonic_py_common to use swsscommon, so real version of swsscommon been load before we load mock lib. then the UT break.

So will check if can load mock lib first by change the code orider.

sonic-pcied/tests/test_pcied.py
```diff
@@ -18,6 +18,7 @@
 # Add mocked_libs path so that the file under test can load mocked modules from there
 mocked_libs_path = os.path.join(tests_path, "mocked_libs")
 sys.path.insert(0, mocked_libs_path)
+from sonic_py_common import daemon_base, device_info
```
qiluo-msft, line 21: This line is same as L13. I think it will be ignored.
Do you want to remove L13?

liuh-80: Fixed, this is a mistake.

qiluo-msft (approved):

PR #249, opened 2022-03-24
[ycable] fix the logic to update cable_info values when ycable is not present; fix read side logic for ycable

vdahiya12:

#### Description
This PR fixes the MUX_CABLE_INFO table insertion logic which was not done before this PR. If the muxcable is plugged in, with this change if the config DB states that it is a muxcable then the logic populates the MUX_CABLE_INFO with 'N/A' or 
 'unknown' entries. This was not done before
This PR also fixes the read_side logic, which if initialized with a -1 value would never be updated to the correct value, this PR makes sure that if the read_side is read correctly if cable not powered on/disconnected from server side, and the read_side is populated correctly in state DB. Before this change the read_side was not populated correctly if the cable was not powered from server side/disconnected. 

#### Motivation and Context
Required for the logic of services looking for MUX_CABLE_INFO values which require this be populated without cable presence.
read_side fix logic makes sure that if the cable is powered off/disconnected from server side then the cable driver works well in those cases

Unit-Tests and tested in the lab

### DIFF
--- a/sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
+++ b/sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
@@ -2529,7 +2529,7 @@ class YCableTableUpdateTask(object):
                             new_status = 'unknown'
 
                         helper_logger.log_debug("Y_CABLE_DEBUG: xcvrd successful to transition port {} from {} to {} and write back to the DB {}".format(port, old_status, new_status, threading.currentThread().getName()))
-                        helper_logger.log_notice("Got a change event for toggle the mux-direction active side for port {} state requested {} from old state {} to new state {} thread {}".format(port, requested_status, old_status, new_status, threading.currentThread().getName()))
+                        helper_logger.log_notice("Got a change event for toggle the mux-direction active side for port {} state requested {} from old state {} to new state {} read_side  {} thread id {}".format(port, requested_status, old_status, new_status, read_side, threading.currentThread().getName()))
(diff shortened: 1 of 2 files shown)


sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
```diff
@@ -460,11 +488,11 @@ def read_y_cable_and_update_statedb_port_tbl(logical_port_name, mux_config_tbl):
                     helper_logger.log_warning("Failed to execute the get_mux_direction for port {} due to {}".format(physical_port,repr(e)))
 
             if active_side is None or active_side not in y_cable_switch_state_values:
-                read_side = active_side = -1
+                active_side = -1
                 update_table_mux_status_for_statedb_port_tbl(
                     mux_config_tbl, "unknown", read_side, active_side, logical_port_name)
                 helper_logger.log_error(
-                    "Error: Could not establish the active side for  Y cable port {} to perform read_y_cable update state db".format(logical_port_name))
+                    "Line 495 Error: Could not establish the active side for Y cable port {} to perform read_y_cable update state db".format(logical_port_name))
```
zjswhhh, line 495: Is `Link 495` intended to be here?

vdahiya12: fixed

zjswhhh (approved):
LGTM.

zjswhhh (approved):

PR #250, opened 2022-03-30
Add serial_num field for chassis info in STATE_DB

pettershao-ragilenetworks:
upload serial number to the STATE_DB under the CHASSIS_INFO table as a new chassis hardware information for users

#### Description
Expose serial number to user space so that CLI utilities have access to it.

#### Motivation and Context
The "show version" command will get "serial" field value in STATE_DB under the CHASSIS_INFO table as the "Serial Number" display. However, the platform API "get_serial" is suppose to be used to get the service tag of the chassis instead of the actual serial number. I made this change in order that the actual serial number can be exposed to SONiC user and the CLI utilities can have access to it.

After the merge of the related CLI utilities changes, use "show version" to check the "Serial Number" field

"show version" command result before this change:
image
"show version" command result after this change

### DIFF
--- a/sonic-chassisd/scripts/chassis_db_init
+++ b/sonic-chassisd/scripts/chassis_db_init
@@ -30,6 +30,7 @@ CHASSIS_INFO_TABLE = 'CHASSIS_INFO'
 CHASSIS_INFO_KEY_TEMPLATE = 'chassis {}'
 CHASSIS_INFO_CARD_NUM_FIELD = 'module_num'
 CHASSIS_INFO_SERIAL_FIELD = 'serial'
+CHASSIS_INFO_SERIAL_NUM_FIELD = 'serial_num'
 CHASSIS_INFO_MODEL_FIELD = 'model'
 CHASSIS_INFO_REV_FIELD = 'revision'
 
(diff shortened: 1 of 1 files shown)


pettershao-ragilenetworks:
@jleveque @alexrallen please help review this pr.

alexrallen (changes requested):
Hi there,

FYI Joe and I no longer maintain SONiC just to set some expectations about how much time we can commit to helping you with this. 

Couple questions....

1. What specific hardware are you running (manufacturer / pn) you can get this via `show platform summary` or dumping the syseeprom for me
2. I'm not familiar with the difference between the "service tag" and the "serial number" but I am fairly certain that the intention of the `get_serial()` endpoint is to get the serial number (it is at least on NVIDIA platforms). Where did you get the impression that the intention of that endpoint is to get the service tag?

There are a few issues with this PR....

1. It seems like your intention is to add a new field to STATE_DB called `serial_num` in addition to adding a new method to the Platform API of the same name. The similarity of these names will result in confusion as they fail to differentiate themselves.
2. It seems you will also be changing sonic-utilities to utilize this new `serial_num` field instead of `serial` this will break the backwards compatibility across *all* platforms in SONiC as everyone will need to populate `serial_num` in order for their serial number to present to the user correctly. This would not be acceptable. (Do you have a sonic-utilities PR? If so please link it)
3. The root of this issue seems to be that the *Platform API* is not returning what you expect it to. Rather than making global changes in the infrastructure of SONiC it appears to me that the specific platform API implementation of the switch platform you are using should be revised if necessary to return the expected value. Right now this change will impact many platforms at the same time which I don't believe are currently experiencing the issue you describe. 

Please let me know if there are any further questions.

dgsudharsan (changes requested):
@pettershao-ragilenetworks As alex pointed out its unclear what's the motivation of this PR. We already have a serial number exposed through platform API. If you have to replace serial number with service tag, please make the change in your corresponding platform API. Do not generic code

PR #252, opened 2022-04-05
Added support for port prefix regex

itamar-talmon:
**- What I did**
    Removed the dependency on the "Ethernet" string in the SONiC code base and added support
	for extending the front panel port name pattern.

**- How I did it**
    1. Introduced FRONT_PANEL_PORT_PREFIX_REGEX that extends the old FRONT_PANEL_PORT_PREFIX ("Ethernet")
	2. Updated all the relevant usage of the "Ethernet" throughout the code base to use the new regex pattern

**- How to verify it**
    Pass all UT and CI testing.


itamar-talmon:
This PR depends on https://github.com/Azure/sonic-swss-common/pull/598 so it will not pass until it merged

prgeor:
@itamar-talmon unit tests are failing

sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
```diff
@@ -102,7 +102,7 @@ def format_mapping_identifier(string):
 
 
 def logical_port_name_to_physical_port_list(port_name):
-    if port_name.startswith("Ethernet"):
+    if re.search(swsscommon.FRONT_PANEL_PORT_PREFIX_REGEX, port_name):
```
qiluo-msft, line 105: Use `re.match` to match the whole `port_name`.

itamar-talmon: done

itamar-talmon:
@kcudnik @qiluo-msft - could you please check it out after my update?
Checks will fail until related commits will be merged (in the description).
Thank you!

kcudnik (approved):

qiluo-msft (dismissed):

prgeor:
Quoting @lguohan here:-

the overall motivation is not quite clear. can you describe what are you trying to achieve here? here is the sonic port naming convention. for new port names convention to be introduced, we need to make a proposal first here. https://github.com/sonic-net/SONiC/blob/master/doc/sonic-port-name.md

PR #254, opened 2022-04-11
[CMIS]Improved 400G link bring up sequence

prgeor:
#### Description
Improved 400G link bring up sequence [HLD](https://github.com/sonic-net/SONiC/blob/master/doc/sfp-cmis/Interface-Link-bring-up-sequence.md)

1. Ensure CMIS compliant transceiver Datapath are initialized and activated IFF the host Tx signal is good and admin_status is UP
2. get_all_namespace() API is not reentrant resulting in crash, so we cache the namespace at the xcvrd start
3. Datapath is deactivated if either 'admin_status' becomes down or 'host_tx_ready' signal becomes OFF.
4. An option "--skip_xcvrd_cmis_mgr" is added to skip launch of CMIS task manager for platforms that are in transition to using SFP-refactor code 
5. Datapath is NOT impacted if Xcvrd crash/restarts

#### Motivation and Context
Deterministically improve the link bring up of 400G link that are using CMIS compliant transceivers.

Tested on 
1. Cisco 8800 chassis: Being validated by Cisco
2. Dell Z9332F

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -912,29 +912,6 @@ def is_fast_reboot_enabled():
 
     return fastboot_enabled
 
-# Returns 'True' if the Host Tx signal(from the host towards the QSFP-DD)
-# is indicated to be good by Orchagent, 'False' otherwise
-def is_host_tx_ready(lport):
-    host_tx_ready = 'False'
-    state_db_host =  daemon_base.db_connect("STATE_DB")
-    port_tbl = swsscommon.Table(state_db_host, 'PORT_TABLE')
-    keys = port_tbl.getKeys()
-
-    if lport in keys:
-       cmd = 'sonic-db-cli STATE_DB hget "PORT_TABLE|{}" host_tx_ready'.format(lport)
-       proc = subprocess.Popen(cmd, shell=True, universal_newlines=True, stdout=subprocess.PIPE)
-       (stdout, stderr) = proc.communicate()
-       host_tx_ready = stdout.rstrip('\n')
-       #helper_logger.log_warning("$$${} host_tx_ready = {}".format(lport, host_tx_ready))
-       # Check admin_status too ...just in case
-       cmd = 'sonic-db-cli APPL_DB hget "PORT_TABLE:{}" admin_status'.format(lport)
-       proc = subprocess.Popen(cmd, shell=True, universal_newlines=True, stdout=subprocess.PIPE)
-       (stdout, stderr) = proc.communicate()
-       admin_status = stdout.rstrip('\n')
-       #helper_logger.log_warning("$$${} admin_status = {}".format(lport, admin_status))
-
-    return admin_status == 'up' and host_tx_ready == 'True'
-
 #
 # Helper classes ===============================================================
 #
@@ -1220,29 +1197,6 @@ class CmisManagerTask:
                 break
 
         return done
-
-    # Returns 'True' if the Host Tx signal(from the host towards the QSFP-DD)
-    # is indicated to be good by Orchagent, 'False' otherwise
-    def is_host_tx_ready(self, lport):
-        host_tx_ready = 'False'
-        admin_status = 'down'
-
-        asic_index = self.port_mapping.get_asic_id_for_logical_port(lport)
-        state_port_tbl = self.xcvr_table_helper.get_state_port_tbl(asic_index)
-        cfg_port_tbl = self.xcvr_table_helper.get_cfg_port_tbl(asic_index)
-
-        found, port_info = cfg_port_tbl.get(lport)
-        if found:
-            # Check admin_status too ...just in case
-            admin_status = dict(port_info)['admin_status']
-            helper_logger.log_warning("$$${} admin_status = {}".format(lport, admin_status))
-
-        helper_logger.log_warning("$$${} host_tx_ready = {}".format(lport, host_tx_ready))
-        found, port_info = state_port_tbl.get(lport)
-        if found and 'host_tx_ready' in dict(port_info):
-            host_tx_ready = dict(port_info)['host_tx_ready']
-
-        return admin_status == 'up' and host_tx_ready == 'True'
 
     def get_host_tx_status(self, lport):
         host_tx_ready = 'False'
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1230,7 +1230,7 @@ class CmisManagerTask:
         print("Starting")
 
         # APPL_DB for CONFIG updates, and STATE_DB for insertion/removal
-        sel, asic_context = port_mapping.subscribe_port_update_event()
+        sel, asic_context = port_mapping.subscribe_port_update_event(self.namespaces)
         while not self.task_stopping_event.is_set():
             # Handle port change event from main thread
             port_mapping.handle_port_update_event(sel,
@@ -1476,7 +1476,7 @@ class DomInfoUpdateTask(object):
         helper_logger.log_info("Start DOM monitoring loop")
         dom_info_cache = {}
         dom_th_info_cache = {}
-        sel, asic_context = port_mapping.subscribe_port_config_change()
+        sel, asic_context = port_mapping.subscribe_port_config_change(self.namespaces)
 
         # Start loop to update dom info in DB periodically
         while not self.task_stopping_event.wait(DOM_INFO_UPDATE_PERIOD_SECS):
@@ -1649,7 +1649,7 @@ class SfpStateUpdateTask(object):
         retry = 0
         timeout = RETRY_PERIOD_FOR_SYSTEM_READY_MSECS
         state = STATE_INIT
-        sel, asic_context = port_mapping.subscribe_port_config_change()
+        sel, asic_context = port_mapping.subscribe_port_config_change(self.namespaces)
         port_change_event_handler = functools.partial(self.on_port_config_change, stopping_event)
         while not stopping_event.is_set():
             port_mapping.handle_port_config_change(sel, asic_context, stopping_event, self.port_mapping, helper_logger, port_change_event_handler)
@@ -2099,7 +2099,7 @@ class DaemonXcvrd(daemon_base.DaemonBase):
         for namespace in self.namespaces:
             self.wait_for_port_config_done(namespace)
 
-        port_mapping_data = port_mapping.get_port_mapping()
+        port_mapping_data = port_mapping.get_port_mapping(self.namespaces)
 
         # Post all the current interface dom/sfp info to STATE_DB
         self.log_info("Post all port DOM/SFP info to DB")
@@ -2116,7 +2116,7 @@ class DaemonXcvrd(daemon_base.DaemonBase):
         self.log_info("Start daemon deinit...")
 
         # Delete all the information from DB and then exit
-        port_mapping_data = port_mapping.get_port_mapping()
+        port_mapping_data = port_mapping.get_port_mapping(self.namespaces)
         logical_port_list = port_mapping_data.logical_port_list
         for logical_port_name in logical_port_list:
             # Get the asic to which this port belongs
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -994,10 +994,10 @@ class CmisManagerTask:
                 self.port_dict[lport]['lanes'] = port_change_event.port_dict['lanes']
             if 'host_tx_ready' in port_change_event.port_dict:
                 self.port_dict[lport]['host_tx_ready'] = port_change_event.port_dict['host_tx_ready']
-            if 'admin_status' in port_change_event.port_dict and 'oper_status' in port_change_event.dict:
+            if 'admin_status' in port_change_event.port_dict and 'oper_status' in port_change_event.port_dict:
                 # At times 'admin_status' is NOT the same in the PORT_TABLE of APPL_DB and STATE_DB
                 # We dont have better way to check if 'admin_status' is from APPL_DB or STATE_DB so this
-                # check is put temporarily to list only to APPL_DB's admin_status and ignore that of STATE_DB
+                # check is put temporarily to listen only to APPL_DB's admin_status and ignore that of STATE_DB
                 self.port_dict[lport]['admin_status'] = port_change_event.port_dict['admin_status']
             self.force_cmis_reinit(lport, 0)
         else:
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1331,6 +1331,8 @@ class CmisManagerTask:
                            self.log_notice("{} Forcing Tx laser OFF".format(lport))
                            # Force DataPath re-init
                            api.tx_disable_channel(host_lanes, True)
+                           self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_READY
+                           continue
 
                         appl = self.get_cmis_application_desired(api, host_lanes, host_speed)
                         if appl < 1:
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -937,15 +937,16 @@ class CmisManagerTask:
     CMIS_STATE_REMOVED   = 'REMOVED'
     CMIS_STATE_FAILED    = 'FAILED'
 
-    def __init__(self, port_mapping, skip_cmis_mgr=False):
+    def __init__(self, namespaces, port_mapping, skip_cmis_mgr=False):
         self.task_stopping_event = multiprocessing.Event()
         self.task_process = None
         self.port_dict = {}
         self.port_mapping = copy.deepcopy(port_mapping)
-        self.xcvr_table_helper = XcvrTableHelper()
+        self.xcvr_table_helper = XcvrTableHelper(namespaces)
         self.isPortInitDone = False
         self.isPortConfigDone = False
         self.skip_cmis_mgr = skip_cmis_mgr
+        self.namespaces = namespaces
 
     def log_notice(self, message):
         helper_logger.log_notice("CMIS: {}".format(message))
@@ -978,27 +979,24 @@ class CmisManagerTask:
 
         # Skip if the port/cage type is not a CMIS
         # 'index' can be -1 if STATE_DB|PORT_TABLE
-        # TODO:Make sure doesnot breakt non-CMIS
-        #ptype = _wrapper_get_sfp_type(pport)
-        #if ptype not in self.CMIS_MODULE_TYPES:
-        #    return
-
         if lport not in self.port_dict:
             self.port_dict[lport] = {}
+
+        if port_change_event.port_dict is None:
+            return
 
         if port_change_event.event_type == port_change_event.PORT_SET:
             if pport >= 0:
                 self.port_dict[lport]['index'] = pport
-            if port_change_event.port_dict is not None and 'speed' in port_change_event.port_dict:
+            if 'speed' in port_change_event.port_dict and port_change_event.port_dict['speed'] != 'N/A':
                 self.port_dict[lport]['speed'] = port_change_event.port_dict['speed']
-            if port_change_event.port_dict is not None and 'lanes' in port_change_event.port_dict:
+            if 'lanes' in port_change_event.port_dict:
                 self.port_dict[lport]['lanes'] = port_change_event.port_dict['lanes']
-            if port_change_event.port_dict is not None and 'host_tx_ready' in port_change_event.port_dict:
+            if 'host_tx_ready' in port_change_event.port_dict:
                 self.port_dict[lport]['host_tx_ready'] = port_change_event.port_dict['host_tx_ready']
-            if port_change_event.port_dict is not None and 'admin_status' in port_change_event.port_dict:
+            if 'admin_status' in port_change_event.port_dict:
                 self.port_dict[lport]['admin_status'] = port_change_event.port_dict['admin_status']
-            self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_INSERTED
-            self.reset_cmis_init(lport, 0)
+            self.force_cmis_reinit(lport, 0)
         else:
             self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_REMOVED
 
@@ -1123,12 +1121,15 @@ class CmisManagerTask:
 
         return True
 
-    def reset_cmis_init(self, lport, retries=0):
+    def force_cmis_reinit(self, lport, retries=0):
+        """
+        Try to force the restart of CMIS state machine
+        """
         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_INSERTED
         self.port_dict[lport]['cmis_retries'] = retries
         self.port_dict[lport]['cmis_expired'] = None # No expiration
 
-    def test_module_state(self, api, states):
+    def check_module_state(self, api, states):
         """
         Check if the CMIS module is in the specified state
 
@@ -1143,7 +1144,7 @@ class CmisManagerTask:
         """
         return api.get_module_state() in states
 
-    def test_config_error(self, api, channel, states):
+    def check_config_error(self, api, channel, states):
         """
         Check if the CMIS configuration states are in the specified state
 
@@ -1171,7 +1172,7 @@ class CmisManagerTask:
 
         return done
 
-    def test_datapath_state(self, api, channel, states):
+    def check_datapath_state(self, api, channel, states):
         """
         Check if the CMIS datapath states are in the specified state
 
@@ -1200,7 +1201,7 @@ class CmisManagerTask:
         return done
 
     def get_host_tx_status(self, lport):
-        host_tx_ready = 'False'
+        host_tx_ready = 'false'
 
         asic_index = self.port_mapping.get_asic_id_for_logical_port(lport)
         state_port_tbl = self.xcvr_table_helper.get_state_port_tbl(asic_index)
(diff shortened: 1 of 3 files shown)


sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -842,6 +842,30 @@ def is_fast_reboot_enabled():
...
+# Returns 'True' if the Host Tx signal(from the host towards the QSFP-DD)
+# is indicated to be good by Orchagent, 'False' otherwise
+def is_host_tx_ready(lport):
+    host_tx_ready = 'False'
+    state_db_host =  daemon_base.db_connect("STATE_DB")
+    port_tbl = swsscommon.Table(state_db_host, 'PORT_TABLE')
+    keys = port_tbl.getKeys()
+
+    if lport in keys:
+       cmd = 'sonic-db-cli STATE_DB hget "PORT_TABLE|{}" host_tx_ready'.format(lport)
```
sacnaik, line 855: Pmon is global container, how this wok under multi NPU? I mean sonic-db-cli does not require to look namesapce db ?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1223,37 +1248,36 @@ def task_worker(self):
...
-                            self.log_notice("{}: READY".format(lport))
+                            self.log_notice("{}: no CMIS application update required...READY".format(lport))
                             self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_READY
                             continue
 
+                        self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_DP_DEINIT
+                    elif state == self.CMIS_STATE_DP_DEINIT:
                         # D.2.2 Software Deinitialization
                         api.set_datapath_deinit(host_lanes)
-                        api.set_lpmode(True)
```
sacnaik, line 1232: You are changing generic code. Isn't this set_lpmode is required function?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1177,11 +1201,13 @@ def task_worker(self):
                     # Skip if XcvrApi is not supported
                     api = sfp.get_xcvr_api()
                     if api is None:
+                        self.log_error("{}: skipping CMIS state machine since no xcvr api!!!".format(lport))
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_READY
                         continue
 
                     # Skip if it's not a paged memory device
                     if api.is_flat_memory():
+                        self.log_notice("{}: skipping CMIS state machine for flat memory xcvr".format(lport))
```
sacnaik, line 1210: You may be periodically logging this, then it is not right.

sonic-xcvrd/xcvrd/xcvrd.py
shyam77git: > Pmon is global container, how this wok under multi NPU? I mean sonic-db-cli does not require to look namesapce db ?

@prgeor updated that he would make this thing namespace aware i.e. xcvrd would subscribe/listen to all namespace's stateDB and then act accordingly. This piece yet to be added

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1271,42 +1295,45 @@ def task_worker(self):
...
                             continue
 
+                        # Ensure the Datapath is NOT Activated unless the host Tx siganl is good.
+                        # NOTE: Some CMIS compliant modules may have 'auto-squelch' feature where
+                        # the module won't take datapaths to Activated state if host tries to enable
+                        # the datapaths while there is no good Tx signal from the host-side.
+                        if not is_host_tx_ready(lport):
+                            self.log_notice("{} waiting for host tx ready...".format(lport))
+                            continue
+
```
shyam77git, line 1313: is_host_tx_ready flag/signal notification to be checked as well when interface/port is admin shutdown.
Tx laser to be turned off. 
Listen to this notification and perform DP_DEACTIVATE followed by DP_DEINIT on the optical module. 

Please add this use-case handling.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1271,42 +1295,45 @@ def task_worker(self):
...
                             if (expired is not None) and (expired <= now):
                                 self.log_notice("{}: timeout for 'ConfigSuccess'".format(lport))
                                 self.reset_cmis_init(lport, retries + 1)
                             continue
 
+                        # Ensure the Datapath is NOT Activated unless the host Tx siganl is good.
+                        # NOTE: Some CMIS compliant modules may have 'auto-squelch' feature where
+                        # the module won't take datapaths to Activated state if host tries to enable
+                        # the datapaths while there is no good Tx signal from the host-side.
+                        if not is_host_tx_ready(lport):
```
jaganbal-a, line 1310: Can this be implemented as a event handler? 
1. Subscribe to PORT_TABLE in state-db, then filter to host_tx_ready field. 
2. Based on the host_tx_ready status, set the appropriate CMIS module state will be set here in the event handler.
3. Since the main CMIS task worker keep checking for CMIS state change for all ports, once the point (2) is set, the CMIS task worker will work on the appropriate state.
4. we can introduce a new state between CMIS_STATE_AP_CONF and CMIS_STATE_DP_INIT,  say 'CMIS_STATE_WAIT_FOR_HOST' . so during module insertion the CMIS task worker will set to this state once it completes CMIS_STATE_AP_CONF and this state will be changed to CMIS_STATE_DP_INIT only in the host_tx_ready handler. this way it will be synchornized.
  
Keeping the host_tx_ready processing as event handler will have following benefit.
1. Keeping the host_tx_ready out of CMIS task worker to event handler will benefit other speed to follow the deterministic link bringup approach.
2. It will avoid continuous polling of host-tx-ready using Redis-DB query, which will save time when scaled to fully populated 48/72 ports.

prgeor: @jaganbal-a I think event handling may NOT work when Xcvrd restarts. Consider this:-
1. 400G Port link is up.
2. Xcvrd crashes and in the meantime someone plugs in another 400G module
3. Xcvrd won't get another 'host_tx_ready' event because 'host_tx_ready' doesn't depend upon whether module is present or not. So, Xcvrd will keep waiting for 'host_tx_ready' or unless user does admin shut followed by admin up.

Another similar case as above
1. Xcvrd is running and got 'host_tx_ready'  but module is absent 
2. Xcvrd crashes and comes up
3. Module is plugged in but Xcvrd is not going to receive 'host_tx_ready' and link won't come up

jaganbal-a: @prince 
It is captured as part of design that the host_tx_ready will be updated regardless of transceiver presence, and it is xcvrd job to check the host_tx_ready during module Init during insertion <a local host_tx_ready dictionary will help here>, so the DP state will be moved to DP_INIT/DP_ACTIVATE state based on host_tx_ready.

prgeor: @jaganbal-a Xcvrd process don't have a persistent storage to capture the 'host_tx_ready' state, so it depends upon the Redis-DB i.e it needs to read the 'host_tx_ready' state from the DB. Let Xcvrd read the state when it requires rather than being notified. Because if we have to take the approach of notification, then we need to handle the above Xcvrd exit case separately by reading the DB which can be avoided if Xcvrd can read the DB when required.

shyam77git: @prgeor 

>  Let Xcvrd read the state when it requires rather than being notified. 

Can you clarify as to what is meant by 'when it requires'? 
Does it mean continuous polling of host-tx-ready using Redis-DB query OR what?
When interface/port moves from config shut to no shut or vice-versa, won't it a notification of host_tx_ready from redisDB otherwise how would it xcvrd be notified of this? repeated/continuous polling indefinitely without any event happening in the system is waste of cpu cycles / system resources. 
Also config interface shut/no shut are rare events along with optics removal/insertion
Also, per HLD we decided to go with notification based mechanism of host_tx_ready.
To handle xcvrd restart case, as Jagan mentioned above, better to check the host_tx_ready during module Init during insertion as source of truth.

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: @sacnaik @shyam77git please check the updated code

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: the state machine moves to CMIS_STATE_READY after this, and doesnot come again here unless the state is changed.

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: i have updated the code please check again

prgeor:
@ds952811 can you review

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: @shyam77git @jaganbal-a please take a look again as i updated the code.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -870,9 +872,11 @@ def __init__(self, port_mapping):
         self.task_process = None
         self.port_dict = {}
         self.port_mapping = copy.deepcopy(port_mapping)
+        self.xcvr_table_helper = XcvrTableHelper()
```
stephenxs, line 875: If `XcvrTableHelper` is implemented as a class member, it should be initialized in `task_worker` instead of `__init__`.
This is because `CmisManagerTask` is a subprocess wrapper, which means a subprocess will be forked once `task_run` is called.
If it is initialized in `__init__`, the flow is
- parent creates the `XcvrTableHelper` along with all the low level objects, like the sockets based on which the tables are accessed
- and then the child is forked, all resources opened by the parent are duplicated
- now there are two copies of all the resources but the parent's one will not really be used.

stephenxs: Of course, initializing it in `__init__` and then duplicating it from parent to child should also work. But it looks like taking a detour to me.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -903,9 +907,11 @@ def on_port_update_event(self, port_change_event):
             return
 
         # Skip if the port/cage type is not a CMIS
-        ptype = _wrapper_get_sfp_type(pport)
-        if ptype not in self.CMIS_MODULE_TYPES:
-            return
+        # 'index' can be -1 if STATE_DB|PORT_TABLE
+        # TODO:Make sure doesnot breakt non-CMIS
+        #ptype = _wrapper_get_sfp_type(pport)
```
jaganbal-a, line 912: I believe get_sfp_type() returns the sfp_type present in the port and not the port cage type. 
So can you rename the variable and uncomment the code?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1150,6 +1182,14 @@ def task_worker(self):
                              self.CMIS_STATE_REMOVED]:
                     continue
 
+                # Handle the case when Xcvrd was NOT running when 'host_tx_ready' or 'admin_status'
+                # was updated or this is the first run so reconcile the above two attributes
+                if 'host_tx_ready' not in self.port_dict[lport]:
```
jaganbal-a, line 1187: This code will not get hit if the port is CMIS_STATE_READY, so with config-interface shutdown the event will be returned in the above if condition and the modules will not be put in DP-DEACTIVATE. 

If any runtime Tx fault handler being implemented and the host_tx_ready is set to false, then the event will be missed.

prgeor: CMIS state reset happens if any port configuration is changed including admin or host_tx_ready status here :[https://github.com/Azure/sonic-platform-daemons/blob/master/sonic-xcvrd/xcvrd/xcvrd.py#L919](https://github.com/Azure/sonic-platform-daemons/blob/master/sonic-xcvrd/xcvrd/xcvrd.py#L919)

sonic-xcvrd/xcvrd/xcvrd.py
ds952811: get_sfp_type() returns only the port/cage type, and it's a constant initialized at chassis.__init__()
e.g.
https://github.com/Azure/sonic-buildimage/blob/master/platform/broadcom/sonic-platform-modules-dell/z9332f/sonic_platform/chassis.py#L132
And a QSFP28 could be attached to the QSFPDD port on the box, and it's better to skip the CMIS init sequence for it.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1120,11 +1129,34 @@ def test_datapath_state(self, api, channel, states):
...
+
+        found, port_info = cfg_port_tbl.get(lport)
+        if found:
+            # Check admin_status too ...just in case
+            admin_status = dict(port_info)['admin_status']
+        return admin_status
+
     def task_worker(self):
-        self.log_notice("Starting...")
+        self.log_notice("Starting CmisTaskMgr with pid:{}".format(self.task_process.pid))
```
ds952811, line 1156: If you checked the syslog of pmon#xcvrd, the pid is available over there, on the other hand, we already have the 'CMIS' prefix in the CmisManagerTask.log_notice(), hence this is not necessary

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1120,11 +1129,34 @@ def test_datapath_state(self, api, channel, states):
...
 
         # APPL_DB for CONFIG updates, and STATE_DB for insertion/removal
-        sel, asic_context = port_mapping.subscribe_port_update_event(['APPL_DB', 'STATE_DB'])
+        sel, asic_context = port_mapping.subscribe_port_update_event()
```
ds952811, line 1159: While APPL_DB is for monitoring config updates arrived at swss#orchagent, the STATE_DB is for the transceiver insertion/removal detection, and it's not necessary to have both listeners registered all the time.
May I have a clarification on why we need this?

sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
```diff
@@ -106,23 +106,21 @@ def subscribe_port_config_change():
...
+        {'STATE_DB': 'PORT_TABLE'},
+    ]
 
     sel = swsscommon.Select()
     asic_context = {}
     namespaces = multi_asic.get_front_end_namespaces()
-    for db_name in db_list:
-        if db_name not in port_tbl_map:
-            continue
+    for d in port_tbl_map:
```
ds952811, line 119: As indicated earlier, do we really want to have all these listeners registered all the time?

sonic-xcvrd/xcvrd/xcvrd.py
ds952811: It sounds like this PR will perform CMIS application selection one by one for each datapath(i.e. logical port on the MAC) instead of a one-shot configuration for all the datapatches on a transceiver.

What's unclear to me is as follows:

1. Is there a switch ASIC available to support asynchronous mixed port mode in a port macro? (i.e. Is it possible to have both 1x100G-NZR + 2x100G-PAM4 on a single QSFPDD port, from MAC perspective?)
2. Is there a CMIS transceiver available to support asynchronous mixed port mode in a single module?

In the above scenario, each datapath will require a dedicated clock divisor while they all share the same reference clock source. As far as I know, this is NOT a supported port mode in Broadcom ASIC, but I'm not sure if this is also applicable to other vendors, could you please clarify if there is a such module available for testing? Thank you

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1223,37 +1272,36 @@ def task_worker(self):
...
                         # D.1.3 Software Configuration and Initialization
                         if not api.tx_disable_channel(host_lanes, True):
                             self.log_notice("{}: unable to turn off tx power".format(lport))
                             self.port_dict[lport]['cmis_retries'] = retries + 1
                             continue
-                        api.set_lpmode(False)
 
-                        # TODO: Use fine grained time when the CMIS memory map is available
+                        # TODO: Make sure this doesn't impact other datapaths
+                        api.set_lpmode(False)
```
keboliu, line 1291: if I understand correctly, the previous  `api.set_lpmode(True)` has been removed, should this line also need to be removed?

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: @stephenxs what is the concern here? only the worker thread is going to use the socket, not the parent.

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: @ds952811 form-factor of the transceiver is not the right check to determine whether the module follows CMIS or not. Even QSFP form-factor may follow CMIS https://github.com/Azure/sonic-platform-common/pull/246 that follows CMIS

sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
prgeor: @ds952811 yes, for these events:-

1. host_tx_ready coming from PORT_TABLE of STATE_DB
2. TRANSCEIVER_INFO for detecting Xcvr presence change
3. APPL_DB for listening to configuration changes for breakout, shutdown/startup etc

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: This is a precondition for Datapath Activation and hence required.

sonic-xcvrd/xcvrd/xcvrd.py
stephenxs: `xcvr_table_helper` is initialized in the parent and then duplicated in the client. but it will never be used by the parent.
so the resource allocated in the parent is wasted.
- if it is initialized in `__init__`: 
  - `xcvr_table_helper` initialized in parent. resource allocated for it but not used.
  - duplicated in the child during fork
- if it is initialized in `task_worker`
  - `xcvr_table_helper` is initialized in the child before it starts to work. 
  - no resource was allocated in the parent.

as I mentioned, it also works in the current way. so I'm ok if you do not decide to change it.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2214,7 +2221,11 @@ def get_state_port_tbl(self, asic_id):
 
 
 def main():
-    xcvrd = DaemonXcvrd(SYSLOG_IDENTIFIER)
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--skip_cmis_mgr', action='store_true')
```
jaganbal-a, line 2225: Is this flag supposed to get updated by reading the pmon_daemon_control.json file isn't it?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1284,6 +1327,13 @@ def task_worker(self):
                 try:
                     # CMIS state transitions
                     if state == self.CMIS_STATE_INSERTED:
+                        if self.port_dict[lport]['host_tx_ready'] != 'True' or \
```
jaganbal-a, line 1330: Please change it 'true' and 'false' <small case> for host_tx_ready handling through out the file as we discussed.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -986,6 +993,10 @@ def on_port_update_event(self, port_change_event):
                 self.port_dict[lport]['speed'] = port_change_event.port_dict['speed']
```
jaganbal-a, line 993: Can you please add the invalid speed check before copying to local dictionary, this workaround is needed until the original issue is fixed <https://github.com/Azure/sonic-swss/issues/2276 >

if (port_change_event.port_dict['speed']  != N/A):
    self.port_dict[lport]['speed'] = port_change_event.port_dict['speed']

jaganbal-a (changes requested):

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -986,6 +993,10 @@ def on_port_update_event(self, port_change_event):
...
             if port_change_event.port_dict is not None and 'lanes' in port_change_event.port_dict:
                 self.port_dict[lport]['lanes'] = port_change_event.port_dict['lanes']
+            if port_change_event.port_dict is not None and 'host_tx_ready' in port_change_event.port_dict:
+                self.port_dict[lport]['host_tx_ready'] = port_change_event.port_dict['host_tx_ready']
+            if port_change_event.port_dict is not None and 'admin_status' in port_change_event.port_dict:
+                self.port_dict[lport]['admin_status'] = port_change_event.port_dict['admin_status']
             self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_INSERTED
```
jaganbal-a, line 1000: Change the CMIS state to inserted and invoke reset_cmis() only when there is a change in value for the following fields speed / lanes / host_tx_ready or admin_status.  
With current change, the cmis state is set to INSERTED even for oper_state state change in value in the DB for a Port.

jaganbal-a, line 1000: this is redundant, reset_cmis() is doing this.

sonic-xcvrd/xcvrd/xcvrd.py
jaganbal-a: tracking the original fix: https://github.com/Azure/sonic-swss/issues/2276

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: xcvrd gets this arugment from its invocation by supervisord. For eg

python3 xcvrd.py --skip_cmis_mgr

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: @ds952811 A new signal "host_tx_ready" is introduce for 400G QSFP-DD transceiver datapath activation as per this [HLD](https://github.com/sonic-net/SONiC/blob/master/doc/sfp-cmis/Interface-Link-bring-up-sequence.md) where the OA will control this signal and Xcvrd will listen to any change of this signal in STATE_DB.

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: the port change event doesn't have a mechanism to filter out the event based upon the corresponding PORT_TABLE. So, until we have this mechanism to filter out, we need this code here otherwise Xcvrd will miss optics jack out/in

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: added

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: Xcvr is lisenting to various port change events related to a ports speed, lane mapping etc and these events across various logical ports in the case of breakout port (say one physical port mapping to 4 logical ports) will arrive asynchronously, so we need to handle these events independently, even though it belong to same physical port.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -994,7 +994,10 @@ def on_port_update_event(self, port_change_event):
                 self.port_dict[lport]['lanes'] = port_change_event.port_dict['lanes']
             if 'host_tx_ready' in port_change_event.port_dict:
                 self.port_dict[lport]['host_tx_ready'] = port_change_event.port_dict['host_tx_ready']
-            if 'admin_status' in port_change_event.port_dict:
+            if 'admin_status' in port_change_event.port_dict and 'oper_status' in port_change_event.dict:
+                # At times 'admin_status' is NOT the same in the PORT_TABLE of APPL_DB and STATE_DB
+                # We dont have better way to check if 'admin_status' is from APPL_DB or STATE_DB so this
+                # check is put temporarily to list only to APPL_DB's admin_status and ignore that of STATE_DB
```
shyam77git, line 1000: instead of 'list only' it should be 'listen only'

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -994,7 +994,10 @@ def on_port_update_event(self, port_change_event):
                 self.port_dict[lport]['lanes'] = port_change_event.port_dict['lanes']
             if 'host_tx_ready' in port_change_event.port_dict:
                 self.port_dict[lport]['host_tx_ready'] = port_change_event.port_dict['host_tx_ready']
-            if 'admin_status' in port_change_event.port_dict:
+            if 'admin_status' in port_change_event.port_dict and 'oper_status' in port_change_event.dict:
```
jaganbal-a, line 997: in port_change_event.dict: -> in port_change_event.**port_dict**:

keboliu (approved):

shyam77git (approved):
LGTM

jaganbal-a:
[Link_bring_up_xcvrd_unit_test_log.txt](https://github.com/Azure/sonic-platform-daemons/files/8983598/Link_bring_up_xcvrd_unit_test_log.txt)

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -972,22 +978,28 @@ def on_port_update_event(self, port_change_event):
...
             self.port_dict[lport] = {}
 
+        if port_change_event.port_dict is None:
+            return
+
         if port_change_event.event_type == port_change_event.PORT_SET:
             if pport >= 0:
                 self.port_dict[lport]['index'] = pport
-            if port_change_event.port_dict is not None and 'speed' in port_change_event.port_dict:
+            if 'speed' in port_change_event.port_dict and port_change_event.port_dict['speed'] != 'N/A':
```
jaganbal-a, line 991: Please add a trace for each event that is processed.

jaganbal-a (approved):

sujinmkang (approved):

PR #255, opened 2022-04-15
Fix checkReplyType failed issue via recreating xcvr_table_helper on forking subprocess

stephenxs:
#### Description

Fix Azure/sonic-buildimage#10530
Fix checkReplyType failed issue via recreating xcvr_table_helper on forking subprocess

In a rare scenario, `xcvrd` can fail due to the following exception:
```
xcvrd Traceback (most recent call last):
xcvrd   File "/usr/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
xcvrd     self.run()
xcvrd   File "/usr/lib/python3.7/multiprocessing/process.py", line 99, in run
xcvrd     self._target(*self._args, **self._kwargs)
xcvrd   File "/usr/local/lib/python3.7/dist-packages/xcvrd/xcvrd.py", line 1140, in task_worker
xcvrd   File "/usr/local/lib/python3.7/dist-packages/xcvrd/xcvrd.py", line 548, in del_port_sfp_dom_info_from_db
xcvrd     dom_tbl._del(port_name)
xcvrd   File "/usr/lib/python3/dist-packages/swsscommon/swsscommon.py", line 2474, in _del
xcvrd     return self.delete(*args, **kwargs)
xcvrd   File "/usr/lib/python3/dist-packages/swsscommon/swsscommon.py", line 2468, in delete
xcvrd     return _swsscommon.Table_delete(self, *args, **kwargs)
xcvrd RuntimeError: Expected to get redis type 3 got type 5, err: NON-STRING-REPLY: Input/output error
```

Signed-off-by: Stephen Sun <stephens@nvidia.com>

#### Motivation and Context

`xcvr_table_helper` is an integrated object containing all table and database connector objects for redis database accessing which are socket-based. It is created at the beginning of the `xcvrd` daemon. The sockets will be shared across parent and children processes after children forked, which is the chief culprit of the issue.

The application communicates with redis-db server in a synchronous way, which means the message flow is like this:

    a process sends the request to redis-db server and then blocks
    the redis-db server handles and replies the request
    the process receives the reply and then continue to run

All the messages will transport via sockets.

For `xcvrd`, it works in a sparse mode which means the frequency at which it operates redis-db tables is relatively low and a child and the parent rarely send/receive messages at the same time, which means it works well even if two application processes, like a child and the parent, shares the socket via which they communicate with the redis-server in most cases.
However, it is still possible that both children and parent operates redis-db tables at the same time and messages interleave, which can cause a process receives a message which was sent to other processes sharing the socket with it. In this case, `checkReplyType` fails because the reply type is not expected.

To resolve the issue, the children should close the socket shared with the parent and recreate a new socket, avoiding sharing sockets with parent.

Regression and manually test.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1966,7 +1966,7 @@ class DaemonXcvrd(daemon_base.DaemonBase):
 
         # Post all the current interface dom/sfp info to STATE_DB
         self.log_info("Post all port DOM/SFP info to DB")
-        retry_eeprom_set = post_port_sfp_dom_info_to_db(is_warm_start, port_mapping_data, self.stop_event)
+        retry_eeprom_set = post_port_sfp_dom_info_to_db(is_warm_start, port_mapping_data, self.xcvr_table_helper, self.stop_event)
 
         # Init port sfp status table
         self.log_info("Init port sfp status table")
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -86,8 +86,6 @@ g_dict = {}
 platform_sfputil = None
 # Global chassis object based on new platform api
 platform_chassis = None
-# Global xcvr table helper
-xcvr_table_helper = None
 
 # Global logger instance for helper functions and classes
 # TODO: Refactor so that we only need the logger inherited
@@ -496,7 +494,7 @@ def post_port_dom_info_to_db(logical_port_name, port_mapping, table, stop_event=
 # Update port dom/sfp info in db
 
 
-def post_port_sfp_dom_info_to_db(is_warm_start, port_mapping, stop_event=threading.Event()):
+def post_port_sfp_dom_info_to_db(is_warm_start, port_mapping, xcvr_table_helper, stop_event=threading.Event()):
     # Connect to STATE_DB and create transceiver dom/sfp info tables
     transceiver_dict = {}
     retry_eeprom_set = set()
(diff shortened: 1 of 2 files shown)


stephenxs:
We also need this PR to go to 202111 but it can not be cherry-picked. I will open new PR to back port it once this has been approved.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1121,6 +1121,10 @@ def test_datapath_state(self, api, channel, states):
         return done
 
     def task_worker(self):
+        global xcvr_table_helper
```
prgeor, line 1124: lets NOT refer to global instance, so that we have local instance

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1442,6 +1446,10 @@ def _mapping_event_from_change_event(self, status, port_dict):
         return event
 
     def task_worker(self, stopping_event, sfp_error_event):
+        global xcvr_table_helper
```
prgeor, line 1449: lets NOT refer to global instance, so that we have local instance

prgeor (commented):
Can we also remove the global reference to 'xcvr_table_helper' in Main task 'init' function?

prgeor:
Thanks @stephenxs for finding the bug and fixing it!

sonic-xcvrd/xcvrd/xcvrd.py
stephenxs: The problem is that `xcvr_table_helper` has been referenced as a global in many places in the `xcvrd` code. Same as the previous one.

stephenxs:
> Can we also remove the global reference to 'xcvr_table_helper' in Main task 'init' function?

The problem is that `xcvr_table_helper` has been referenced as a global in many places in the `xcvrd` code. Same as the previous one.

stephenxs:
> Can we also remove the global reference to 'xcvr_table_helper' in Main task 'init' function?

Done. All global `xcvr_table_helper` has been moved to class scope.

prgeor (approved):

PR #260, opened 2022-05-02
Mem leak caused by Xcvrd in Send-Q of REDIS-DB socket connection

prgeor:
#### Description
Ensure Xcvrd close the unused socket to prevent memory leak. Unix socket cannot drain the data if the listener is NOT consuming the data so we ensure to close the unused socket by Xcvrd. Xcvrd opens the connection to listen to PORT_TABLE of APPL_DB and later on leaves the socket open.

#### Motivation and Context
Monit service complains of crossing the expected threshold for memory usage by database container as below:
```
Oct 17 01:05:10.970410 SONIC-T0 ERR monit[538]: 'container_memory_database' status failed (3) -- [database] Memory Usage (1.57391e+08 Bytes) is larger than the threshold (157286400 Bytes)!
 ```

The following shows which redis-db client is using the memory:-
 
```
GECHIANG@SONIC-T0:/var/log$ redis-cli client list | grep -v omem=0
id=613 addr=127.0.0.1:42212 fd=288 name= age=30229173 idle=27168066 flags=N db=0 sub=0 psub=1 multi=-1 qbuf=0 qbuf-free=0 obl=16384 oll=4185 omem=120041048 events=rw cmd=psubscribe
GECHIANG@SONIC-T0:/var/log$
 ```
```
GECHIANG@SONIC-T0:/var/log$ sudo netstat -antp | grep 42212
tcp        0 2547136 127.0.0.1:6379          127.0.0.1:42212         ESTABLISHED 903/redis-server 12 
tcp   1822707      0 127.0.0.1:42212         127.0.0.1:6379          ESTABLISHED 8308/python         
GECHIANG@SONIC-T0:/var/log$ sudo ps aux | grep 8308
root      8308  6.7  0.1 207788 15316 pts/0    Sl    2021 34146:26 /usr/bin/python /usr/bin/xcvrd
GECHIANG  9794  0.0  0.0  13376   976 pts/0    S+   01:38   0:00 grep 8308
GECHIANG@SONIC-T0:/var/log$
```

From above its clear Xcvrd is consuming the memory 

Create continuous port flap events for 48 hours+ that gets queued over a period of time such that the APPL_DB's PORT_TABLE receives large number of events. Ensure REDIS DB's memory usage over a period of time is NOT increasing.

### DIFF
--- a/sonic-xcvrd/scripts/xcvrd
+++ b/sonic-xcvrd/scripts/xcvrd
@@ -389,6 +389,33 @@ def waiting_time_compensation_with_sleep(time_start, time_to_wait):
     if time_diff < time_to_wait:
         time.sleep(time_to_wait - time_diff)
 
+def wait_for_all_ports_configured():
+    # Connect to APPL_DB and subscribe to PORT table notifications
+    appl_db = swsscommon.DBConnector(swsscommon.APPL_DB,
+                                     REDIS_HOSTNAME,
+                                     REDIS_PORT,
+                                     REDIS_TIMEOUT_MSECS)
+    sel = swsscommon.Select()
+    sst = swsscommon.SubscriberStateTable(appl_db, swsscommon.APP_PORT_TABLE_NAME)
+    sel.addSelectable(sst)
+
+    # Make sure this daemon started after all port configured.
+    while XCVRD_MAIN_TASK_RUNNING_FLAG:
+        (state, c) = sel.select(SELECT_TIMEOUT_MSECS)
+        if state == swsscommon.Select.TIMEOUT:
+            continue
+        if state != swsscommon.Select.OBJECT:
+            log_warning("sel.select() did not return swsscommon.Select.OBJECT")
+            continue
+
+        (key, op, fvp) = sst.pop()
+        if key in ["PortConfigDone", "PortInitDone"]:
+            break
+    # Close the unused socket to prevent mem-leak. Since Xcvrd opens socket to subscribe to
+    # APP_DB's PORT_TABLE events, if Xcvrd doesnot listen on the opened socket, then over a
+    # period of time, redis-DB Send-Q will accumulate events(Eg due to continous port flap)
+    # that can grow up slowly over a period of time resulting in memory leak.
+    sel.removeSelectable(sst)
 
 # Timer thread wrapper class to update dom info to DB periodically
 class dom_info_update_task:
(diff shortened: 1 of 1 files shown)


sonic-xcvrd/scripts/xcvrd
```diff
@@ -389,6 +389,33 @@ def waiting_time_compensation_with_sleep(time_start, time_to_wait):
...
+            continue
+
+        (key, op, fvp) = sst.pop()
+        if key in ["PortConfigDone", "PortInitDone"]:
+            break
+    # Close the unused socket to prevent mem-leak. Since Xcvrd opens socket to subscribe to
+    # APP_DB's PORT_TABLE events, if Xcvrd doesnot listen on the opened socket, then over a
+    # period of time, redis-DB Send-Q will accumulate events(Eg due to continous port flap)
+    # that can grow up slowly over a period of time resulting in memory leak.
+    sel.removeSelectable(sst)
```
qiluo-msft, line 418: >sel.removeSelectable(sst) [](http://example.com/codeflow?start=4&length=25)

Is it the most critical fix?
I am considering enforce it to all the use case by default.

prgeor: The critical one is the ```sel``` object at line 398 which gets deleted after the function completes which results in socket getting closed and the Send-Q buffer cleared. But it's good to call removeSelectable() as a clean up before object gets deleted.

sonic-xcvrd/scripts/xcvrd
```diff
@@ -389,6 +389,33 @@ def waiting_time_compensation_with_sleep(time_start, time_to_wait):
     if time_diff < time_to_wait:
         time.sleep(time_to_wait - time_diff)
 
+def wait_for_all_ports_configured():
+    # Connect to APPL_DB and subscribe to PORT table notifications
```
prsunny, line 393: This table is not intended to be subscribed outside of orchagent.

prgeor: This subscription is NOT recent changes. Its existing since 201811.

qiluo-msft (approved):
LGTM. Please check with other reviewers.

sonic-xcvrd/scripts/xcvrd
prsunny: This can potentially break orchagent functionality. We need to revisit this and not add more case.

prsunny (changes requested):
As comments

sonic-xcvrd/scripts/xcvrd
prsunny: Unblocking as this is already existing code. Let's eventually move this subscription to STATE_DB instead.

prsunny (approved):

PR #264, opened 2022-06-08
[sonic-ycabled] fix grpc logic for timeout,cli HWSTATUS value retrival logic for active-active cable

vdahiya12:
This PR adds some improvement logic to ycabled daemon

1. correct the loopback IP's for active-active cable type for libra gRPC read_side establishment
2. improves the code for timeout value injection for different RPC's for gRPC
3. for cli HW_STATUS the daemon tries to use cached old value, if not present try to do RPC for getting ForwardingState with timeout of 0.1 ms

#### Motivation and Context
Required for gRPC logic improvement in few active-active scenario's

UT and deploying the changes in actual testbed


lolyu (dismissed):

zjswhhh (approved):
lgtm

PR #268, opened 2022-06-28
[ycabled] add some retry logic for gRPC channel setup;fix no channel gRPC notification

vdahiya12:

This PR adds some retry logic for setting up channels for gRPC with a prolonged time period
This PR also fixes the gRPC notification handling, when a request come for `appl DB:HW_MUX_CABLE_TABLE` ->
`state DB:HW_MUX_CABLE_TABLE`, previously if the channel was not setup, the daemon does not give back a response, with this change this condition is properly handled.
This PR also enhances the channel to keep a keepalive message between server/soc and ycabled when the connections are idle. This enhancement comes from  https://github.com/grpc/grpc/blob/master/doc/keepalive.md this feature in gRPC


vdahiya12:
https://github.com/Azure/sonic-platform-daemons/pull/269

PR #273, opened 2022-07-13
[ycabled] remove some redundant logging for active-active cable type

vdahiya12:

#### Description
This change is required for stop posting `MUX_CABLE_INFO` entries into state DB for `active-active` cable_type since for these cables there is no i2c/eeprom or muxcable. 
This loop is independent of main loop, hence it needs to be changed 

#### Motivation and Context
remove some unnecessary logs

Unit-tests


vdahiya12:
closed in favor #274

PR #286, opened 2022-08-23
Use get() to fetch default value from dictionary for port admin_status 

anamehra:
Signed-off-by: anamehra <anamehra@cisco.com>

#### Description
    The config db dictionary may not have admin_status available for a
port if the same is not set in config db file. Use get() function to
prevent crashes and set the default status to 'down'.

#### Motivation and Context

xcvrd crash is observed during boot on multi-asic chassis system.
The crash is reported in https://github.com/sonic-net/sonic-buildimage/issues/11707

On debugging the issue, found that at the point of the crash, the data being used is from config db.
It was observed that the Ethernet ports, which are not configured for admin_state field explicitly in the config  is causing the crash. This is a valid config as admin_status is not a mandatory field and is considerd by default 'down' if missing in port config in config db file.

for example, the Ethernet31 is configured for a far-end, and has this field populated while Ethernet32, as was not being used, has this field missing.

    "Ethernet31": {
        "admin_status": "up",
        "alias": "Eth0/2/31",
        "asic_port_name": "Eth31-ASIC2",
        "description": "ARISTA32T3:Ethernet1",
        "fec": "rs",
        "index": "31",
        "lanes": "512,513,514,515",
        "mtu": "9100",
        "pfc_asym": "off",
        "role": "Ext",
        "speed": "100000",
        "tpid": "0x8100"
    },
    "Ethernet32": {
        "alias": "Eth0/2/32",
        "asic_port_name": "Eth32-ASIC2",
        "description": "Eth0/2/32",
        "fec": "rs",
        "index": "32",
        "lanes": "264,265,266,267",
        "mtu": "9100",
        "pfc_asym": "off",
        "role": "Ext",
        "speed": "100000",
        "tpid": "0x8100"
    },

closes https://github.com/sonic-net/sonic-buildimage/issues/11707

Loaded image on single and multi-asic system. Tested for normal boot. No crash observed.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1253,7 +1253,7 @@ class CmisManagerTask:
         found, port_info = cfg_port_tbl.get(lport)
         if found:
             # Check admin_status too ...just in case
-            admin_status = dict(port_info)['admin_status']
+            admin_status = dict(port_info).get('admin_status', 'down')
         return admin_status
 
     def configure_tx_output_power(self, api, lport, tx_power):


anamehra:
@prgeor , @abdosi , please review.

prgeor:
@anamehra can you please check your minigraph why 'admin_status' is missing for some ports?

anamehra:
> @anamehra can you please check your minigraph why 'admin_status' is missing for some ports?

@abdosi , could you please comment on minigraph?
@prgeor , a config db json with Ethernet ports missing admin_status should be a valid config. If admin_status is not defined, its assumed as down. It has been the case always. In the fix, I am just handling the same scenario.  If a dictionary entry is missing, use the default instead of crashing.

abdosi (approved):

abdosi:
> > @anamehra can you please check your minigraph why 'admin_status' is missing for some ports?
> 
> @abdosi , could you please comment on minigraph? @prgeor , a config db json with Ethernet ports missing admin_status should be a valid config. If admin_status is not defined, its assumed as down. It has been the case always. In the fix, I am just handling the same scenario. If a dictionary entry is missing, use the default instead of crashing.

@prgeor i checked for the ports that do not have any neighbor associated minigraph parser do not assign `admin_status` attribute in port table. We can fix minigraph parser to always have `admin_status`  attribute but I think this PR change is good and is independent of minigraph parser change

PR #288, opened 2022-08-25
[PSU daemon] Support PSU power threshold checking

stephenxs:
#### Description

Support PSU power threshold checking
- Two platform APIs are introduced to represent the warning and critical thresholds of a PSU's power
- In the main loop
  - Whenever a PSU becomes good, PSU daemon tries calling platform APIs to fetch both thresholds. If `None` is returned or `NotImplemented` is thrown by either API, the PSU power threshold checking will not be performed for the PSU
  - the PSU daemon compares the power with thresholds, exposing status to the database, and logging messages.

#### Motivation and Context

To notify the user if the PSU's power exceeds the threshold.

Unit test cases have been implemented.

### DIFF
--- a/sonic-psud/scripts/psud
+++ b/sonic-psud/scripts/psud
@@ -529,7 +529,11 @@ class DaemonPsud(daemon_base.DaemonBase):
         power_threshold = try_get(psu.get_psu_power_threshold, NOT_AVAILABLE)
         power_critical_threshold = try_get(psu.get_psu_power_critical_threshold, NOT_AVAILABLE)
         if psu_status.check_psu_power_threshold:
-            if psu_status.power_exceeded_threshold:
+            if power_threshold == NOT_AVAILABLE or power_critical_threshold == NOT_AVAILABLE:
+                self.log_error("PSU power thresholds become invalid: threshold {} critical threshold {}".format(power_threshold, power_critical_threshold))
+                psu_status.check_psu_power_threshold = False
+                psu_status.power_exceeded_threshold = False
+            elif psu_status.power_exceeded_threshold:
                 # The failing threshold is the warning threshold
                 if power < power_threshold:
                     # Clear alarm
--- a/sonic-psud/scripts/psud
+++ b/sonic-psud/scripts/psud
@@ -352,7 +352,7 @@ class PsuStatus(object):
         return True
 
     def is_ok(self):
-        return self.presence and self.power_good and self.voltage_good and self.temperature_good and not self.power_exceeded_threshold
+        return self.presence and self.power_good and self.voltage_good and self.temperature_good
 
 
 #
--- a/sonic-psud/scripts/psud
+++ b/sonic-psud/scripts/psud
@@ -58,7 +58,7 @@ PSU_INFO_VOLTAGE_MIN_TH_FIELD = 'voltage_min_threshold'
 PSU_INFO_CURRENT_FIELD = 'current'
 PSU_INFO_POWER_FIELD = 'power'
 PSU_INFO_POWER_OVERLOAD = 'power_overload'
-PSU_INFO_POWER_WARNING_THRESHOLD = 'power_warning_threshold'
+PSU_INFO_POWER_WARNING_SUPPRESS_THRESHOLD = 'power_warning_suppress_threshold'
 PSU_INFO_POWER_CRITICAL_THRESHOLD = 'power_critical_threshold'
 PSU_INFO_FRU_FIELD = 'is_replaceable'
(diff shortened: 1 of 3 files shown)


Junchao-Mellanox (dismissed):

keboliu (approved):

sonic-psud/scripts/psud
```diff
@@ -57,6 +57,9 @@ PSU_INFO_VOLTAGE_MAX_TH_FIELD = 'voltage_max_threshold'
 PSU_INFO_VOLTAGE_MIN_TH_FIELD = 'voltage_min_threshold'
 PSU_INFO_CURRENT_FIELD = 'current'
 PSU_INFO_POWER_FIELD = 'power'
+PSU_INFO_POWER_OVERLOAD = 'power_overload'
+PSU_INFO_POWER_WARNING_THRESHOLD = 'power_warning_threshold'
```
prgeor, line 61: 'warning' can be misleading here as no warning is raised if this threshold is crossed. can we rename it as 'warning-suppress' levle?

sonic-psud/scripts/psud
```diff
@@ -495,13 +509,46 @@ class DaemonPsud(daemon_base.DaemonBase):
...
+                # Initialize power exceeding threshold state in this case
+                if (try_get(psu.get_psu_power_critical_threshold) and try_get(psu.get_psu_power_warning_threshold) and power != NOT_AVAILABLE):
+                    psu_status.check_psu_power_threshold = True
+
+        power_exceeded_threshold = psu_status.power_exceeded_threshold
+        power_warning_threshold = try_get(psu.get_psu_power_warning_threshold, NOT_AVAILABLE)
+        power_critical_threshold = try_get(psu.get_psu_power_critical_threshold, NOT_AVAILABLE)
+        if psu_status.check_psu_power_threshold:
+            if power_warning_threshold == NOT_AVAILABLE or power_critical_threshold == NOT_AVAILABLE:
+                self.log_error("PSU power thresholds become invalid: threshold {} critical threshold {}".format(power_warning_threshold, power_critical_threshold))
```
prgeor, line 533: if the platform does not support these thresholds, why syslog error every time?

sonic-psud/scripts/psud
```diff
@@ -495,13 +509,46 @@ class DaemonPsud(daemon_base.DaemonBase):
...
+                psu_status.check_psu_power_threshold = False
+                psu_status.power_exceeded_threshold = False
+            elif psu_status.power_exceeded_threshold:
+                # The failing threshold is the warning threshold
+                if power < power_warning_threshold:
+                    # Clear alarm
+                    power_exceeded_threshold = False
+            else:
+                # The rising threshold is the critical threshold
+                if power >= power_critical_threshold:
```
prgeor, line 543: 'power' is current power?

sonic-psud/scripts/psud
```diff
@@ -495,13 +509,46 @@ class DaemonPsud(daemon_base.DaemonBase):
...
+                    # Raise alarm
+                    power_exceeded_threshold = True
+
+            if psu_status.set_power_exceed_threshold(power_exceeded_threshold):
+                log_on_status_changed(self, not psu_status.power_exceeded_threshold,
```
prgeor, line 548: no LED status indicator for critical power threshold crossing?

prgeor (commented):
@stephenxs could you run sonic-mgmt test on a platform that does not support these new thresholds?

sonic-psud/scripts/psud
stephenxs: This is not the case that the platform does not support the thresholds.
If the platform doesn't support the thresholds, it is handled in [L525](https://github.com/sonic-net/sonic-platform-daemons/pull/288/files/4ee8a218b1cd1cd351b5dabe47d5f4308b1cc09e#diff-738ebd7d7e53d61f937ba7c9933d45bfa18cd8619edeec182fd66901e3c72a96R525) and no error message will be logged
It is the case that the platform supported thresholds when the PSU was plugged but it doesn't support them now. We treat it as an error in the platform API

sonic-psud/scripts/psud
stephenxs: yes.

sonic-psud/scripts/psud
stephenxs: No. System LED will be changed in the system health daemon.

sonic-psud/scripts/psud
stephenxs: Done. Changed to `power_warning_suppress_threshold`

prgeor (approved):

Issue #290, opened 2022-09-02
CMIS FSM (state machine) timer to be fetched fom optics EEPROM instead of a hard-coded value


shyam77git:
While validating 400G ZR optical module, found an issue in CMIS FSM whereby it timed-out on 1 minute hard-coded value. 
Optical module's DPinit timer / timeout value differs based on optics type/variant and  is defined in the spec (byte 144) - lower nibble (0-3) for DPinit timeout.

Fetch this value from the optics EEPROM along with optics variant/type in CMIS FSM instead of present hard-coded value

shyam77git:
https://github.com/sonic-net/sonic-platform-common/pull/312 is fixing this.
Follow comments in #312 to see if anything more required from xcvrApi/CMIS spec etc. standpoint to take care of this issue besides #312.
Note that https://github.com/sonic-net/sonic-platform-daemons/issues/294 to take care for xcvrd side changes and is separate.

PR #285, opened 2022-08-18
[Xcvrd] Soak duplicate events and process only updated interested events

prgeor:
#### Description
Xcvrd should be listening to 
CONFIG_DB - For changes to speed, lane mapping, tx_power, frequency, admin_status
STATE_DB - For optics insertion/removal, host_tx_ready
APPL_DB - only during start to know if PortConfigDone is completed

Duplicate events are now soaked such that only the last event in a barrage of duplicate events is considered for further processing. Also, if a filter is installed then filter events having fields of interest that got changed from last time.

#### Motivation and Context
Portmgrd is writing unmodified fields to the APL_DB whenever there is configuration change resulting in too many event updates causing CMIS task manager that listens to these events to be woken up and restart the state machine unnecessarily. 

There can be events like port operationally going up in STATE_DB's PORT_TABLE that can restart the CMIS state machine while the Rx and Tx datapaths are provisioned independently.

#### How Has This Been Tested?
Verified 400G ZR and 400G DR4 optics are initializing fine

PR #293, opened 2022-09-16
Fix xcvrd to support 400G ZR optic

abohanyang:
Fix Xcvrd code to support 400G ZR optic

#### Description
While bringing the 400G ZR module, I found an issue where some 400G ZR optics may take more than 60 secs to finish up the Datapath initialization. This is a problem because the xcvrd uses a hardcoded 60 secs Datapath initialization timeout for all CMIS modules and apparently, this 60 secs timeout is not sufficient for the 400G ZR module.
The CMIS spec has Datapath init/de-init timeout values defined in Byte 144. We can use these values from the EEPROM instead of presenting any hardcoded values.

#### Motivation and Context
This is needed to bring up 400G ZR module.

These changes are tested against arista_7800r3a_36d2_lc and arista_7280cr3_32p4 platforms.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1637,7 +1637,7 @@ class DomInfoUpdateTask(object):
         self.task_stopping_event.set()
         self.task_thread.join()
 
-    def on_port_config_change(self, stopping_event, port_change_event):
+    def on_port_config_change(self, port_change_event):
         if port_change_event.event_type == port_mapping.PortChangeEvent.PORT_REMOVE:
             self.on_remove_logical_port(port_change_event)
         self.port_mapping.handle_port_change_event(port_change_event)
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1490,7 +1490,9 @@ class CmisManagerTask:
                         # TODO: Make sure this doesn't impact other datapaths
                         api.set_lpmode(False)
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_AP_CONF
-                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.get_cmis_dp_deinit_duration_secs(api))
+                        dpDeinitDuration = self.get_cmis_dp_deinit_duration_secs(api)
+                        self.log_notice("{} DpDeinit duration {} secs".format(lport, dpDeinitDuration))
+                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=dpDeinitDuration)
                     elif state == self.CMIS_STATE_AP_CONF:
                         # TODO: Use fine grained time when the CMIS memory map is available
                         if not self.check_module_state(api, ['ModuleReady']):
@@ -1553,7 +1555,9 @@ class CmisManagerTask:
 
                         # D.1.3 Software Configuration and Initialization
                         api.set_datapath_init(host_lanes)
-                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.get_cmis_dp_init_duration_secs(api))
+                        dpInitDuration = self.get_cmis_dp_init_duration_secs(api)
+                        self.log_notice("{} DpInit duration {} secs".format(lport, dpInitDuration))
+                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=dpInitDuration)
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_DP_TXON
                     elif state == self.CMIS_STATE_DP_TXON:
                         if not self.check_datapath_state(api, host_lanes, ['DataPathInitialized']):
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1090,7 +1090,10 @@ class CmisManagerTask:
         return (appl_code & 0xf)
 
     def get_cmis_dp_init_duration(self, api):
-        return max(api.get_datapath_init_duration()/1000, self.CMIS_DEF_EXPIRED)
+        return api.get_datapath_init_duration()/1000
+
+    def get_cmis_dp_deinit_duration(self, api):
+        return api.get_datapath_deinit_duration()/1000
 
     def is_cmis_application_update_required(self, api, channel, speed):
         """
@@ -1194,6 +1197,32 @@ class CmisManagerTask:
                 break
 
         return done
+
+    def check_datapath_init_pending(self, api, channel):
+        """
+        Check if the CMIS datapath init is pending
+
+        Args:
+            api:
+                XcvrApi object
+            channel:
+                Integer, a bitmask of the lanes on the host side
+                e.g. 0x5 for lane 0 and lane 2.
+
+        Returns:
+            Boolean, true if any lanes are pending datapath init, otherwise false
+        """
+        pending = False
+        dpinit_pending_dict = api.get_dpinit_pending()
+        for lane in range(self.CMIS_NUM_CHANNELS):
+            if ((1 << lane) & channel) == 0:
+                continue
+            key = "DPInitPending{}".format(lane + 1)
+            if dpinit_pending_dict[key]:
+                pending = True
+                break
+
+        return pending
 
     def check_datapath_state(self, api, channel, states):
         """
@@ -1477,7 +1506,7 @@ class CmisManagerTask:
                         # TODO: Make sure this doesn't impact other datapaths
                         api.set_lpmode(False)
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_AP_CONF
-                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.get_cmis_dp_init_duration(api))
+                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.get_cmis_dp_deinit_duration(api))
                     elif state == self.CMIS_STATE_AP_CONF:
                         # TODO: Use fine grained time when the CMIS memory map is available
                         if not self.check_module_state(api, ['ModuleReady']):
@@ -1513,8 +1542,11 @@ class CmisManagerTask:
                             self.force_cmis_reinit(lport, retries + 1)
                             continue
 
-                        # TODO: Use fine grained time when the CMIS memory map is available
-                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.get_cmis_dp_init_duration(api))
+                        if self.check_datapath_init_pending(api, host_lanes):
+                            self.log_notice("{}: datapath init pending".format(lport))
+                            self.force_cmis_reinit(lport, retries + 1)
+                            continue
+
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_DP_INIT
                     elif state == self.CMIS_STATE_DP_INIT:
                         if not self.check_config_error(api, host_lanes, ['ConfigSuccess']):
@@ -1534,7 +1566,6 @@ class CmisManagerTask:
 
                         # D.1.3 Software Configuration and Initialization
                         api.set_datapath_init(host_lanes)
-                        # TODO: Use fine grained timeout when the CMIS memory map is available
                         self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.get_cmis_dp_init_duration(api))
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_DP_TXON
                     elif state == self.CMIS_STATE_DP_TXON:
@@ -1956,7 +1987,7 @@ class SfpStateUpdateTask(object):
         self.task_stopping_event.set()
         os.kill(self.task_process.pid, signal.SIGKILL)
 
-    def on_port_config_change(self, stopping_event, port_change_event):
+    def on_port_config_change(self , port_change_event):
         if port_change_event.event_type == port_mapping.PortChangeEvent.PORT_REMOVE:
             self.on_remove_logical_port(port_change_event)
             self.port_mapping.handle_port_change_event(port_change_event)
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -986,38 +986,22 @@ class CmisManagerTask:
             return
 
         if port_change_event.event_type == port_change_event.PORT_SET:
-            need_update = False
             if pport >= 0:
-                if self.port_dict[lport].get('index') != pport:
-                    self.port_dict[lport]['index'] = pport
-                    need_update = True
+                self.port_dict[lport]['index'] = pport
             if 'speed' in port_change_event.port_dict and port_change_event.port_dict['speed'] != 'N/A':
-                if self.port_dict[lport].get('speed') != port_change_event.port_dict['speed']:
-                    self.port_dict[lport]['speed'] = port_change_event.port_dict['speed']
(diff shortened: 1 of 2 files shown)


abohanyang:
@arlakshm

prgeor:
@abohanyang can you please check if on_port_update_event() is triggered even after this PR [https://github.com/sonic-net/sonic-platform-daemons/pull/285](https://github.com/sonic-net/sonic-platform-daemons/pull/285). If so please share the syslog file.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1614,7 +1637,7 @@ def task_stop(self):
         self.task_stopping_event.set()
         self.task_thread.join()
 
-    def on_port_config_change(self, port_change_event):
+    def on_port_config_change(self, stopping_event, port_change_event):
```
prgeor, line 1640: where is 'stopping_even' used in this function? can you point me to the caller of this function where its passing 'stopping_even' argument?

prgeor (changes requested):

sonic-xcvrd/xcvrd/xcvrd.py
abohanyang: Hi Prince - thank you for taking a look. On line 1756, 
"port_change_event_handler = functools.partial(self.on_port_config_change, stopping_event)"

adds an argument, stopping_event, to this handler function. when port_change_event_handler gets called in read_port_config_change (xcvrd_utilities/port_mapping.py), it actually calls self.on_port_config_change(stopping_event,port_change_event).

abohanyang: I guess I know why you are confused about my change. I just notice that I did not patch the code correctly. You can check my latest diff.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -923,6 +923,7 @@ class CmisManagerTask:
 
     CMIS_MAX_RETRIES     = 3
     CMIS_DEF_EXPIRED     = 60 # seconds, default expiration time
+    CMIS_DEF_EXPIRED_ZR  = 200 # seconds, expiration time for ZR module
```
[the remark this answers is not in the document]
shyam77git: Following PRs were raised:
https://github.com/sonic-net/sonic-platform-daemons/issues/290
https://github.com/sonic-net/sonic-platform-daemons/issues/294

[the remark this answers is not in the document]
abohanyang: To read CMIS dp state duration, I create another PR in sonic-platform-common (https://github.com/sonic-net/sonic-platform-common/pull/312).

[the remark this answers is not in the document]
shyam77git: @abohanyang  - How is #312 (above) different from #290 ?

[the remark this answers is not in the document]
shyam77git: Hi @abohanyang 
Is #312 sufficient to take care of what's mentioned in #290 or anything more required?
Note that #294 to take care for xcvrd side changes.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1516,7 +1539,7 @@ def task_worker(self):
                         # D.1.3 Software Configuration and Initialization
                         api.set_datapath_init(host_lanes)
                         # TODO: Use fine grained timeout when the CMIS memory map is available
-                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.CMIS_DEF_EXPIRED)
+                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.get_cmis_expired(api))
```
shyam77git, line 1542: @abohanyang : get_cmis_expired() - is this a temporary function defined to validate this fix - #293?

practically,  get_datapath_init_duration() and  get_datapath_init_duration() and get_datapath_deinit_duration() to be invoked instead. 
Would you be taking care of this as part of this PR? or shall we take care of this in xcvrd.py via #294 ?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1937,7 +1960,7 @@ def task_stop(self):
         self.task_stopping_event.set()
         os.kill(self.task_process.pid, signal.SIGKILL)
 
-    def on_port_config_change(self , port_change_event):
+    def on_port_config_change(self, stopping_event, port_change_event):
```
shyam77git, line 1963: @abohanyang 
Per the description, "on_port_config_change takes three arguments, not two. This issue is found when the port breakout mode changes"
Have a Q: 'stopping event' is passed as an argument but not used/consumed?
How is this related to port breakout changes? what's missing/expected?
the platform/board you are finding & validating 'port breakout change' is PHY-less (i.e. does not have Ext PHY (gearbox))?

sonic-xcvrd/xcvrd/xcvrd.py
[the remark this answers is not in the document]
abohanyang: No. @312 contains the implementation of API, get_datapath_init_duration. I just commit more changes in this PR.

sonic-xcvrd/xcvrd/xcvrd.py
abohanyang: Yes. I just commit more changes in this PR to call get_datapath_init_duration and get the expiration time.

sonic-xcvrd/xcvrd/xcvrd.py
abohanyang: Yes, this stopping_event is passed and is not used at all. I actually don't know why we have this stopping_event. I guess this stopping_event will be used to handle some error cases.

My understanding is that this on_port_config_change function is triggered during the port breakout change since the old port is removed (PORT_REMOVE) from the port_mapping and the new port is added (PORT_ADD) to the port_mapping. 

My platform is a PHY-less platform. DNX NIF is connected to the OSFP/QSFP-DD front panel directly.

sonic-xcvrd/xcvrd/xcvrd.py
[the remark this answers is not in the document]
shyam77git: > No. 312 contains the implementation of API, get_datapath_init_duration. I just commit more changes in this PR.

Thanks for the clarification. 
This PR now is going to take care of git issue #293 - I'm reviewing this further

sonic-xcvrd/xcvrd/xcvrd.py
shyam77git: > Yes. I just commit more changes in this PR to call get_datapath_init_duration and get the expiration time.

OK.
@abohanyang : can you please attach UT cases, logs for this case (400G ZR optics FSM working via xcrd.py as orchestrator with these timeouts all the way to interface in 'operational up' state)

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1458,7 +1477,7 @@ def task_worker(self):
                         # TODO: Make sure this doesn't impact other datapaths
                         api.set_lpmode(False)
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_AP_CONF
-                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.CMIS_DEF_EXPIRED)
+                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.get_cmis_dp_init_duration(api))
```
shyam77git, line 1480: this block is to handle CMIS DP_DEINIT state.
Instead of get_cmis_dp_init_duration(api) there should be function get_cmis_dp_deinit_duration(api) defined and invoked here?

DP_PATH_DEINIT_DURATION is defined at the bit range 4-8 vs DP_PATH_INIT_DURATION is defined at bit range 0-4 per CMIS spec (cmis.py), so values could be different.

sonic-xcvrd/xcvrd/xcvrd.py
shyam77git: > Yes, this stopping_event is passed and is not used at all. I actually don't know why we have this stopping_event. I guess this stopping_event will be used to handle some error cases.
> 

If unclear, can we avoid adding this change to this PR?

> My understanding is that this on_port_config_change function is triggered during the port breakout change since the old port is removed (PORT_REMOVE) from the port_mapping and the new port is added (PORT_ADD) to the port_mapping.
> 

Would you be validating port breakout as part of this PR? If not, then IMO, this should be handled in separate PR

> My platform is a PHY-less platform. DNX NIF is connected to the OSFP/QSFP-DD front panel directly.
Thanks for clarifying.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1495,7 +1514,7 @@ def task_worker(self):
                             continue
 
                         # TODO: Use fine grained time when the CMIS memory map is available
```
prgeor, line 1516: TODO not required now

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1516,7 +1535,7 @@ def task_worker(self):
                         # D.1.3 Software Configuration and Initialization
                         api.set_datapath_init(host_lanes)
                         # TODO: Use fine grained timeout when the CMIS memory map is available
```
prgeor, line 1537: TODO not required now

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1495,7 +1514,7 @@ def task_worker(self):
                             continue
 
                         # TODO: Use fine grained time when the CMIS memory map is available
-                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.CMIS_DEF_EXPIRED)
+                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.get_cmis_dp_init_duration(api))
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_DP_INIT
```
prgeor, line 1518: We actually dont need this because at this stage DpInit is not triggered, it just we selected the application so we should check the DpInitPending bit at Page 11h Byte 235

prgeor (approved):

sonic-xcvrd/xcvrd/xcvrd.py
abohanyang: Okay. I can create a separate PR to fix this.

sonic-xcvrd/xcvrd/xcvrd.py
abohanyang: I can add code to check DpInitPending additionally but I am not sure if we can simply remove this line. If the module never sets the pending bits somehow, then it will get stuck in the CMIS_STATE_AP_CONF state.

sonic-xcvrd/xcvrd/xcvrd.py
abohanyang: [log.txt](https://github.com/sonic-net/sonic-platform-daemons/files/9709709/log.txt)

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1073,6 +1089,12 @@ def get_cmis_application_desired(self, api, channel, speed):
 
         return (appl_code & 0xf)
 
+    def get_cmis_dp_init_duration(self, api):
+        return api.get_datapath_init_duration()/1000
+
+    def get_cmis_dp_deinit_duration(self, api):
+        return api.get_datapath_deinit_duration()/1000
```
prgeor, line 1096: can you specify return value duration msec/sec/min?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1176,6 +1198,32 @@ def check_config_error(self, api, channel, states):
...
+        for lane in range(self.CMIS_NUM_CHANNELS):
+            if ((1 << lane) & channel) == 0:
+                continue
+            key = "DPInitPending{}".format(lane + 1)
+            if dpinit_pending_dict[key]:
+                pending = True
+                break
+
+        return pending
+
```
prgeor, line 1226: I think we should check if DpInitPending is set for ALL host lanes and return TRUE. Return False if any of the lanes has DpInitPending=0


image

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1494,8 +1542,11 @@ def task_worker(self):
                             self.force_cmis_reinit(lport, retries + 1)
                             continue
 
-                        # TODO: Use fine grained time when the CMIS memory map is available
-                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.CMIS_DEF_EXPIRED)
+                        if self.check_datapath_init_pending(api, host_lanes):
```
prgeor, line 1545: if any of the lanes has DPInitPending=0 i.e if this fn call return False then we should retry re-init

abohanyang: I have updated the code based on our discussion.

abohanyang:
Hi @shyam77git - The code is pending review. I am waiting for @prgeor  comments/approval on my latest commit.

prgeor:
> Fix Xcvrd code to support 400G DR and 400G ZR optics
> 
> #### Description
> 1. Increase the initialization timeout value for the 400GBASE-ZR module as 400GBASE-ZR may take more than 60 sec to come up. Based on the spec,  ZR may take up to 200 sec to become ready.
> 2. on_port_update_event is triggered when 400G optics enter the ready state. To avoid the infinite initialization loop, in on_port_update_event, call force_cmis_reinit only when port configuration changes.
> 3. on_port_config_change takes three arguments, not two. This issue is found when the port breakout mode changes.
> 
> #### Motivation and Context
> * Item 1 and Item 2  are needed to bring up 400G ZR/DR optics.
> * Item 3 fixes a potential issue during dynamic breakout mode change.
> 
> #### How Has This Been Tested?
> * These changes are tested against arista_7800r3a_36d2_lc and arista_7280cr3_32p4 platforms.

@abohanyang can you check the description. I think you are addressing only #1

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1458,7 +1490,7 @@ def task_worker(self):
                         # TODO: Make sure this doesn't impact other datapaths
                         api.set_lpmode(False)
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_AP_CONF
-                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.CMIS_DEF_EXPIRED)
+                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.get_cmis_dp_deinit_duration_secs(api))
```
prgeor, line 1493: can you log the deinit duration?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1515,8 +1553,7 @@ def task_worker(self):
 
                         # D.1.3 Software Configuration and Initialization
                         api.set_datapath_init(host_lanes)
-                        # TODO: Use fine grained timeout when the CMIS memory map is available
-                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.CMIS_DEF_EXPIRED)
+                        self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=self.get_cmis_dp_init_duration_secs(api))
```
prgeor, line 1556: can you please log the dp_init pending duration?

sonic-xcvrd/xcvrd/xcvrd.py
abohanyang: Sure. I just update the code and the summary of this PR.

prgeor (approved):

prgeor:
@yxieca can you help cherry-pick to 202205?

PR #295, opened 2022-09-20
[ycabled] add notification for gRPC connection state transitions to  IDLE/TRANSIENT_FAILURE

vdahiya12:
For the cases where gRPC connectivity to server becomes IDLE/TRANSIENT_FAILURE, the ToR should periodically query/retry establish admin state and revert the Forwarding state back to active-active for both the T0's
This PR tries to attain that via adding a transient failure message to APP DB when connectivity is lost, so each time connectivity state changes back to not desired state(IDLE/TRANSIENT_FAILURE) we log a message, and linkmgr will query again to get gRPC in sync

Unit-tests and deploying changes to testbed


zjswhhh (approved):
lgtm

zjswhhh (approved):

PR #300, opened 2022-09-27
Remove shell=True

maipbui:
Signed-off-by: maipbui <maibui@microsoft.com>

#### Description

`subprocess()` - use `shell=False`, use list of strings Ref: [https://semgrep.dev/docs/cheat-sheets/python-command-injection/#mitigation](https://semgrep.dev/docs/cheat-sheets/python-command-injection/#mitigation)
#### Motivation and Context

`subprocess()` - when using with `shell=True` is dangerous. Using subprocess function without a static string can lead to command injection.


qiluo-msft (approved):

prgeor (approved):

prgeor:
@maipbui please fix the code coverage

PR #301, opened 2022-09-28
[ycabled] add some exception catching logic to some vendor specific API's

vdahiya12:

This PR adds a try/catch block for some abstract muxcable API's. In particular the exception logic is added for all the API's where there is a possibility for exceptions to be passed by Vendor API's implementation, in this regard ycabled will have all abstract muxcable API's covered by exception logic with this PR.
If the exception is caught it will be just logged and daemon will resume its normal operation.

Unit-tests and deploying changes on testbed

### DIFF
--- a/sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
+++ b/sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
@@ -895,7 +895,11 @@ def toggle_mux_tor_direction_and_update_read_side(state, logical_port_name, phys
         helper_logger.log_error("Error: Could not get port instance for read side for while processing a toggle Y cable port {} {}".format(physical_port, threading.currentThread().getName()))
         return (-1, -1)
 
-    read_side = port_instance.get_read_side()
+    try:
+        read_side = port_instance.get_read_side()
+    except Exception as e:
+        read_side = None
+        helper_logger.log_warning("Failed to execute the get_read_side API for port {} due to {} from update_read_side".format(logical_port_name,repr(e)))
 
     if read_side is None or read_side is port_instance.EEPROM_ERROR or read_side < 0:
         helper_logger.log_error(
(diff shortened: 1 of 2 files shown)


zjswhhh:
What is this fix for? Is it for the "toggles fail while probes work" issue? Can you please provide more details?

sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
```diff
@@ -895,7 +895,11 @@ def toggle_mux_tor_direction_and_update_read_side(state, logical_port_name, phys
         helper_logger.log_error("Error: Could not get port instance for read side for while processing a toggle Y cable port {} {}".format(physical_port, threading.currentThread().getName()))
         return (-1, -1)
 
-    read_side = port_instance.get_read_side()
+    try:
+        read_side = port_instance.get_read_side()
+    except Exception as e:
+        read_side = None
+        helper_logger.log_warning("Failed to execute the get_read_side API for port {} due to {} from update_read_side".format(logical_port_name,repr(e)))
```
zjswhhh, line 902: So there will be a `read` before the actual `toggle` command? Is this `read` different from normal mux probe?

vdahiya12: No, this is a case where if read_side was not known beforehand, driver will try to get read_side again, before toggling the mux.

vdahiya12: normally we wont require reading read_side again,
https://github.com/sonic-net/sonic-platform-daemons/blob/534f8399afae4c601086a3f818257c55badb8a32/sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py#L974
only in case it is not known we try to re read it gain

vdahiya12:
> What is this fix for? Is it for the "toggles fail while probes work" issue? Can you please provide more details?

This PR just adds some exception logic for API's that were not covered, but mostly it wont fix the 
"toggles fail while probes work"

zjswhhh (approved):

zjswhhh:
> only in case it is not known we try to re read it gain

Hi @vdahiya12  - One more question, will this re-read writes the output to DBs?

PR #326, opened 2022-12-28
Xcvrd should restart if any child thread crashes

mihirpat1:
#### Description

Make all task workers as child thread of main process. 
The main process should watch for any child thread crash and do suicide to restart itself. Also ensure the backtrace of crashing child thread is shown

#### Motivation and Context

Overall, the change set now adheres to a template for creating thread and handles exceptions followed by restarting xcvrd.
With this change set, 3 threads will be created:
1. CmisManagerTask (This was formerly a process)
2. DomInfoUpdateTask (This already exists as a thread)
3. SfpStateUpdateTask (This was formerly a process)

All the above threads have now been named accordingly.
Once an exception is raised, the run() method will generate a traceback and will set the main_thread_stop event.
The xcvrd daemon keeps waiting until main_thread_stop event is generated. Once this event is generated, we first check if any thread is dead. If so, we capture the exception, evaluate if other threads have an exception generated as well and then force the xcvrd process to kill itself.
In case all threads are alive, we allow xcvrd daemon to gracefully exit.

Note - In order to kill SfpStateUpdateTask thread gracefully, we are manually raising an exception since SfpStateUpdateTask  has a call to an API which could potentially sleep in the order of seconds and hence, could block the xcvrd daemon graceful shutdown process for a prolonged time.

Please find the summary below.
	1. Kill xcvrd
	2. Restart xcvrd
	3. Restart pmon
	4. Kill 1 thread and ensure all threads are killed and xcvrd is respawned
	5. Ensure backtrace is generated with a manual exception - CmisManagerTask (Please refer to the below note for details)
	6. Ensure backtrace is generated with a manual exception - DomInfoUpdateTask (Please refer to the below note for details)
	7. Ensure backtrace is generated with a manual exception - SfpStateUpdateTask (Please refer to the below note for details)
	8. Ensure on Mellanox, no side effect is seen since they do not spawn CMIS thread

For a detailed UT, please refer to the below enclosure.
[xcvrd_thread_support.txt](https://github.com/sonic-net/sonic-platform-daemons/files/10313858/xcvrd_thread_support.txt)

Signed-off-by: Mihir Patel <patelmi@microsoft.com>

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -2056,6 +2056,9 @@ class SfpStateUpdateTask(threading.Thread):
             self.exc = e
             self.main_thread_stop_event.set()
 
+    # SfpStateUpdateTask thread has a call to an API which could potentially sleep in the order of seconds and hence,
+    # could block the xcvrd daemon graceful shutdown process for a prolonged time. Raising an exception will allow us to
+    # interrupt the SfpStateUpdateTask thread while sleeping and will allow graceful shutdown of the thread
     def raise_exception(self):
         res = ctypes.pythonapi.PyThreadState_SetAsyncExc(ctypes.c_ulong(self.thread_id),
               ctypes.py_object(SystemExit))
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -9,7 +9,6 @@ try:
     import ast
     import copy
     import json
-    import multiprocessing
     import os
     import signal
     import sys
@@ -19,6 +18,8 @@ try:
     import subprocess
     import argparse
     import re
+    import traceback
+    import ctypes
 
     from sonic_py_common import daemon_base, device_info, logger
     from sonic_py_common import multi_asic
@@ -938,7 +939,7 @@ def is_fast_reboot_enabled():
 
 # Thread wrapper class for CMIS transceiver management
 
-class CmisManagerTask:
+class CmisManagerTask(threading.Thread):
 
     CMIS_MAX_RETRIES     = 3
     CMIS_DEF_EXPIRED     = 60 # seconds, default expiration time
@@ -956,9 +957,12 @@ class CmisManagerTask:
     CMIS_STATE_REMOVED   = 'REMOVED'
     CMIS_STATE_FAILED    = 'FAILED'
 
-    def __init__(self, namespaces, port_mapping, skip_cmis_mgr=False):
-        self.task_stopping_event = multiprocessing.Event()
-        self.task_process = None
+    def __init__(self, namespaces, port_mapping, main_thread_stop_event, skip_cmis_mgr=False):
+        threading.Thread.__init__(self)
+        self.name = "CmisManagerTask"
+        self.exc = None
+        self.task_stopping_event = threading.Event()
+        self.main_thread_stop_event = main_thread_stop_event
         self.port_dict = {}
         self.port_mapping = copy.deepcopy(port_mapping)
         self.xcvr_table_helper = XcvrTableHelper(namespaces)
@@ -1605,7 +1609,7 @@ class CmisManagerTask:
 
         self.log_notice("Stopped")
 
-    def task_run(self):
+    def run(self):
         if platform_chassis is None:
             self.log_notice("Platform chassis is not available, stopping...")
             return
@@ -1614,23 +1618,34 @@ class CmisManagerTask:
             self.log_notice("Skipping CMIS Task Manager")
             return
 
-        self.task_process = multiprocessing.Process(target=self.task_worker)
-        if self.task_process is not None:
-            self.task_process.start()
+        try:
+            self.task_worker()
+        except Exception as e:
+            helper_logger.log_error("Exception occured at {} thread due to {}".format(threading.current_thread().getName(), repr(e)))
+            exc_type, exc_value, exc_traceback = sys.exc_info()
+            msg = traceback.format_exception(exc_type, exc_value, exc_traceback)
+            for tb_line in msg:
+                for tb_line_split in tb_line.splitlines():
+                    helper_logger.log_error(tb_line_split)
+            self.exc = e
+            self.main_thread_stop_event.set()
 
-    def task_stop(self):
+    def join(self):
         self.task_stopping_event.set()
-        if self.task_process is not None:
-            self.task_process.join()
-            self.task_process = None
+        threading.Thread.join(self)
+        if self.exc:
+            raise self.exc
 
 # Thread wrapper class to update dom info periodically
 
 
-class DomInfoUpdateTask(object):
-    def __init__(self, namespaces, port_mapping):
-        self.task_thread = None
+class DomInfoUpdateTask(threading.Thread):
+    def __init__(self, namespaces, port_mapping, main_thread_stop_event):
+        threading.Thread.__init__(self)
+        self.name = "DomInfoUpdateTask"
+        self.exc = None
         self.task_stopping_event = threading.Event()
+        self.main_thread_stop_event = main_thread_stop_event
         self.port_mapping = copy.deepcopy(port_mapping)
         self.namespaces = namespaces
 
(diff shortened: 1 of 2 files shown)


prgeor:
@keboliu can you review

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2011,18 +2040,34 @@ def task_worker(self, stopping_event, sfp_error_event):
...
+            helper_logger.log_error("Exception occured at {} thread due to {}".format(threading.current_thread().getName(), repr(e)))
+            exc_type, exc_value, exc_traceback = sys.exc_info()
+            msg = traceback.format_exception(exc_type, exc_value, exc_traceback)
+            for tb_line in msg:
+                for tb_line_split in tb_line.splitlines():
+                    helper_logger.log_error(tb_line_split)
+            self.exc = e
+            self.main_thread_stop_event.set()
+
+    def raise_exception(self):
```
prgeor, line 2059: can you add comment here why this exception is raised here?

prgeor:
@mihirpat1 can you check the behavior if you kill the threads with `kill -9`

prgeor:
@jaganbal-a @shyam77git please review

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have added a comment now.

mihirpat1:
> @mihirpat1 can you check the behavior if you kill the threads with `kill -9`

It seems we won't be able to kill threads using `kill -9` since `kill` requires process ID as a parameter. I have attached UT for killing xcvrd process though.

keboliu:
What if "CmisManagerTask " is disabled with intention? On the Mellanox platforms "CmisManagerTask" is disabled by setting the "skip_xcvrd_cmis_mgr" to "true".

mihirpat1:
> What if "CmisManagerTask " is disabled with intention? On the Mellanox platforms "CmisManagerTask" is disabled by setting the "skip_xcvrd_cmis_mgr" to "true".

If "CmisManagerTask" is disabled with intention, we will not spawn thread for it. Also, I have attached the UT for it (`#8` in [xcvrd_thread_support.txt](https://github.com/sonic-net/sonic-platform-daemons/files/10313858/xcvrd_thread_support.txt))

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -956,9 +957,12 @@ class CmisManagerTask:
...
     CMIS_STATE_FAILED    = 'FAILED'
 
-    def __init__(self, namespaces, port_mapping, skip_cmis_mgr=False):
-        self.task_stopping_event = multiprocessing.Event()
-        self.task_process = None
+    def __init__(self, namespaces, port_mapping, main_thread_stop_event, skip_cmis_mgr=False):
+        threading.Thread.__init__(self)
+        self.name = "CmisManagerTask"
+        self.exc = None
+        self.task_stopping_event = threading.Event()
```
jaganbal-a, line 964: Can you please explain how creating a event inside this thread will be used?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2199,9 +2247,10 @@ class DaemonXcvrd(daemon_base.DaemonBase):
     def __init__(self, log_identifier, skip_cmis_mgr=False):
         super(DaemonXcvrd, self).__init__(log_identifier)
         self.stop_event = threading.Event()
-        self.sfp_error_event = multiprocessing.Event()
+        self.sfp_error_event = threading.Event()
```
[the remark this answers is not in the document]
mihirpat1: It seems we are using it at the below:
https://github.com/sonic-net/sonic-platform-daemons/blob/master/sonic-xcvrd/xcvrd/xcvrd.py#L2391-L2392
        if self.sfp_error_event.is_set():
            sys.exit(SFP_SYSTEM_ERROR)

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: This event will be used to stop the handling with task_worker routine (in case of exception handling or graceful shutdown of the thread). The task_worker routine will be running in a while loop waiting for task_stopping_event to be set.
Hence, once we want the thread to exit, we must stop the execution of task_worker routine so that it comes out of the while loop.

sonic-xcvrd/xcvrd/xcvrd.py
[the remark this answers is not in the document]
jaganbal-a: But that code will never get hit, when the event get set due to error.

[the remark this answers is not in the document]
mihirpat1: Agreed, since we are killing the parent process (xcvrd) if next_state is STATE_EXIT, the above code will never get hit. This seems like a day 1 behavior though.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2351,36 +2400,54 @@ def run(self):
...
-        dom_info_update.task_stop()
-
-        # Stop the sfp state info update process
-        sfp_state_update.task_stop()
+        if dom_info_update.is_alive():
+            dom_info_update.join()
 
+        # Stop the sfp state info update thread
+        if sfp_state_update.is_alive():
+            sfp_state_update.raise_exception()
```
jaganbal-a, line 2449: Will it be good to join(1) with timeout and raise exception if timeout expires and the thread is still alive?

jaganbal-a (approved):

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2011,18 +2040,37 @@ def task_worker(self, stopping_event, sfp_error_event):
...
+    # could block the xcvrd daemon graceful shutdown process for a prolonged time. Raising an exception will allow us to
+    # interrupt the SfpStateUpdateTask thread while sleeping and will allow graceful shutdown of the thread
+    def raise_exception(self):
+        res = ctypes.pythonapi.PyThreadState_SetAsyncExc(ctypes.c_ulong(self.thread_id),
+              ctypes.py_object(SystemExit))
+        if res > 1:
+            ctypes.pythonapi.PyThreadState_SetAsyncExc(ctypes.c_ulong(self.thread_id), 0)
+            helper_logger.log_error('Exception raise failure for SfpStateUpdateTask')
+
+    def join(self):
```
keboliu, line 2069: it seems these 3 newly added "join" functions are all the same, suggest finding someway to reduce duplicated code.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2351,36 +2400,54 @@ def run(self):
...
-
-        # Start the Y-cable state info update process if Y cable presence established
+        # Start the sfp state info update thread
+        sfp_state_update = SfpStateUpdateTask(self.namespaces, port_mapping_data, retry_eeprom_set, self.stop_event, self.sfp_error_event)
+        sfp_state_update.start()
+        self.threads.append(sfp_state_update)
 
         # Start main loop
-        self.log_info("Start daemon main loop")
+        self.log_info("Start daemon main loop with thread count {}".format(len(self.threads)))
```
keboliu, line 2418: Is it possible to print out all the threads' names in order to give more information?

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I feel it's better to raise exception right away since based on my current analysis on Arista testbed, it takes about ~5s for the vendor api to come out of sleep. This timeout I assume could be dynamic based on the transceivers present on the device so I think having 1s delay will not give us a much advantage for join function with 1s timeout.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2351,36 +2400,54 @@ def run(self):
...
 
         self.stop_event.wait()
 
         self.log_info("Stop daemon main loop")
 
+        generate_sigkill = False
+        # check all threads are alive
+        for thread in self.threads:
```
vdahiya12, line 2426: will the main thread continue monitor the child threads forever ?
if not 
suggest using stop_event.wait() to wait for the exception to be set.

vdahiya12 (dismissed):

sonic-xcvrd/xcvrd/xcvrd.py
vdahiya12: ok, I din see stop_event.wait() is put just before.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have added log to display this information

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: Based on further changes, I have modified join function of CmisManagerTask to ensure that threading.Thread.join() is not called if skip_cmis_mgr is set. This is required to ensure that on Mellanox wherein CMIS thread is not spawned, xcvrd doesn't end up calling CmisManagerTask.join() and eventually results in a infinite wait and hence, prompting supervisord to kill xcvrd forcefully during graceful shutdown.
This leaves 2 other places with duplicated code for "join" function which I have planned to leave it currently as is to allow future modifications if required.

keboliu (approved):

prgeor (approved):

prgeor:
@StormLiangMS please pick to 202211

PR #329, opened 2023-01-04
Fix bug where transceiver info is missing after port breakout change

talber-nvidia:
#### Description
Added a call to the `self.on_add_logical_port()` method when handling `PORT_ADD` event in `class SfpStateUpdateTask`.
This call was removed in commit 94fa239a, which caused transceiver info to be missing after port breakout change.

#### Motivation and Context
Fix bug where transceiver info is missing after port breakout change.

Apply port breakout and ensure transceiver info exists for each sub-port.


liat-grozovik (approved):

prgeor:
@liat-grozovik please wait for this to be tested on Dell Z9332F

Junchao-Mellanox (approved):

PR #338, opened 2023-02-07
[ycabled] add more coverage to ycabled; add minor name change for vendor API CLI return key-values pairs

vdahiya12:
The main motivation of this PR is improve coverage of ycabled.
While fixing unit  test cases found an improvement to the name key-values returned to CLI, changed those names for keys as well

Unit-Tests and running the changes on testbed.

### DIFF
--- a/sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
+++ b/sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
@@ -2739,7 +2739,6 @@ def handle_config_mux_switchmode_arg_tbl_notification(fvp, xcvrd_config_hwmode_s
             config_mode = str(fvp_dict["config"])
 
             status = 'False'
-            result = None
             physical_port = get_ycable_physical_port_from_logical_port(port)
             if physical_port is None or physical_port == PHYSICAL_PORT_MAPPING_ERROR:
                 # error scenario update table accordingly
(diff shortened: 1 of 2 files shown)


sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
```diff
@@ -2665,7 +2666,7 @@ def handle_show_ber_cmd_arg_tbl_notification(fvp, xcvrd_show_ber_cmd_arg_tbl, xc
                     helper_logger.log_warning("Failed to execute cli cmd get_alive_status API for port {} due to {}".format(physical_port,repr(e)))
             if res is not None:
                 fvs_log = swsscommon.FieldValuePairs(
-                    [("cable", str(res))])
```
prgeor, line 2668: without this fix what was the impact?

prgeor (approved):

sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
vdahiya12: This will just result in executing the show CLI
$show mux cablealivestatus as
```PORT       ATTR    HEALTH
---------  ------  --------
 Ethernet0  cable True```

I think preferred way would be 
```PORT       ATTR    HEALTH
---------  ------  --------
Ethernet0 cable_alive True```

PR #340, opened 2023-02-21
Fixed chassis: test_container_checker failure for swss/syncd/teamd on multi-asic Supervisor - issue #13559

prem-nokia:
#### Description
Code changes are implemented in deinit function such that it does not remove entries in CHASSI_FABRIC_ASIC_TABLE when supervisorctl stop chassisd is triggered.

#### Motivation and Context
fixes https://github.com/sonic-net/sonic-buildimage/issues/13559: The container_cheaker depends on CHASSIS_FABRIC_ASIC_TABLE to provide the available asic list to calculate the expected dockers but as pmon is down and the table is cleared from chassis db, asic information is not present. Due to this, container_checker does not include the swss, syncd, teamd services in the expected docker list and fails to report the error.

1) Execute the below cmd.
```
admin@supervisor:~$docker exec pmon supervisorctl stop chassisd
chassisd: stopped
```
2) Look for the table entries using below cmd.
```
admin@supervisor:~$ sonic-db-dump -n CHASSIS_STATE_DB -y -k "CHASSIS*"
{
  "CHASSIS_FABRIC_ASIC_TABLE|asic6": {
    "expireat": 1677006173.948722,
    "ttl": -0.001,
    "type": "hash",
    "value": {
      "asic_id_in_module": "0",
      "asic_pci_address": "nokia-bdb:4:0",
      "name": "FABRIC-CARD3"
    }
  },
  "CHASSIS_FABRIC_ASIC_TABLE|asic7": {
    "expireat": 1677006173.9487307,
    "ttl": -0.001,
    "type": "hash",
    "value": {
      "asic_id_in_module": "1",
      "asic_pci_address": "nokia-bdb:4:1",
      "name": "FABRIC-CARD3"
    }
  }
}
```

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -208,9 +208,10 @@ class ModuleUpdater(logger.Logger):
             self.chassis_table._del(CHASSIS_INFO_KEY_TEMPLATE.format(1))
 
         if self.asic_table is not None:
-            asics = list(self.asic_table.getKeys())
-            for asic in asics:
-                self.asic_table._del(asic)
+            if not self._is_supervisor():
+                asics = list(self.asic_table.getKeys())
+                for asic in asics:
+                    self.asic_table._del(asic)
 
     def modules_num_update(self):
         # Check if module list is populated
--- a/sonic-chassisd/tests/test_chassisd.py
+++ b/sonic-chassisd/tests/test_chassisd.py
@@ -449,10 +449,8 @@ def test_asic_presence():
     midplane_table = module_updater.midplane_table
     fvs = midplane_table.get(name)
     assert fvs == None
-    fvs = fabric_asic_table.get("asic4")
-    assert fvs == None
-    fvs = fabric_asic_table.get("asic5")
-    assert fvs == None
+    verify_fabric_asic("asic4", "0000:04:00.0", name, "0")
+    verify_fabric_asic("asic5", "0000:05:00.0", name, "1")
 
 def test_signal_handler():
     exit_code = 0


sonic-chassisd/scripts/chassisd
```diff
@@ -208,9 +208,10 @@ class ModuleUpdater(logger.Logger):
             self.chassis_table._del(CHASSIS_INFO_KEY_TEMPLATE.format(1))
 
         if self.asic_table is not None:
-            asics = list(self.asic_table.getKeys())
-            for asic in asics:
-                self.asic_table._del(asic)
+            if not self._is_supervisor():
```
prgeor, line 211: can we debug why PMON is down at the first place? Can we ensure PMON is running? Can we modify the test to ensure PMON is running?

prem-nokia: > can we debug why PMON is down at the first place? -> this is done by the test case.
>Can we ensure PMON is running? -> that will be a change in the test case.
>Can we modify the test to ensure PMON is running? -> this will change the behavior of the test case. 

This PR is intended to keep the Fabric entries in STATE_DB when PMON is stopped in supervisor by the test case so syncd and swss will remain up state. Regardless of changes required in test case, this code change is required to make sure that the syncd and swss are up and running in case pmon is crashed/shutdown.

abdosi:
instead of note deleting it should we repopulate every time during init ?

rlhui:
@anamehra - would you like to review this?

rlhui:
Can we please have a more descriptive PR title?  thanks.

sonic-chassisd/scripts/chassisd
anamehra: @prgeor , bringing the pmon down was part of the test case. Test case failed because container checker could not find any asic entry to report for corresponding syncd/swss docker down.

@prem-nokia , Thanks for the fix! During the init, are these entries recreated from fresh in case any asic is removed due to physical FC removal?

anamehra:
> @anamehra - would you like to review this?

I will review, @rlhui . Thanks for tagging me!

sonic-chassisd/scripts/chassisd
mlok-nokia: > 
@anamehra  The Entry will be recreated when PMON is restarted.  We have tested it and works as expected

rlhui:
@prgeor , @anamehra - please approve if it looks good to you, thanks

anamehra (approved):

anamehra:
> Can we please have a more descriptive PR title? thanks.

Please resolve this comment as well.

sonic-chassisd/scripts/chassisd
abdosi: @mlok-nokia can you point to code where this happens? If i understand correctly issue was once we do de-init we were losing the entry and not getting repopulated.

abdosi (changes requested):
Please address the comment

sonic-chassisd/scripts/chassisd
mlok-nokia: Before this PR, these "FABRIC_ASIC" entries  will be removed if stop chassid or PMON (which trigger the de_init).   The entries will be recreated when chassisd or PMON is restarted

anamehra: Hi @abdosi , in case of a failure scenario, the pmon docker stays down (as per testcase) and that is why init was not called again. So the plan was to not clear the entries in de-init path so that the container_checker could still access the FC asic entries even if the chassisd is down.

prem-nokia:
> 



> Can we please have a more descriptive PR title? thanks.

Title changed as suggested. Thx

rlhui:
> Please address the comment

@prem-nokia reminder on this

mlok-nokia:
> Please address the comment
> @mlok-nokia can you point to code where this happens? If i understand correctly issue was once we do de-init we were losing the entry and not getting repopulated

@rlhui @abdosi  The reply has been made last week.  Somehow it has been moved to the section before this comment. 
 --  Before this PR, these "FABRIC_ASIC" entries will be removed if stop chassisd or PMON (which trigger the de_init). The entries will be recreated when chassisd or PMON is restarted. Function module_updater.module_db_update() will recreate them. We have manaully test it. But, after this PR, these fabric entries will not be removed when PMON/chassis is stopped.

prem-nokia:
> > Please address the comment
> 
> @prem-nokia reminder on this

This is already addressed by @mlok-nokia last week. Thanks

prgeor:
@prem-nokia Instead of making this workaround the problem, can we make the test more intelligent? It doesn't look correct to NOT delete the DB entries owned/created by PMON to be NOT deleted once PMON is dead. In the test we can get the list of asics before PMON  is stopped and use that list after PMON is stopped.?

mlok-nokia:
> @prem-nokia Instead of making this workaround the problem, can we make the test more intelligent? It doesn't look correct to NOT delete the DB entries owned/created by PMON to be NOT deleted once PMON is dead. In the test we can get the list of asics before PMON is stopped and use that list after PMON is stopped.?

@arlakshm @prgeor @anamehra @rlhui  Please help to decide what we should do here.  Based on the feature, these entries should not be removed since they are used to start the swss/syncd on the Supervisor card.

anamehra:
> @prem-nokia Instead of making this workaround the problem, can we make the test more intelligent? It doesn't look correct to NOT delete the DB entries owned/created by PMON to be NOT deleted once PMON is dead. In the test we can get the list of asics before PMON is stopped and use that list after PMON is stopped.?

@prgeor , the issue is not limited to the test case. It's more of a design issue. How do we want to handle a situation where pmon goes down due to some error and later syncd/swss restarts?

prgeor (approved):

prgeor:
> > @prem-nokia Instead of making this workaround the problem, can we make the test more intelligent? It doesn't look correct to NOT delete the DB entries owned/created by PMON to be NOT deleted once PMON is dead. In the test we can get the list of asics before PMON is stopped and use that list after PMON is stopped.?
> 
> @prgeor , the issue is not limited to the test case. It's more of a design issue. How do we want to handle a situation where pmon goes down due to some error and later syncd/swss restarts?

@prem-nokia @anamehra PMON going down is highly unlikely event which could happen with swss/syncd docker as well.  I think the issue is mainly to do with the design where PMON is now critical docker

anamehra:
> > > @prem-nokia Instead of making this workaround the problem, can we make the test more intelligent? It doesn't look correct to NOT delete the DB entries owned/created by PMON to be NOT deleted once PMON is dead. In the test we can get the list of asics before PMON is stopped and use that list after PMON is stopped.?
> > 
> > 
> > @prgeor , the issue is not limited to the test case. It's more of a design issue. How do we want to handle a situation where pmon goes down due to some error and later syncd/swss restarts?
> 
> @prem-nokia @anamehra PMON going down is highly unlikely event which could happen with swss/syncd docker as well. I think the issue is mainly to do with the design where PMON is now critical docker

Yes, @prgeor , I agree with that. As we discussed in the meeting, this PR is good to have fix to the time being to handle double failure scenarios. We should track the design improvements to handle the possible PMON failure as a critical issue.

PR #341, opened 2023-03-04
[ycable] add changes for correcting telemetry values for 'active-active'

vdahiya12:
This change creates `unknown, N/A` values for `active-standby` cable types for `active-active` type cable.
Since streaming telemetry today looks for both `active-active` and `active-standby` for posting values to State DB, this change introduces a  one time posting of fields with 'pseudo-cable' type when channels are attempted to be created by ycabled.

UT and posting the changes on test device.

### DIFF
--- a/sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
+++ b/sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
@@ -734,6 +734,7 @@ def setup_grpc_channels(stop_event):
     loopback_keys = {}
     hw_mux_cable_tbl = {}
     hw_mux_cable_tbl_peer = {}
+    mux_tbl = {}
 
     namespaces = multi_asic.get_front_end_namespaces()
     for namespace in namespaces:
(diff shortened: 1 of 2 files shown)


zjswhhh (commented):
lgtm!

@prgeor - can you also take a look please? We need this fix in next 202205 image. Thanks!!

zjswhhh (approved):

sonic-ycabled/ycable/ycable_utilities/y_cable_helper.py
```diff
@@ -2232,7 +2236,7 @@ def post_port_mux_info_to_db(logical_port_name, mux_tbl, asic_index, y_cable_tbl
 
     for physical_port in physical_port_list:
 
-        if not y_cable_wrapper_get_presence(physical_port):
+        if not y_cable_wrapper_get_presence(physical_port) or cable_type == 'pseudo-cable':
```
prgeor, line 2239: what is 'pseudo-cable'?

vdahiya12: this is a one-time call to make sure when the gRPC channels/stubs etc are created, ycabled posts the active-standby fields but only as `unknown/N/a`. The reason for doing it is ST today depends on having this fields populated albeit N/A or unknown, to post the full data to Kusto DB. 
this `pseudo-cable` logic takes care of it, and is only called once when Cable is either inserted/or ycabled spawns

PR #342, opened 2023-03-07
Retrieve subport from CONFIG_DB to enable breakout support

mihirpat1:
The changes with this PR needs to be merged along with sonic-net/sonic-platform-common#352

#### Description

Currently, xcvrd has a fixed logic to calculate active host lanes for a port. This needs to be changed to support ports with breakout cable.

#### Motivation and Context

The fixed host lane mapping will now be replaced by reading the subport from the CONFIG_DB. The subport will then be used to calculate active host lanes by reading the Application Descriptors.

Finding the desired application
The HostInterfaceID and HostLaneCount from Application Descriptors will be used to find the desired application for the port with breakout cable by comparing these with speed and lanes retrieved from CONFIG_DB.

Finding active host lane
Once the desired application is found, the HostLaneAssignmentOptions (bitmap of active host lanes) for the corresponding application and the subport from CONFIG_DB will be used to find the active host lanes.

In addition to the above, check_datapath_init_pending is now being called after ConfigStatusLane register has the value "ConfigSuccess". This has been done to be inline with the CMIS spec.

**Note**
Once the port_config.ini file has a new column for subport, the CONFIG_DB will be updated with the corresponding subport values (requires minigraph reload though).
The field "subport" will have the below value:
0 - This means the xcvr in non-breakout type
non-zero - The value here corresponds to the subport group (formed by a group of active lanes) to which the port belongs to

Testcase summary
The value of host_lanes can be seen in the logs by finding lanemask.
400G related TCs
1. SFP EEPROM dump
2. show int status
3. LLDP o/p
4. Log dump for CMIS init

200G breakout related TCs - 200G side
1. SFP EEPROM dump
2. show int status
3. LLDP o/p
4. Log dump for CMIS init

200G breakout related TCs - 100G side
1. SFP EEPROM dump
2. show int status
3. LLDP o/p
4. Log dump for CMIS init

5. Shut/no shut on subports (200G and 100G sides)
6. config reload (of Fixed chassis) with both native and break-out optics inserted
7. reboot (of Fixed chassis) with both native and break-out optics inserted
8. Remove and re-insert 2x100G break-out optic at run-time (system running)
9. xcvrd restart - Ensure link flap is not seen
10. PMON restart - Ensure link flap is not seen
11. SWSS restart
12. SYNCD restart
13. CONFIG_DB for a port after adding subport to port_config.ini

For additional details, please refer to [Unit-test_breakout_channel_v5.txt](https://github.com/sonic-net/sonic-platform-daemons/files/11081746/Unit-test_breakout_channel_v5.txt)

The code changes have been tested on a chassis with 200G AOC breakout cable.
Also, the code changes have been tested on a chassis with 400ZR transceiver and 10GBase-SR

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1097,13 +1097,15 @@ class CmisManagerTask(threading.Thread):
     def get_cmis_dp_deinit_duration_secs(self, api):
         return api.get_datapath_deinit_duration()/1000
 
-    def get_cmis_host_lane_mask(self, api, host_lane_count, channel, speed):
+    def get_cmis_host_lanes_mask(self, api, appl, host_lane_count, channel, speed):
         """
         Retrieves mask of active host lanes based on host lane count, channel and speed
 
         Args:
             api:
                 XcvrApi object
+            appl:
+                Integer, the transceiver-specific application code
             host_lane_count:
                 Integer, number of lanes on the host side
             channel:
@@ -1115,28 +1117,26 @@ class CmisManagerTask(threading.Thread):
             Integer, a mask of the active lanes on the host side
             e.g. 0x5 for lane 0 and lane 2.
         """
-        host_lane_mask = 0
+        host_lanes_mask = 0
 
-        if host_lane_count <= 0 or channel < 0 or speed <= 0:
-            self.log_error("Invalid input to get host lane mask - host_lane_count {} channel {} speed {}!".format(
-                            host_lane_count, channel, speed))
-            return host_lane_mask
+        if host_lane_count <= 0 or channel < 0 or speed <= 0 or appl < 1:
+            self.log_error("Invalid input to get host lane mask - host_lane_count {} "
+                            "channel {} speed {} appl {}!".format(
+                            host_lane_count, channel, speed, appl))
+            return host_lanes_mask
 
-        if channel != 0:
-            appl = self.get_cmis_application_desired(api, host_lane_count, speed)
-            if appl < 1:
-                self.log_error("Failed to get host lanes as no suitable app {} found host_lane_count {} speed {}".format(
-                                appl, host_lane_count, speed))
-                return host_lane_mask
-
-            host_lane_assignment_option = api.get_host_lane_assignment_option(appl)
-            host_lane_start_bit = (host_lane_count * (channel - 1))
-            if host_lane_assignment_option & (1 << host_lane_start_bit):
-                host_lane_mask = ((1 << host_lane_count) - 1) << host_lane_start_bit
+        host_lane_assignment_option = api.get_host_lane_assignment_option(appl)
+        host_lane_start_bit = (host_lane_count * (0 if channel == 0 else channel - 1))
+        if host_lane_assignment_option & (1 << host_lane_start_bit):
+            host_lanes_mask = ((1 << host_lane_count) - 1) << host_lane_start_bit
         else:
-            host_lane_mask = (1 << host_lane_count) - 1
+            self.log_error("Unable to find lowest host lane - host_lane_assignment_option {}"
+                            " host_lane_start_bit {} host_lane_count {} channel {} speed {}"
+                            " appl {}!".format(
+                            host_lane_assignment_option, host_lane_start_bit, host_lane_count,
+                            channel, speed, appl))
 
-        return host_lane_mask
+        return host_lanes_mask
 
     def is_cmis_application_update_required(self, api, host_lane_count, channel, speed):
         """
@@ -1464,10 +1464,17 @@ class CmisManagerTask(threading.Thread):
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_READY
                         continue
 
-                    host_lanes = self.get_cmis_host_lane_mask(api, host_lane_count, channel, speed)
-                    if host_lanes == 0:
-                        self.log_error("{}: Invalid lane mask received host_lane_count {} channel {} speed {}!".format(
-                                        lport, host_lane_count, channel, speed))
+                    appl = self.get_cmis_application_desired(api, host_lane_count, host_speed)
+                    if appl < 1:
+                        self.log_error("{}: no suitable app for the port".format(lport))
+                        self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_FAILED
+                        continue
+
+                    host_lanes_mask = self.get_cmis_host_lanes_mask(api, appl, host_lane_count, channel, speed)
+                    if host_lanes_mask == 0:
+                        self.log_error("{}: Invalid lane mask received - host_lane_count {} channel {} "
+                                        "speed {} appl {}!".format(
+                                        lport, host_lane_count, channel, speed, appl))
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_FAILED
                         continue
 
@@ -1489,8 +1496,8 @@ class CmisManagerTask(threading.Thread):
                 now = datetime.datetime.now()
                 expired = self.port_dict[lport].get('cmis_expired')
                 retries = self.port_dict[lport].get('cmis_retries', 0)
-                self.log_notice("{}: {}G, lanemask=0x{:x}, state={}, retries={}".format(
-                                lport, int(speed/1000), host_lanes, state, retries))
+                self.log_notice("{}: {}G, lanemask=0x{:x}, state={}, appl {} retries={}".format(
+                                lport, int(speed/1000), host_lanes_mask, state, appl, retries))
                 if retries > self.CMIS_MAX_RETRIES:
                     self.log_error("{}: FAILED".format(lport))
                     self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_FAILED
@@ -1503,7 +1510,7 @@ class CmisManagerTask(threading.Thread):
                                 self.port_dict[lport]['admin_status'] != 'up':
                            self.log_notice("{} Forcing Tx laser OFF".format(lport))
                            # Force DataPath re-init
-                           api.tx_disable_channel(host_lanes, True)
+                           api.tx_disable_channel(host_lanes_mask, True)
                            self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_READY
                            continue
                     # Configure the target output power if ZR module
@@ -1516,13 +1523,7 @@ class CmisManagerTask(threading.Thread):
                               else:
                                  self.log_notice("{} Successfully configured Tx power = {}".format(lport, tx_power))
 
-                        appl = self.get_cmis_application_desired(api, host_lane_count, host_speed)
-                        if appl < 1:
-                            self.log_error("{}: no suitable app for the port".format(lport))
-                            self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_FAILED
-                            continue
-
-                        need_update = self.is_cmis_application_update_required(api, host_lane_count, host_lanes, host_speed)
+                        need_update = self.is_cmis_application_update_required(api, host_lane_count, host_lanes_mask, host_speed)
 
                         # For ZR module, Datapath needes to be re-initlialized on new channel selection
                         if api.is_coherent_module():
@@ -1541,11 +1542,11 @@ class CmisManagerTask(threading.Thread):
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_DP_DEINIT
                     elif state == self.CMIS_STATE_DP_DEINIT:
                         # D.2.2 Software Deinitialization
-                        api.set_datapath_deinit(host_lanes)
+                        api.set_datapath_deinit(host_lanes_mask)
 
                         # D.1.3 Software Configuration and Initialization
-                        if not api.tx_disable_channel(host_lanes, True):
-                            self.log_notice("{}: unable to turn off tx power".format(lport))
+                        if not api.tx_disable_channel(host_lanes_mask, True):
+                            self.log_notice("{}: unable to turn off tx power with host_lanes_mask {}".format(lport, host_lanes_mask))
                             self.port_dict[lport]['cmis_retries'] = retries + 1
                             continue
 
@@ -1563,7 +1564,7 @@ class CmisManagerTask(threading.Thread):
                                 self.force_cmis_reinit(lport, retries + 1)
                             continue
 
-                        if not self.check_datapath_state(api, host_lanes, ['DataPathDeactivated']):
+                        if not self.check_datapath_state(api, host_lanes_mask, ['DataPathDeactivated']):
                             if (expired is not None) and (expired <= now):
                                 self.log_notice("{}: timeout for 'DataPathDeactivated state'".format(lport))
                                 self.force_cmis_reinit(lport, retries + 1)
@@ -1579,29 +1580,23 @@ class CmisManagerTask(threading.Thread):
                                    self.log_notice("{} configured laser frequency {} GHz".format(lport, freq))
 
                         # D.1.3 Software Configuration and Initialization
-                        appl = self.get_cmis_application_desired(api, host_lane_count, host_speed)
-                        if appl < 1:
-                            self.log_error("{}: no suitable app for the port".format(lport))
-                            self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_FAILED
-                            continue
-
-                        if not api.set_application(host_lanes, appl):
+                        if not api.set_application(host_lanes_mask, appl):
                             self.log_notice("{}: unable to set application".format(lport))
                             self.force_cmis_reinit(lport, retries + 1)
                             continue
 
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_DP_INIT
                     elif state == self.CMIS_STATE_DP_INIT:
-                        if not self.check_config_error(api, host_lanes, ['ConfigSuccess']):
+                        if not self.check_config_error(api, host_lanes_mask, ['ConfigSuccess']):
                             if (expired is not None) and (expired <= now):
                                 self.log_notice("{}: timeout for 'ConfigSuccess'".format(lport))
                                 self.force_cmis_reinit(lport, retries + 1)
                             continue
-
-                        if getattr(api, 'get_cmis_rev', None):
+                        
+                        if hasattr(api, 'get_cmis_rev'):
                             # Check datapath init pending on module that supports CMIS 5.x
                             majorRev = int(api.get_cmis_rev().split('.')[0])
-                            if majorRev >= 5 and not self.check_datapath_init_pending(api, host_lanes):
+                            if majorRev >= 5 and not self.check_datapath_init_pending(api, host_lanes_mask):
                                 self.log_notice("{}: datapath init not pending".format(lport))
                                 self.force_cmis_reinit(lport, retries + 1)
                                 continue
@@ -1616,24 +1611,24 @@ class CmisManagerTask(threading.Thread):
                             continue
 
                         # D.1.3 Software Configuration and Initialization
-                        api.set_datapath_init(host_lanes)
+                        api.set_datapath_init(host_lanes_mask)
                         dpInitDuration = self.get_cmis_dp_init_duration_secs(api)
                         self.log_notice("{}: DpInit duration {} secs".format(lport, dpInitDuration))
                         self.port_dict[lport]['cmis_expired'] = now + datetime.timedelta(seconds=dpInitDuration)
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_DP_TXON
                     elif state == self.CMIS_STATE_DP_TXON:
-                        if not self.check_datapath_state(api, host_lanes, ['DataPathInitialized']):
+                        if not self.check_datapath_state(api, host_lanes_mask, ['DataPathInitialized']):
                             if (expired is not None) and (expired <= now):
                                 self.log_notice("{}: timeout for 'DataPathInitialized'".format(lport))
                                 self.force_cmis_reinit(lport, retries + 1)
                             continue
 
                         # Turn ON the laser
-                        api.tx_disable_channel(host_lanes, False)
+                        api.tx_disable_channel(host_lanes_mask, False)
                         self.log_notice("{}: Turning ON tx power".format(lport))
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_DP_ACTIVATE
                     elif state == self.CMIS_STATE_DP_ACTIVATE:
-                        if not self.check_datapath_state(api, host_lanes, ['DataPathActivated']):
+                        if not self.check_datapath_state(api, host_lanes_mask, ['DataPathActivated']):
                             if (expired is not None) and (expired <= now):
                                 self.log_notice("{}: timeout for 'DataPathActivated'".format(lport))
                                 self.force_cmis_reinit(lport, retries + 1)
@@ -1642,8 +1637,8 @@ class CmisManagerTask(threading.Thread):
                         self.log_notice("{}: READY".format(lport))
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_READY
 
-                except (NotImplementedError, AttributeError):
-                    self.log_error("{}: internal errors".format(lport))
+                except (NotImplementedError, AttributeError) as e:
+                    self.log_error("{}: internal errors due to {}".format(lport, e))
                     self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_FAILED
 
         self.log_notice("Stopped")
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1023,8 +1023,8 @@ class CmisManagerTask(threading.Thread):
                 self.port_dict[lport]['laser_freq'] = int(port_change_event.port_dict['laser_freq'])
             if 'tx_power' in port_change_event.port_dict:
                 self.port_dict[lport]['tx_power'] = float(port_change_event.port_dict['tx_power'])
-            if 'channel' in port_change_event.port_dict:
-                self.port_dict[lport]['channel'] = int(port_change_event.port_dict['channel'])
+            if 'subport' in port_change_event.port_dict:
+                self.port_dict[lport]['subport'] = int(port_change_event.port_dict['subport'])
 
             self.force_cmis_reinit(lport, 0)
         else:
@@ -1097,9 +1097,9 @@ class CmisManagerTask(threading.Thread):
     def get_cmis_dp_deinit_duration_secs(self, api):
         return api.get_datapath_deinit_duration()/1000
 
-    def get_cmis_host_lanes_mask(self, api, appl, host_lane_count, channel):
+    def get_cmis_host_lanes_mask(self, api, appl, host_lane_count, subport):
         """
-        Retrieves mask of active host lanes based on appl, host lane count and channel
+        Retrieves mask of active host lanes based on appl, host lane count and subport
 
         Args:
             api:
@@ -1108,8 +1108,8 @@ class CmisManagerTask(threading.Thread):
                 Integer, the transceiver-specific application code
             host_lane_count:
                 Integer, number of lanes on the host side
-            channel:
-                Integer, channel id of the group which the host lanes belong to (1 based),
+            subport:
+                Integer, subport id of the group which the host lanes belong to (1 based),
                 0 means port is a non-breakout port
 
         Returns:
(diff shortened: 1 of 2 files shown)


mihirpat1:
@prgeor @jaganbal-a @keboliu @shyam77git - It will be great if you can help in reviewing this PR

prgeor:
@shyam77git @jaganbal-a please review

prgeor:
@keboliu please review

jaganbal-a:
how the 4x10 or 4x25 breakout is taken cared for 100G/40G module ?

mihirpat1:
> how the 4x10 or 4x25 breakout is taken cared for 100G/40G module ?

The API get_cmis_application_desired will return the application ID by matching the speed and number of active host lanes from CONFIG_DB. This application ID will then be used by get_cmis_host_lanes function to find the appropriate active host lanes.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1102,13 +1097,63 @@ def get_cmis_dp_init_duration_secs(self, api):
...
+            speed:
+                Integer, the port speed of the host interface
+
+        Returns:
+            Integer, a bitmask of the lanes on the host side
+            e.g. 0x5 for lane 0 and lane 2.
+        """
+        host_lanes = 0
+
+        if channel > host_lane_count:
```
prgeor, line 1120: wont this check fail for 8x100G ? 8th channel > 2 host_lane_count ?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1102,13 +1097,63 @@ def get_cmis_dp_init_duration_secs(self, api):
...
+                if host_lane_assignment_option & mask != 0:
+                    curr_channel_num += 1
+                    if curr_channel_num == channel:
+                        for i in range(host_lane_count):
+                            host_lanes |= mask
+                            mask = mask << 1
+                        break
+                mask = mask << 1
+        else:
+            for i in range(host_lane_count):
```
prgeor, line 1143: won't this be correct? `(1 << host_lane_count) - 1` since host_lane_count != 0

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1102,13 +1097,63 @@ def get_cmis_dp_init_duration_secs(self, api):
...
+
+        Returns:
+            Integer, a bitmask of the lanes on the host side
+            e.g. 0x5 for lane 0 and lane 2.
+        """
+        host_lanes = 0
+
+        if channel > host_lane_count:
+            self.log_error("Failed to get host lanes as channel > host_lane_count!")
+            return host_lanes
```
prgeor, line 1122: can we put check speed !=0 , host_lane_count !=0 and channel >=0

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1102,13 +1097,63 @@ def get_cmis_dp_init_duration_secs(self, api):
...
+            for bit in range(8):
+                mask = 1 << bit
+                if host_lane_assignment_option & mask != 0:
+                    curr_channel_num += 1
+                    if curr_channel_num == channel:
+                        for i in range(host_lane_count):
+                            host_lanes |= mask
+                            mask = mask << 1
+                        break
+                mask = mask << 1
```
prgeor, line 1141: this mask value is overwritten at the line 1133, so whats the purpose of updating here?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1102,13 +1097,63 @@ def get_cmis_dp_init_duration_secs(self, api):
...
+                return host_lanes
+
+            host_lane_assignment_option = api.get_host_lane_assignment_option(appl)
+            curr_channel_num = 0
+            for bit in range(8):
+                mask = 1 << bit
+                if host_lane_assignment_option & mask != 0:
+                    curr_channel_num += 1
+                    if curr_channel_num == channel:
+                        for i in range(host_lane_count):
```
prgeor, line 1137: what if the `host_lane_count` goes beyond the host_lane_assignment? i don't that case is taken care of?
consider this host_lane_count = 4 and host_lane_assignment = 01010101

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1102,13 +1097,63 @@ def get_cmis_dp_init_duration_secs(self, api):
...
+            e.g. 0x5 for lane 0 and lane 2.
+        """
+        host_lanes = 0
+
+        if channel > host_lane_count:
+            self.log_error("Failed to get host lanes as channel > host_lane_count!")
+            return host_lanes
+
+        if channel != 0:
+            appl = self.get_cmis_application_desired(api, host_lane_count, speed)
```
prgeor, line 1125: I dont see we are iterating through all advertised 'appl' to find out if the host lane assignment satisfies the requested lane mapping and speed

mihirpat1:
> hannel > host_lane_



> > I get_cmis_application_desired will return the application ID by matching the speed and number of active host lanes from CONFIG_DB. This application ID will then be used by get_cmis_host_lanes function to find the appropriate active host lanes.
> 
> get_cmis_host_lanes() is not supposed to be invoked for non-CMIS module isn't it ? the question is for breakout 4x25G and 4x10G on non-cmis (100G/40G modules)

Yes,  get_cmis_host_lanes() will not be invoked for non-CMIS module.
Since 4x25G and 4x10G are non-CMIS modules, datapath initialization from xcvrd is not required and hence, the current changes are not needed for such modules.
The expectation is that existing bring-up sequence outside of xcvrd should take care of initial reset for such modules and then, it should work as is without the intervention of xcvrd.

shyam77git:
> > hannel > host_lane_
> 
> > > I get_cmis_application_desired will return the application ID by matching the speed and number of active host lanes from CONFIG_DB. This application ID will then be used by get_cmis_host_lanes function to find the appropriate active host lanes.
> > 
> > 
> > get_cmis_host_lanes() is not supposed to be invoked for non-CMIS module isn't it ? the question is for breakout 4x25G and 4x10G on non-cmis (100G/40G modules)
> 
> Yes, get_cmis_host_lanes() will not be invoked for non-CMIS module. Since 4x25G and 4x10G are non-CMIS modules, datapath initialization from xcvrd is not required and hence, the current changes are not needed for such modules. The expectation is that existing bring-up sequence outside of xcvrd should take care of initial reset for such modules and then, it should work as is without the intervention of xcvrd.

@mihirpat1 , @prgeor 
For non-CMIS compliant modules (i.e. SFF* spec based ones), one cannot expect them to work seamlessly. 
Beside xcvrd, following areas to be checked from code-flow and validation standpoint:
- syncd vertical (syncd-> SAI-> SDK) w.r.t handling the port brea-out
- use case: shut, no-shut on a particular channel (broken-down/logical port) of an interface
- though xcvrd vertical seems to have no role (per above-mentioned explanation), still its part of end-to-end validation to ensure no open item/issue
I'd suggest to file a separate git issue for this

shyam77git:
@mihirpat1 
Besides the UT cases you mentioned in description section and attached UT logs, could you validate the following
1. config reload (of Fixed chassis) with both native and break-out optics inserted
2. reboot (of Fixed chassis) with both native and break-out optics inserted
3. Mode change: Change existing (inserted) optics to 4x100G or 2x100G breakout mode 
4. Mode change: Change/revert from breakout mode to native mode
5. Insert break-out optics at run-time (system running): 4x100G or 2x100G break-out optics

shyam77git:
Wanted to keep following comment separate from above UT comment:
How about port break-out validation on Modular chassis's LC?
IMO, should be covered here to ensure no issue

shyam77git:
@mihirpat1 
Don't see mention and/or validation of SONiC breakout CLI !
Could you configure and validate following SONiC breakout CLIs?

#config interface breakout <interface_name>
show interfaces breakout
show interfaces breakout current-mode
show interfaces breakout current-mode <interface_name>

mihirpat1:
> or non-CMIS compliant modules (i.e. SFF* spec based ones), one cannot expect them to work seamlessly. Beside xcvrd, following areas to be checked from code-flow and validation standpoint:
> 
> * syncd vertical (syncd-> SAI-> SDK) w.r.t handling the port brea-out
> * use case: shut, no-shut on a particular channel (broken-down/logical port) of an interface
> * though xcvrd vertical seems to have no role (per above-mentioned explanation), still its part of end-to-end validation to ensure no open item/issue

I have now filed sonic-net/sonic-buildimage/issues/14223 to track this

mihirpat1:
> @mihirpat1 Don't see mention and/or validation of SONiC breakout CLI ! Could you configure and validate following SONiC breakout CLIs?
> 
> #config interface breakout <interface_name> show interfaces breakout show interfaces breakout current-mode show interfaces breakout current-mode <interface_name>

We currently do not support dynamic port breakout config so the above testcases cannot be accomplished

mihirpat1:
> @mihirpat1 Besides the UT cases you mentioned in description section and attached UT logs, could you validate the following
> 
> 1. config reload (of Fixed chassis) with both native and break-out optics inserted
> 2. reboot (of Fixed chassis) with both native and break-out optics inserted
> 3. Mode change: Change existing (inserted) optics to 4x100G or 2x100G breakout mode
> 4. Mode change: Change/revert from breakout mode to native mode
> 5. Insert break-out optics at run-time (system running): 4x100G or 2x100G break-out optics

I have attached the UT for #1, #2 and #5 now.
#4 and #5 cannot be tested since we do not support dynamic port breakout support yet.

mihirpat1:
> Wanted to keep following comment separate from above UT comment: How about port break-out validation on Modular chassis's LC? IMO, should be covered here to ensure no issue

@shyam77git We currently cannot test the cable on our modular chassis. It will be great if Cisco can help to test this

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1432,6 +1471,13 @@ def task_worker(self):
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_READY
                         continue
 
+                    host_lanes = self.get_cmis_host_lane_mask(api, host_lane_count, channel, speed)
+                    if host_lanes == 0:
+                        self.log_error("{}: Invalid lane mask received host_lane_count {} channel {} speed {}!".format(
+                                        lport, host_lane_count, channel, speed))
+                        self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_FAILED
```
jaganbal-a, line 1478: CMIS_STATE_FAILED, I dont see this state is handled, so what is the expectation when SW set this state.
Do we need to bring the link down/move to DP-Deactivated if the state is already DP_Activation /DP_initialized?

mihirpat1:
> > > @mihirpat1 Besides the UT cases you mentioned in description section and attached UT logs, could you validate the following
> > > 
> > > 1. config reload (of Fixed chassis) with both native and break-out optics inserted
> > > 2. reboot (of Fixed chassis) with both native and break-out optics inserted
> > > 3. Mode change: Change existing (inserted) optics to 4x100G or 2x100G breakout mode
> > > 4. Mode change: Change/revert from breakout mode to native mode
> > > 5. Insert break-out optics at run-time (system running): 4x100G or 2x100G break-out optics
> > 
> > 
> > I have attached the UT for 1, 2 and 5 now. 4 and 5 cannot be tested since we do not support dynamic port breakout support yet.
> 
> Could you share the UT link/logs for 1, 2 and 5?

@shyam77git - I have attached the results to [Unit-test_breakout_channel_v3.txt](https://github.com/sonic-net/sonic-platform-daemons/files/10960056/Unit-test_breakout_channel_v3.txt). Please refer to 6, 7 and 8 in this file.

shyam77git:
> > > > @mihirpat1 Besides the UT cases you mentioned in description section and attached UT logs, could you validate the following
> > > > 
> > > > 1. config reload (of Fixed chassis) with both native and break-out optics inserted
> > > > 2. reboot (of Fixed chassis) with both native and break-out optics inserted
> > > > 3. Mode change: Change existing (inserted) optics to 4x100G or 2x100G breakout mode
> > > > 4. Mode change: Change/revert from breakout mode to native mode
> > > > 5. Insert break-out optics at run-time (system running): 4x100G or 2x100G break-out optics
> > > 
> > > 
> > > I have attached the UT for 1, 2 and 5 now. 4 and 5 cannot be tested since we do not support dynamic port breakout support yet.
> > 
> > 
> > Could you share the UT link/logs for 1, 2 and 5?
> 
> @shyam77git - I have attached the results to [Unit-test_breakout_channel_v3.txt](https://github.com/sonic-net/sonic-platform-daemons/files/10960056/Unit-test_breakout_channel_v3.txt). Please refer to 6, 7 and 8 in this file.

Thanks ! 6, 7 and 8 UT looks good.
BTW, we discussed to validate xcvrd restart. Could you do that (if not done yet)? Additionally, docker container restart (primarily PMON, SWSS, SYNCD) with these native (400G) and break-out (2x100G) optical modules, cables

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: @jaganbal-a - Once a port is in CMIS_STATE_FAILED, we don't want to pass any config to the module and hence, any future config will be ignored when a port is in this state.
I am marking the state to failed before starting the CMIS initialization so we need not take care of other CMIS init relevant steps. Also, in this case, since we don't have the host_lane count, it is not possible to take any further action.

mihirpat1:
> > > > > @mihirpat1 Besides the UT cases you mentioned in description section and attached UT logs, could you validate the following
> > > > > 
> > > > > 1. config reload (of Fixed chassis) with both native and break-out optics inserted
> > > > > 2. reboot (of Fixed chassis) with both native and break-out optics inserted
> > > > > 3. Mode change: Change existing (inserted) optics to 4x100G or 2x100G breakout mode
> > > > > 4. Mode change: Change/revert from breakout mode to native mode
> > > > > 5. Insert break-out optics at run-time (system running): 4x100G or 2x100G break-out optics
> > > > 
> > > > 
> > > > I have attached the UT for 1, 2 and 5 now. 4 and 5 cannot be tested since we do not support dynamic port breakout support yet.
> > > 
> > > 
> > > Could you share the UT link/logs for 1, 2 and 5?
> > 
> > 
> > @shyam77git - I have attached the results to [Unit-test_breakout_channel_v3.txt](https://github.com/sonic-net/sonic-platform-daemons/files/10960056/Unit-test_breakout_channel_v3.txt). Please refer to 6, 7 and 8 in this file.
> 
> Thanks ! 6, 7 and 8 UT looks good. BTW, we discussed to validate xcvrd restart. Could you do that (if not done yet)? Additionally, docker container restart (primarily PMON, SWSS, SYNCD) with these native (400G) and break-out (2x100G) optical modules, cables

@shyam77git - I have now added test cases 9 - 12 to address the same.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have filed #344 to track this.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1477,13 +1523,13 @@ def task_worker(self):
...
 
-                        appl = self.get_cmis_application_desired(api, host_lanes, host_speed)
+                        appl = self.get_cmis_application_desired(api, host_lane_count, host_speed)
                         if appl < 1:
                             self.log_error("{}: no suitable app for the port".format(lport))
                             self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_FAILED
                             continue
 
-                        need_update = self.is_cmis_application_update_required(api, host_lanes, host_speed)
+                        need_update = self.is_cmis_application_update_required(api, host_lane_count, host_lanes, host_speed)
```
shyam77git, line 1532: Please refer to the comments I put in Conversation/General section

shyam77git (dismissed):
LGTM

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have simplified the logic now.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1102,13 +1097,56 @@ def get_cmis_dp_init_duration_secs(self, api):
...
+                                appl, host_lane_count, speed))
+                return host_lane_mask
+
+            host_lane_assignment_option = api.get_host_lane_assignment_option(appl)
+            host_lane_start_bit = (host_lane_count * (channel - 1))
+            if host_lane_assignment_option & (1 << host_lane_start_bit):
+                host_lane_mask = ((1 << host_lane_count) - 1) << host_lane_start_bit
+        else:
+            host_lane_mask = (1 << host_lane_count) - 1
+
```
prgeor, line 1138: can we add an error message if host_lane_mask is ZERO so that we know something went wrong?

prgeor:
@mihirpat1 can we test if the link is UP and Xcvrd crashes or PMON restarts, the link does not flap?

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: @mihirpat1 can we return (host_lane_mask, appl) so that caller can print the selected application and host_lane_mask for error and info

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1102,13 +1097,56 @@ def get_cmis_dp_init_duration_secs(self, api):
...
+        if channel != 0:
+            appl = self.get_cmis_application_desired(api, host_lane_count, speed)
+            if appl < 1:
+                self.log_error("Failed to get host lanes as no suitable app {} found host_lane_count {} speed {}".format(
+                                appl, host_lane_count, speed))
+                return host_lane_mask
+
+            host_lane_assignment_option = api.get_host_lane_assignment_option(appl)
+            host_lane_start_bit = (host_lane_count * (channel - 1))
```
Junchao-Mellanox, line 1133: just wanna confirm that channel is 0 based o r 1 based? I guess (0 - 1) will generate a negative value, is it expected?

sonic-xcvrd/xcvrd/xcvrd.py
Junchao-Mellanox, line 1: Not sure that 'channel' is a required configuration here. I suppose we can deduce `channel` value according to `index` and `lanes`. E.g, Ethernet0 got lanes (0,1,2,3), Ethernet4 got lanes (4,5,6,7), we can deduce the Etherent0.channel=0 and Ethernet4.channel=1. Maybe I missed something, could you please elaborate why channel is a required configuration here?

Junchao-Mellanox (changes requested):

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: Channel is 0 based wherein 0 means the ports doesn't have a breakout module.
Non zero channel number will point to the logical group of lanes we would use.

In this case, (0 - 1) is not a valid test case since this piece of code gets executed if and only if channel != 0. Also, we have a check in this function to ensure channel is >= 0.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: Please note that xcvrd has the physical lane and not logical lane coming from the CONFIG_DB. 
Or if we wish to use index as a search key, we would need to go through the data of all ports to find the ports with matching indexes and then eventually, derive channel number from it.

With given index and physical lane, I am not sure how will we figure out the group/channel to which a port belongs to.
In case of 8 x 50G, we would have Ethernet8.. Ethernet15 using logical lanes 0..7 and physical lanes 2041..2048 respectively (for example).
Let's say I need to configure Ethernet12 which uses physical lane 2045 and index 1. 
With given physical lane and index, I am not sure how will we derive the logical lane mapping unless we explore all ports which have the same index?

Junchao-Mellanox: thanks

Junchao-Mellanox (approved):

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1567,6 +1598,14 @@ def task_worker(self):
                                 self.force_cmis_reinit(lport, retries + 1)
                             continue
 
+                        if getattr(api, 'get_cmis_rev', None):
```
Junchao-Mellanox, line 1601: hasattr?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1102,13 +1097,56 @@ def get_cmis_dp_init_duration_secs(self, api):
...
+        if channel != 0:
+            appl = self.get_cmis_application_desired(api, host_lane_count, speed)
+            if appl < 1:
+                self.log_error("Failed to get host lanes as no suitable app {} found host_lane_count {} speed {}".format(
+                                appl, host_lane_count, speed))
+                return host_lane_mask
+
+            host_lane_assignment_option = api.get_host_lane_assignment_option(appl)
+            host_lane_start_bit = (host_lane_count * (channel - 1))
+            if host_lane_assignment_option & (1 << host_lane_start_bit):
```
Junchao-Mellanox, line 1134: Can we add a log if this statement return False?

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have addressed this now.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I am now taking appl is an input parameter to this function as per our discussion.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have addressed this now.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: > @mihirpat1 can we test if the link is UP and Xcvrd crashes or PMON restarts, the link does not flap?

Yes, I have verified that link remains up on local as well as peer port when xcvrd crash/restart and PMON restart

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1432,6 +1464,20 @@ def task_worker(self):
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_READY
                         continue
 
+                    appl = self.get_cmis_application_desired(api, host_lane_count, host_speed)
+                    if appl < 1:
+                        self.log_error("{}: no suitable app for the port".format(lport))
```
prgeor, line 1469: @mihirpat1 can you log the appl, lane count, speed?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1102,13 +1097,56 @@ def get_cmis_dp_init_duration_secs(self, api):
...
+                            "channel {} speed {} appl {}!".format(
+                            host_lane_count, channel, speed, appl))
+            return host_lanes_mask
+
+        host_lane_assignment_option = api.get_host_lane_assignment_option(appl)
+        host_lane_start_bit = (host_lane_count * (0 if channel == 0 else channel - 1))
+        if host_lane_assignment_option & (1 << host_lane_start_bit):
+            host_lanes_mask = ((1 << host_lane_count) - 1) << host_lane_start_bit
+        else:
+            self.log_error("Unable to find lowest host lane - host_lane_assignment_option {}"
```
prgeor, line 1133:
```suggestion
            self.log_error("Unable to find starting host lane - host_lane_assignment_option {}"
```

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1432,6 +1464,20 @@ def task_worker(self):
                         self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_READY
                         continue
 
+                    appl = self.get_cmis_application_desired(api, host_lane_count, host_speed)
+                    if appl < 1:
+                        self.log_error("{}: no suitable app for the port".format(lport))
+                        self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_FAILED
+                        continue
+
+                    host_lanes_mask = self.get_cmis_host_lanes_mask(api, appl, host_lane_count, channel, speed)
```
prgeor, line 1473: @mihirpat1  can you check if the fn get_cmis_application_desired() and get_cmis_host_lanes_mask() are getting executed during every state machine transition? which is unnecessary? may be we should call once inside the CMIS_STATE_INSERTED ?

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: @mihirpat1 is_cmis_application_update_required() still expects application = 1. can you fix?

prgeor: > @mihirpat1 is_cmis_application_update_required() still expects application = 1. can you fix?

@mihirpat1 also the log "Non-default application is not supported"

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1477,13 +1523,7 @@ def task_worker(self):
...
                                  self.log_notice("{} Successfully configured Tx power = {}".format(lport, tx_power))
 
-                        appl = self.get_cmis_application_desired(api, host_lanes, host_speed)
-                        if appl < 1:
-                            self.log_error("{}: no suitable app for the port".format(lport))
-                            self.port_dict[lport]['cmis_state'] = self.CMIS_STATE_FAILED
-                            continue
-
-                        need_update = self.is_cmis_application_update_required(api, host_lanes, host_speed)
+                        need_update = self.is_cmis_application_update_required(api, host_lane_count, host_lanes_mask, host_speed)
```
prgeor, line 1526: definition of this function takes 'channel' argument name, can we keep both same?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1102,13 +1097,56 @@ def get_cmis_dp_init_duration_secs(self, api):
...
+        else:
+            self.log_error("Unable to find lowest host lane - host_lane_assignment_option {}"
+                            " host_lane_start_bit {} host_lane_count {} channel {} speed {}"
+                            " appl {}!".format(
+                            host_lane_assignment_option, host_lane_start_bit, host_lane_count,
+                            channel, speed, appl))
+
+        return host_lanes_mask
+
+    def is_cmis_application_update_required(self, api, host_lane_count, channel, speed):
```
prgeor, line 1141: @mihirpat1 `channel` arg is not used?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1121,7 +1159,7 @@ def is_cmis_application_update_required(self, api, channel, speed):
         if speed == 0 or channel == 0 or api.is_flat_memory():
             return False
 
-        app_new = self.get_cmis_application_desired(api, channel, speed)
+        app_new = self.get_cmis_application_desired(api, host_lane_count, speed)
```
prgeor, line 1162: 1. current logic checks for all 8 host lanes whereas we should care for the requested host lanes only
2. Also we should make use of the selected/desired application during init state and compare it with the current active application
3. Also we should check for DP != Activated early in the function

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have addressed this now.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have handled #2 now. As discussed, we are already handling #1 and #3 and have changed CMIS_NUM_CHANNELS to CMIS_MAX_HOST_LANES.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have addressed this now.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I am now calling get_cmis_application_desired and get_cmis_host_lanes_mask only in CMIS_STATE_INSERTED. In other states, I am using the value from port_dict dictionary.
I am also making appl and host_lanes_mask to 0 while handling CMIS_STATE_REMOVED.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have fixed this now.

keboliu:
In the Notes, "Once the **port_config.ini** file has a new column for channel, the CONFIG_DB will be updated with the corresponding channel values (requires minigraph reload though)."

I am a bit of concern that to extend "port_config.ini" for this purpose, AFAIK,  "platform.json" was introduced for the dynamic port breakout, I think we should consider extending it instead of "port_config.ini".
by the way, have you already had a PR for this part?

prgeor:
> In the Notes, "Once the **port_config.ini** file has a new column for channel, the CONFIG_DB will be updated with the corresponding channel values (requires minigraph reload though)."
> 
> I am a bit of concern that to extend "port_config.ini" for this purpose, AFAIK, "platform.json" was introduced for the dynamic port breakout, I think we should consider extending it instead of "port_config.ini". by the way, have you already had a PR for this part?

@keboliu  we can remove the reference to port_config.ini as its not relevant to this Xcvrd changes. Please see my email.

keboliu (dismissed):

rajann:
I believe that instead of explicitly giving the subport/channel info in port_config.ini it should be derived from the existing information in port_config.ini (i.e.) we have defined the rules for "name" , "alias" and "index".  Same convention should be following in DPB as well. If there is difference in DPB with the above rules, it should be addressed in the DPB document to match it.  If we diverge here then we will have to do the same additional information in DPB as well.

mihirpat1:
> I believe that instead of explicitly giving the subport/channel info in port_config.ini it should be derived from the existing information in port_config.ini (i.e.) we have defined the rules for "name" , "alias" and "index". Same convention should be following in DPB as well. If there is difference in DPB with the above rules, it should be addressed in the DPB document to match it. If we diverge here then we will have to do the same additional information in DPB as well.

As discussed, having a logic in xcvrd to calculate subport info introduces dependencies on neighboring subports belonging to the same physical group. This also relies on the fact that redis-db will always populate info for all subports in a physical port group before xcvrd actually starts using info for a subport.
In addition to this, the current physical to logical port list is not in sorted order and will need to be sorted in appropriate order before using it.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1102,46 +1097,85 @@ def get_cmis_dp_init_duration_secs(self, api):
...
+
+        Args:
+            api:
+                XcvrApi object
+            appl:
+                Integer, the transceiver-specific application code
+            host_lane_count:
+                Integer, number of lanes on the host side
+            subport:
+                Integer, subport id of the group which the host lanes belong to (1 based),
```
prgeor, line 1112: @mihirpat1 subport:
`1-based logical port number of the physical port after breakout`

prgeor (approved):

keboliu (approved):

shyam77git (approved):

prgeor:
@yxieca please help cherry pick to 202205
@StormLiangMS please help cherry pick to 202211

PR #347, opened 2023-03-16
chassisd: Fix crash on exit on linecard

patrickmacarthur:
#### Description
Set the `config_manager` variable to `None` if we are running on a linecard and thus don't need to set up the config manager.

#### Motivation and Context
During cleanup, the chassid service tries to clean up the `config_manager`, but the `config_manager` variable is only ever initialized if we are on the supervisor. Thus, checking if it is `None` is insufficient because this results in an `UnboundLocalError` that prevents the cleanup from succeeding on a linecard.

This was tested via the sonic-mgmt `platform_tests/daemon/test_chassisd.py` test. Without this change, the `test_pmon_chassisd_stop_and_start_status` test would fail and a traceback would be present in syslog. The test passes with this patch applied.

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -435,6 +435,8 @@ class ChassisdDaemon(daemon_base.DaemonBase):
         if self.module_updater.supervisor_slot == self.module_updater.my_slot:
             config_manager = ConfigManagerTask()
             config_manager.task_run()
+        else:
+            config_manager = None
 
         # Start main loop
         self.log_info("Start daemon main loop")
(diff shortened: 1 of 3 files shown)


wenyiz2021 (commented):
@patrickmacarthur thanks for the fix, can you please add a UT to cover this? PR is blocked by coverage check

abdosi:
@gechiang for viz.

abdosi (dismissed):

gechiang:
@patrickmacarthur can you also include a test to cover the one line change you fixed?  This is required before the change can be merged.
You can reference this original PR to see how the testcases were originally added and add one test to cover your new change for non-supervisor case.
https://github.com/sonic-net/sonic-platform-daemons/pull/97

abdosi:
@bmridul

gechiang (approved):
LGTM

StormLiangMS:
@patrickmacarthur cherry pick to 202211 with conflict, could you submit PR separately for 202211? 
@gechiang  @abdosi for vis.

PR #328, opened 2023-01-03
Chassisd do an explicit stop of the config_manager

judyjoseph:
#### Description
Do an explicit task stop of config_manager task in the chassisd 
 
#### Motivation and Context
Fixes issue https://github.com/sonic-net/sonic-buildimage/issues/11955

#### How Has This Been Tested?
Updated the test_chassisd.py as per https://github.com/sonic-net/sonic-mgmt/pull/7150 and made sure it is a consistent PASS

PR #352, opened 2023-04-20
[202211] chassisd: Fix crash on exit on linecard

patrickmacarthur:
#### Description
Set the `config_manager` variable to `None` if we are running on a
linecard and thus don't need to set up the config manager.

#### Motivation and Context
During cleanup, the chassid service tries to clean up the
`config_manager`, but the `config_manager` variable is only ever
initialized if we are on the supervisor. Thus, checking if it is
`None` is insufficient because this results in an `UnboundLocalError`
that prevents the cleanup from succeeding on a linecard.

This has been tested via both the sonic-mgmt platform_tests test case
for the chassid daemon as well as the updated unit tests in this PR.

#### Additional Information

This is a cherry-pick of PR #347 to 202211 release branch. Note that
PR #328 is also backported to 202211 in this PR as it is required for the
unit test update to pass.

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -435,6 +435,8 @@ class ChassisdDaemon(daemon_base.DaemonBase):
         if self.module_updater.supervisor_slot == self.module_updater.my_slot:
             config_manager = ConfigManagerTask()
             config_manager.task_run()
+        else:
+            config_manager = None
 
         # Start main loop
         self.log_info("Start daemon main loop")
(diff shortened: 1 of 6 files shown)


gechiang (approved):

prgeor:
@StormLiangMS to merge for 202211

kenneth-arista:
@arlakshm Test failure is due to a pipeline infra problem. Specifically,

```
No space left on device : '/home/vsts/work/1/target/b3eeaa0d-789f-4325-9cb3-90f2d8fab6d3.tmp'
```

PR #367, opened 2023-06-01
[PSU power threshold] Fix logic error: compare the system power with the PSU's power threshold

stephenxs:
#### Motivation and Context

Compare the system power instead of PSU power with the PSU's threshold.
On a system with more than one PSU, the power consumption will not distribute evenly on both PSUs, as a result, it gives a wrong message sometimes.
To fix the error, it should compare system power with the PSU's threshold.

Unit test
Manual test
Regression test


liat-grozovik:
@prgeor , @keboliu please help to review/merge

keboliu (approved):

StormLiangMS:
@prgeor could you create ADO for this?

PR #372, opened 2023-06-05
[ycabled] correct the wrong function call for 'config hwmode state'

vdahiya12:
This PR fixes the other sub routine by adding the param `hw_mux_cable_tbl` to 
`handle_config_hwmode_state_cmd_arg_tbl_notification` method, since this is called by wrong params
UT and testing this on DUT


zjswhhh (approved):
Have we scanned the repo to see if all references are updated?

PR #376, opened 2023-06-09
[ycable] change logging level for stop main loop to NOTICE from ERR

vdahiya12:

### DIFF
--- a/sonic-ycabled/ycable/ycable.py
+++ b/sonic-ycabled/ycable/ycable.py
@@ -409,7 +409,7 @@ class DaemonYcable(daemon_base.DaemonBase):
                     os.kill(os.getpid(), signal.SIGKILL)
 
 
-        self.log_error("Stop daemon main loop")
+        self.log_notice("Stop daemon main loop")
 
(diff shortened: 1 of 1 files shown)


zjswhhh (commented):
Hi @vdahiya12 - would it be more appropriate to add it to log analyzer's white-list instead of lowering it to `notice`? 

Is there real impact when we see this log?

vdahiya12:
> Hi @vdahiya12 - would it be more appropriate to add it to log analyzer's white-list instead of lowering it to `notice`?
> 
> Is there real impact when we see this log?

not really, but I think we should change it inside the daemon here as well, since it's not an error .
We could white-list it in sonic-mgmt but image change shouls also be okay.

zjswhhh (approved):

zjswhhh:
Please add proper labels for requested branches.

ayurkiv-nvda:
When will this PR be merged to 202211?

Thanks

moshemos:
@StormLiangMS can you cherry pick this PR to 202211 please?

PR #377, opened 2023-06-13
Add intelligence to xcvrd to understand process restart and config reload

### LINKED ISSUE #356: Add intelligence to xcvrd to understand process restart and config reload
Upon xcvrd/pmon docker restart set the media settings in app DB which sends spurious SI programming from OA upon receiving.  xcvrd should not set the settings upon process restart/pmon restart. 

So add intelligence to xcvrd to understand the process restart and config reload. 

-Create a new table "TABLE_xcvrd" by xcvrd to set a magic number and process restart count in STATE_DB upon fresh start.  (config reload/cold reboot)
-Upon xcvrd process restart /pmon docker restart, xcvrd to check the magic number presence on coming up path to understand whether it is a restart case or a fresh start case.
- Add a hook in config reload CLI handler to clear this magic number upon config reload.
-Mask spurious notification <media setting to app-DB> during process restart
- Possibly another flag can be added to the same table to differentiate between config reload and cold boot.

mihirpat1:
#### Description

Fixes #356 
sonic-net/sonic-buildimage/pull/15453 is also required along with the current changes so that XCVRD is successfully able to receive DEL notification when PROC_INFO table is deleted

Upon xcvrd/pmon docker restart set the media settings in app DB which sends spurious SI programming from OA upon receiving. xcvrd should not set the settings upon process restart/pmon restart.

So add intelligence to xcvrd to understand the process restart and config reload/OA crash.

#### Motivation and Context

Summary of the changes:
1.	XCVRD main thread creates “MEDIA_NOTIFY_REQUIRED” and "CMIS_REINIT_REQUIRED" with value as True in APPL_DB PROC_INFO:XCVRD table for the relevant namespace while coming up if both the keys are not present and sets the port_reinit_request_tbl to True for the relevant ports in the namespace.
	a.	port_reinit_request_tbl is a dictionary which helps the child thread in deciding if media settings renotify and CMIS reinit are required
		1.	If both the keys “MEDIA_NOTIFY_REQUIRED” and "CMIS_REINIT_REQUIRED" are set to False, port_reinit_request_tbl will be set as False. In all other cases, port_reinit_request_tbl will be set as True and both the keys will be set to True as well. 
	b.	“MEDIA_NOTIFY_REQUIRED” is a boolean which states if media notify is required per namespace upon xcvrd boot-up. The SfpStateUpdateTask thread will set this to False after all ports for the corresponding namespace have been initialized 
	c.	"CMIS_REINIT_REQUIRED" is a boolean which states if CMIS reinit is required per namespace upon xcvrd boot-up. The CmisManagerTask thread will set this to False after all ports for the corresponding namespace have been processed (CMIS SM reaches to a steady state for a port)
			1.	Why not bitmap of ports for the keys and boolean instead?
				For creating bitmap, I will need to translate the logical port name (eg Ethernet8) to a number which represents a bit. This might become more difficult to manage on systems with multiple ASICs and/or systems which have multiple speeds supported (such as 400G + 10G on some ports)
2.	If only xcvrd crashes, then “MEDIA_NOTIFY_REQUIRED” and "CMIS_REINIT_REQUIRED" are already set as False, so xcvrd will not renotify media_settings and will not CMIS reinit the modules
3.	XCVRD will subscribe to PROC_INFO:XCVRD table and trigger self-restart if the PROC_INFO:XCVRD table is deleted for the namespace. All threads will be gracefully terminated and xcvrd deinit will be performed followed by issuing a SIGKILL (or some other signal) to ensure XCVRD is restarted automatically by supervisord. After respawn, CMIS re-init and media_settings notified is triggered for the ports belonging to the affected namespace
	1. The drawback here is xcvrd will restart even if OA crashes for just 1 ASIC.
4.	Based on the port_reinit_request_tbl, the CMIS FSM will perform DP_DEACTIVATE and DP reinitialization.
5.	syncd/swss restart clears the entire APP-DB, including the “MEDIA_NOTIFY_REQUIRED” and "CMIS_REINIT_REQUIRED" in PROC_INFO:XCVRD table in the namespace context.

Following is the summary of Unit-test performed

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:w="urn:schemas-microsoft-com:office:word"
xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Word.Document>
<meta name=Generator content="Microsoft Word 15">
<meta name=Originator content="Microsoft Word 15">
<link rel=File-List
href="file:///C:/Users/patelmi/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

<link rel=themeData
href="file:///C:/Users/patelmi/AppData/Local/Temp/msohtmlclip1/01/clip_themedata.thmx">
<link rel=colorSchemeMapping
href="file:///C:/Users/patelmi/AppData/Local/Temp/msohtmlclip1/01/clip_colorschememapping.xml">

<style>

</style>

</head>

<body lang=EN-US style='tab-interval:.5in;word-wrap:break-word'>

Event | APPL_DB cleared | Xcvrd restarted | Media renotify | CMIS re-init   triggered | Link flap
-- | -- | -- | -- | -- | --
Xcvrd restart | N | Y | N | N | N
Pmon restart | N | Y | N | N | N
Swss restart | Y | Y | Y | Y | Y
Syncd restart | Y | Y | Y | Y | Y
config reload | Y | Y – Restarted   through config reload and not due to DEL event handling | Y | Y | Y
Cold reboot | Y | Y | Y | Y | Y
Config shut | N | N | N | N | Y
Config no shut | N | N | N | N | Y

</body>

</html>

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -937,7 +937,7 @@ class CmisManagerTask(threading.Thread):
         self.isPortConfigDone = False
         self.skip_cmis_mgr = skip_cmis_mgr
         self.namespaces = namespaces
-        self.port_reinit_request_tbl = port_reinit_request_tbl
+        self.port_reinit_request_tbl = copy.deepcopy(port_reinit_request_tbl)
 
     def log_notice(self, message):
         helper_logger.log_notice("CMIS: {}".format(message))
@@ -1857,7 +1857,7 @@ class SfpStateUpdateTask(threading.Thread):
         self.sfp_error_dict = {}
         self.sfp_insert_events = {}
         self.namespaces = namespaces
-        self.port_reinit_request_tbl = port_reinit_request_tbl
+        self.port_reinit_request_tbl = copy.deepcopy(port_reinit_request_tbl)
 
     def _mapping_event_from_change_event(self, status, port_dict):
         """
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -24,6 +24,7 @@ try:
     from sonic_py_common import daemon_base, device_info, logger
     from sonic_py_common import multi_asic
     from swsscommon import swsscommon
+    from swsscommon.swsscommon import SonicV2Connector
 
     from .xcvrd_utilities import sfp_status_helper
     from .xcvrd_utilities import port_mapping
@@ -46,6 +47,8 @@ TRANSCEIVER_STATUS_TABLE = 'TRANSCEIVER_STATUS'
 TRANSCEIVER_PM_TABLE = 'TRANSCEIVER_PM'
 
 TRANSCEIVER_STATUS_TABLE_SW_FIELDS = ["status", "error"]
+
+APPL_PROC_INFO_TABLE = 'PROC_INFO'
 
 # Mgminit time required as per CMIS spec
 MGMT_INIT_TIME_DELAY_SECS = 2
@@ -856,6 +859,43 @@ def delete_port_from_status_table_hw(logical_port_name, port_mapping, status_tbl
                 continue
             status_tbl.hdel(physical_port_name, f)
 
+# Initialize the APPL_DB PROC_INFO:XCVRD table with XCVRD_COLD_RESTART as the key
+# if not already populated and update the port_reinit_request_tbl accordingly
+# XCVRD_COLD_RESTART is the key which is used to determine if ports in a namespace
+# need to be reinitialized and media settings are to be renotified
+def init_appl_proc_info_tbl(namespaces, port_mapping_data, xcvr_table_helper):
+    namespace_reinit_request_tbl = {}
+    port_reinit_request_tbl = {}
+
+    for namespace in namespaces:
+        asic_id = multi_asic.get_asic_index_from_namespace(namespace)
+        namespace_reinit_request_tbl[asic_id] = False
+        appl_db = SonicV2Connector(use_unix_socket_path=False, namespace=namespace)
+        if appl_db is None:
+            helper_logger.log_error("Appl {} tbl is None for asic_id {} during init".format(APPL_PROC_INFO_TABLE, asic_id))
+            continue
+        appl_db.connect(appl_db.APPL_DB)
+        APPL_PROC_INFO_XCVRD_TABLE = '{}:XCVRD'.format(APPL_PROC_INFO_TABLE)
+        proc_info_xcvrd_dict = appl_db.get_all(appl_db.APPL_DB, '{}'.format(APPL_PROC_INFO_XCVRD_TABLE))
+        if proc_info_xcvrd_dict is not None and 'XCVRD_COLD_RESTART' in proc_info_xcvrd_dict:
+            helper_logger.log_notice("Key XCVRD_COLD_RESTART already exists in {} tbl for asic {}!".format(APPL_PROC_INFO_XCVRD_TABLE, asic_id))
+        else:
+            appl_db.set(appl_db.APPL_DB, '{}:XCVRD'.format(APPL_PROC_INFO_TABLE), 'XCVRD_COLD_RESTART', 'true')
+            namespace_reinit_request_tbl[asic_id] = True
+            helper_logger.log_notice("Added {} tbl for asic {}!".format(APPL_PROC_INFO_XCVRD_TABLE, asic_id))
+        appl_db.close(appl_db.APPL_DB)
+
+    for logical_port_name in port_mapping_data.logical_port_list:
+        port_reinit_request_tbl[logical_port_name] = False
+        # Get the asic to which this port belongs
+        asic_index = port_mapping_data.get_asic_id_for_logical_port(logical_port_name)
+        if asic_index is None:
+            helper_logger.log_warning("Got invalid asic index for {} during appl PROC_INFO tbl init".format(logical_port_name))
+            continue
+        port_reinit_request_tbl[logical_port_name] = namespace_reinit_request_tbl[asic_index]
+
+    return port_reinit_request_tbl
+
 def is_fast_reboot_enabled():
     fastboot_enabled = subprocess.check_output('sonic-db-cli STATE_DB hget "FAST_RESTART_ENABLE_TABLE|system" enable', shell=True, universal_newlines=True)
     return "true" in fastboot_enabled
@@ -884,7 +924,7 @@ class CmisManagerTask(threading.Thread):
     CMIS_STATE_REMOVED   = 'REMOVED'
     CMIS_STATE_FAILED    = 'FAILED'
 
-    def __init__(self, namespaces, port_mapping, main_thread_stop_event, skip_cmis_mgr=False):
+    def __init__(self, namespaces, port_mapping, main_thread_stop_event, port_reinit_request_tbl, skip_cmis_mgr=False):
         threading.Thread.__init__(self)
         self.name = "CmisManagerTask"
         self.exc = None
@@ -897,6 +937,7 @@ class CmisManagerTask(threading.Thread):
         self.isPortConfigDone = False
         self.skip_cmis_mgr = skip_cmis_mgr
         self.namespaces = namespaces
+        self.port_reinit_request_tbl = port_reinit_request_tbl
 
     def log_notice(self, message):
         helper_logger.log_notice("CMIS: {}".format(message))
@@ -1534,6 +1575,12 @@ class CmisManagerTask(threading.Thread):
                            if 0 != freq and freq != api.get_laser_config_freq():
                               need_update = True
 
+                        if self.port_reinit_request_tbl[lport] == True:
+                            # To ensure forced CMIS init is done only once for a port after xcvrd boot-up
+                            self.port_reinit_request_tbl[lport] = False
+                            need_update = True
+                            self.log_notice("{}: Forcing CMIS init as port_reinit_request_tbl is True".format(lport))
+
                         if not need_update:
                             # No application updates
                             self.log_notice("{}: no CMIS application update required...READY".format(lport))
@@ -1793,7 +1840,7 @@ class DomInfoUpdateTask(threading.Thread):
 
 class SfpStateUpdateTask(threading.Thread):
     RETRY_EEPROM_READING_INTERVAL = 60
-    def __init__(self, namespaces, port_mapping, main_thread_stop_event, sfp_error_event):
+    def __init__(self, namespaces, port_mapping, port_reinit_request_tbl, main_thread_stop_event, sfp_error_event):
         threading.Thread.__init__(self)
         self.name = "SfpStateUpdateTask"
         self.exc = None
@@ -1810,6 +1857,7 @@ class SfpStateUpdateTask(threading.Thread):
         self.sfp_error_dict = {}
         self.sfp_insert_events = {}
         self.namespaces = namespaces
+        self.port_reinit_request_tbl = port_reinit_request_tbl
 
     def _mapping_event_from_change_event(self, status, port_dict):
         """
@@ -1841,11 +1889,6 @@ class SfpStateUpdateTask(threading.Thread):
         transceiver_dict = {}
         retry_eeprom_set = set()
 
-        warmstart = swsscommon.WarmStart()
-        warmstart.initialize("xcvrd", "pmon")
-        warmstart.checkWarmStart("xcvrd", "pmon", False)
-        is_warm_start = warmstart.isWarmStart()
-
         # Post all the current interface sfp/dom threshold info to STATE_DB
         logical_port_list = port_mapping.logical_port_list
         for logical_port_name in logical_port_list:
@@ -1862,7 +1905,8 @@ class SfpStateUpdateTask(threading.Thread):
                 post_port_dom_threshold_info_to_db(logical_port_name, port_mapping, xcvr_table_helper.get_dom_threshold_tbl(asic_index), stop_event)
 
                 # Do not notify media settings during warm reboot to avoid dataplane traffic impact
-                if is_warm_start == False:
+                if self.port_reinit_request_tbl[logical_port_name] == True:
+                    helper_logger.log_notice("Notify media setting during boot-up for {}".format(logical_port_name))
                     notify_media_setting(logical_port_name, transceiver_dict, xcvr_table_helper.get_app_port_tbl(asic_index), port_mapping)
                     transceiver_dict.clear()
             else:
@@ -2330,6 +2374,7 @@ class DaemonXcvrd(daemon_base.DaemonBase):
         self.skip_cmis_mgr = skip_cmis_mgr
         self.namespaces = ['']
         self.threads = []
+        self.port_reinit_request_tbl = {}
 
     # Signal handler
     def signal_handler(self, sig, frame):
@@ -2377,6 +2422,16 @@ class DaemonXcvrd(daemon_base.DaemonBase):
 
         with open(media_settings_file_path, "r") as media_file:
             g_dict = json.load(media_file)
+
+    def subscribe_appl_proc_info_update(self):
+        for namespace in self.namespaces:
+            appl_db = daemon_base.db_connect("APPL_DB", namespace=namespace)
+
+            sel = swsscommon.Select()
+            proc_info_tbl = swsscommon.SubscriberStateTable(appl_db, APPL_PROC_INFO_TABLE)
+            sel.addSelectable(proc_info_tbl)
+
+            return sel, proc_info_tbl
 
     # Initialize daemon
     def init(self):
@@ -2429,7 +2484,12 @@ class DaemonXcvrd(daemon_base.DaemonBase):
             self.wait_for_port_config_done(namespace)
 
         self.log_notice("XCVRD INIT: After port config is done")
-        return port_mapping.get_port_mapping(self.namespaces)
+        port_mapping_data = port_mapping.get_port_mapping(self.namespaces)
+
+        self.port_reinit_request_tbl = init_appl_proc_info_tbl(self.namespaces, port_mapping_data, self.xcvr_table_helper)
+        self.log_info("port_reinit_request_tbl value after init is: {}".format(self.port_reinit_request_tbl))
+
+        return port_mapping_data
 
     # Deinitialize daemon
     def deinit(self):
@@ -2465,7 +2525,7 @@ class DaemonXcvrd(daemon_base.DaemonBase):
         port_mapping_data = self.init()
 
         # Start the CMIS manager
-        cmis_manager = CmisManagerTask(self.namespaces, port_mapping_data, self.stop_event, self.skip_cmis_mgr)
+        cmis_manager = CmisManagerTask(self.namespaces, port_mapping_data, self.stop_event, self.port_reinit_request_tbl, self.skip_cmis_mgr)
         if not self.skip_cmis_mgr:
             cmis_manager.start()
             self.threads.append(cmis_manager)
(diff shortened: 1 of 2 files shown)


mihirpat1:
@amnsinghal @shyam77git - It will be great if you can help in reviewing this

Junchao-Mellanox:
What is the use case to restart xcvrd only?

mihirpat1:
> What is the use case to restart xcvrd only?

Restarting xcvrd after OA crash will now allow us with the below 2 things:
1. CmisManagerTask thread can initiate CMIS re-init only for the ports belonging to the affected namespace with init routine itself
2. Media setting can be renotified for the ports belonging to the affected namespace

Following alternate options were evaluated before concluding on the xcvrd restart part.
1. Perform TX Disable upon APPL_DB PORT_DEL event in CMIS thread
	1. Taking this approach would mean we will need to decide on the thread responsible for populating the APPL_DB PROC_INFO:XCVRD table entry? We would need to narrow down between Xcvrd main thread or CMIS manager since initially, xcvrd populated the PROC_INFO:XCVRD table entry
	2. Also, performing TX Disable upon receiving APPL_DB DEL event in CMIS thread will now require SfpStateUpdateTask thread to re-notify the media settings.
	
2. If XCVRD main thread handles APPL_DB DEL event handling, we would need a way to notify other threads of this action. If xcvrd main thread puts the port in Disabled and DeInit state, we would need to modify the CMIS SM from the main thread as well.

Overall, both of the above approaches would increase dependency between the threads and hence, we chose to proceed with the xcvrd restart option

snider-nokia:
>XCVRD will subscribe to PROC_INFO:XCVRD table and trigger self-restart if the XCVRD_COLD_RESTART is deleted for the namespace. XCVRD will be killed using **SIGKILL**. After respawn, CMIS re-init and media_settings notified is triggered for the ports belonging to the affected namespace

Please don't use SIGKILL. SIGTERM is more appropriate, just in case there is platform specific cleanup that needs to done.

mihirpat1:
> > XCVRD will subscribe to PROC_INFO:XCVRD table and trigger self-restart if the XCVRD_COLD_RESTART is deleted for the namespace. XCVRD will be killed using **SIGKILL**. After respawn, CMIS re-init and media_settings notified is triggered for the ports belonging to the affected namespace
> 
> Please don't use SIGKILL. SIGTERM is more appropriate, just in case there is platform specific cleanup that needs to done.

Since the default "stopsignal" setting in supervisord config is TERM for xcvrd, we don't see xcvrd restarting after receiving a SIGTERM signal. Hence, I opted to do SIGKILL.

snider-nokia:
> > > XCVRD will subscribe to PROC_INFO:XCVRD table and trigger self-restart if the XCVRD_COLD_RESTART is deleted for the namespace. XCVRD will be killed using **SIGKILL**. After respawn, CMIS re-init and media_settings notified is triggered for the ports belonging to the affected namespace
> > 
> > 
> > Please don't use SIGKILL. SIGTERM is more appropriate, just in case there is platform specific cleanup that needs to done.
> 
> Since the default "stopsignal" setting in supervisord config is TERM for xcvrd, we don't see xcvrd restarting after receiving a SIGTERM signal. Hence, I opted to do SIGKILL.

If SIGKILL is used then the shutdown will (obviously) be graceless and our platform will be deprived of doing the graceful shutdown related things that are warranted.  I implore you to reconsider the use of SIGKILL, as standard practice dictates that it really should only be used as a last resort.  SIGABRT would work too.

mihirpat1:
> > > > XCVRD will subscribe to PROC_INFO:XCVRD table and trigger self-restart if the XCVRD_COLD_RESTART is deleted for the namespace. XCVRD will be killed using **SIGKILL**. After respawn, CMIS re-init and media_settings notified is triggered for the ports belonging to the affected namespace
> > > 
> > > 
> > > Please don't use SIGKILL. SIGTERM is more appropriate, just in case there is platform specific cleanup that needs to done.
> > 
> > 
> > Since the default "stopsignal" setting in supervisord config is TERM for xcvrd, we don't see xcvrd restarting after receiving a SIGTERM signal. Hence, I opted to do SIGKILL.
> 
> If SIGKILL is used then the shutdown will (obviously) be graceless and our platform will be deprived of doing the graceful shutdown related things that are warranted. I implore you to reconsider the use of SIGKILL, as standard practice dictates that it really should only be used as a last resort. SIGABRT would work too.

Sure, will explore into this and try to handle graceful shutdown.

snider-nokia:
> > If SIGKILL is used then the shutdown will (obviously) be graceless and our platform will be deprived of doing the graceful shutdown related things that are warranted. I implore you to reconsider the use of SIGKILL, as standard practice dictates that it really should only be used as a last resort. SIGABRT would work too.
> 
> Sure, will explore into this and try to handle graceful shutdown.

Thanks.   Along the same lines, hope you will also make the following change (or something similar) to force the main Xcvrd thread to cause underlying platform initialization of SFP (transceiver) subsystem prior to the other Xcvrd threads being spun up.  The reason this is requested is because Python does not allow signal handlers to be installed/set from anywhere except main thread.

For platforms with meaningful signal handling implemented (like ours), something of this sort is a requirement, otherwise signal handler(s) will not be allowed by Python to be installed/setup.

image

image

Junchao-Mellanox:
I just have another thought: why not fix the issue from portsorch.cpp side? Here is my proposal:

In portsorch.cpp, we shall cache all the serdes attributes. We shall compare the cached serdes attributes values with incoming values while doPortTask is called. If the value equals, no need apply the value to SAI.

Any comment? @prgeor @mihirpat1

prgeor:
> I just have another thought: why not fix the issue from portsorch.cpp side? Here is my proposal:
> 
> In portsorch.cpp, we shall cache all the serdes attributes. We shall compare the cached serdes attributes values with incoming values while doPortTask is called. If the value equals, no need apply the value to SAI.
> 
> Any comment? @prgeor @mihirpat1



> I just have another thought: why not fix the issue from portsorch.cpp side? Here is my proposal:
> 
> In portsorch.cpp, we shall cache all the serdes attributes. We shall compare the cached serdes attributes values with incoming values while doPortTask is called. If the value equals, no need apply the value to SAI.
> 
> Any comment? @prgeor @mihirpat1

@Junchao-Mellanox its not a good idea for OA to decide when to apply the media setting because consider a case where same cable is removed and plugged in, it should re-apply same media setting (as previous values) to keep various scenarios where Xcvrd need to init the CMIS state machine during various cases like config reload, swss/syncd crash

Junchao-Mellanox:
> > I just have another thought: why not fix the issue from portsorch.cpp side? Here is my proposal:
> > In portsorch.cpp, we shall cache all the serdes attributes. We shall compare the cached serdes attributes values with incoming values while doPortTask is called. If the value equals, no need apply the value to SAI.
> > Any comment? @prgeor @mihirpat1
> 
> > I just have another thought: why not fix the issue from portsorch.cpp side? Here is my proposal:
> > In portsorch.cpp, we shall cache all the serdes attributes. We shall compare the cached serdes attributes values with incoming values while doPortTask is called. If the value equals, no need apply the value to SAI.
> > Any comment? @prgeor @mihirpat1
> 
> @Junchao-Mellanox its not a good idea for OA to decide when to apply the media setting because consider a case where same cable is removed and plugged in, it should re-apply same media setting (as previous values) to keep various scenarios where Xcvrd need to init the CMIS state machine during various cases like config reload, swss/syncd crash

Thanks for the reply. For cable re-plug case, I suppose xcvrd should remove the media settings from APPL DB and re-apply it explicitly. The current solution looks a litter bit complicated to me.

mihirpat1:
> > > If SIGKILL is used then the shutdown will (obviously) be graceless and our platform will be deprived of doing the graceful shutdown related things that are warranted. I implore you to reconsider the use of SIGKILL, as standard practice dictates that it really should only be used as a last resort. SIGABRT would work too.
> > 
> > 
> > Sure, will explore into this and try to handle graceful shutdown.
> 
> Thanks. Along the same lines, hope you will also make the following change (or something similar) to force the main Xcvrd thread to cause underlying platform initialization of SFP (transceiver) subsystem prior to the other Xcvrd threads being spun up. The reason this is requested is because Python does not allow signal handlers to be installed/set from anywhere except main thread.
> 
> For platforms with meaningful signal handling implemented (like ours), something of this sort is a requirement, otherwise signal handler(s) will not be allowed by Python to be installed/setup.
> 
> image
> 
> image

I have now addressed this

mihirpat1:
> > > I just have another thought: why not fix the issue from portsorch.cpp side? Here is my proposal:
> > > In portsorch.cpp, we shall cache all the serdes attributes. We shall compare the cached serdes attributes values with incoming values while doPortTask is called. If the value equals, no need apply the value to SAI.
> > > Any comment? @prgeor @mihirpat1
> > 
> > 
> > > I just have another thought: why not fix the issue from portsorch.cpp side? Here is my proposal:
> > > In portsorch.cpp, we shall cache all the serdes attributes. We shall compare the cached serdes attributes values with incoming values while doPortTask is called. If the value equals, no need apply the value to SAI.
> > > Any comment? @prgeor @mihirpat1
> > 
> > 
> > @Junchao-Mellanox its not a good idea for OA to decide when to apply the media setting because consider a case where same cable is removed and plugged in, it should re-apply same media setting (as previous values) to keep various scenarios where Xcvrd need to init the CMIS state machine during various cases like config reload, swss/syncd crash
> 
> Thanks for the reply. For cable re-plug case, I suppose xcvrd should remove the media settings from APPL DB and re-apply it explicitly. The current solution looks a litter bit complicated to me.

@Junchao-Mellanox - The current solution will update media setting only in the case of switch boot-up or OA crash. For the cable re-plug case, the below existing code will handle the corresponding scenario
https://github.com/sonic-net/sonic-platform-daemons/blob/81048cd1fb63da9fdca567aa863c7d1d43156ed3/sonic-xcvrd/xcvrd/xcvrd.py#L2088

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -134,6 +136,17 @@ def get_physical_port_name_dict(logical_port_name, port_mapping):
 
     return port_name_dict
 
+# Get dictionary of asic to set of corresponding logical ports
+def get_asic_to_logical_ports_dict(port_mapping):
```
Junchao-Mellanox, line 140: should we move this function to src/sonic-platform-daemons/sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py and make it a member of PortMapping class?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -856,6 +878,46 @@ def delete_port_from_status_table_hw(logical_port_name, port_mapping, status_tbl
...
+        proc_info_tbl = xcvr_table_helper.get_appl_proc_info_tbl(asic_id)
+        if proc_info_tbl is None:
+            helper_logger.log_error("Appl {} tbl is None for asic_id {} during init".format(APPL_PROC_INFO_TABLE, asic_id))
+            continue
+        _, proc_info_xcvrd_fvs = proc_info_tbl.get('XCVRD')
+        if proc_info_xcvrd_fvs is not None:
+            proc_info_xcvrd_fvs_dict = dict(proc_info_xcvrd_fvs)
+            media_notify_rqd = proc_info_xcvrd_fvs_dict.get('MEDIA_NOTIFY_REQUIRED', 'True')
+            cmis_reinit_rqd = proc_info_xcvrd_fvs_dict.get('CMIS_REINIT_REQUIRED', 'True')
+            if media_notify_rqd == 'False' and cmis_reinit_rqd == 'False':
```
Junchao-Mellanox, line 903: Is it possible that media_notify_rqd is "True" but cmis_reinit_rqd is "False" or vice versa?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1385,6 +1453,14 @@ def task_worker(self):
                              self.CMIS_STATE_FAILED,
                              self.CMIS_STATE_READY,
                              self.CMIS_STATE_REMOVED]:
+                    if self.port_reinit_request_tbl[lport] is True:
+                        self.log_info("{}: Setting port_reinit_request_tbl to True".format(lport))
```
Junchao-Mellanox, line 1457: typo？ True -> False?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2524,12 +2656,15 @@ def run(self):
 
         if self.sfp_error_event.is_set():
             sys.exit(SFP_SYSTEM_ERROR)
+        elif generate_sigabrt is True:
+            self.log_notice("Sending SIGABRT to xcvrd!")
```
Junchao-Mellanox, line 2660: Why do we send a SIGABRT here?

Junchao-Mellanox (changes requested):

Junchao-Mellanox:
One question, maybe not relevant: how could CmisManagerTask re-initialize relevant logical port if a CMIS cable is replaced by another CMIS cable (plug out old one and plug in new one)?

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: Yes - that is a possibility if either CmisManagerTask or SfpStateUpdateTask thread crashes while performing the initialization.
In such scenario, xcvrd would perform the CMIS reinit and media setting notify for all ports in the affected namespace (irrespective of the completion status of media settings and CMIS reinit of ports before the thread got killed)

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: Thanks! Will correct it in the new diff.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: This was done for adhering to the below comment
https://github.com/sonic-net/sonic-platform-daemons/pull/377#issuecomment-1593737539

mihirpat1:
> One question, maybe not relevant: how could CmisManagerTask re-initialize relevant logical port if a CMIS cable is replaced by another CMIS cable (plug out old one and plug in new one)?

In such case, CmisManagerTask will be receiving Sfp insertion/removal event due to a change in TRANSCEIVER_INFO table. Following is the code flow for it. 
https://github.com/sonic-net/sonic-platform-daemons/blob/81048cd1fb63da9fdca567aa863c7d1d43156ed3/sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py#L118

https://github.com/sonic-net/sonic-platform-daemons/blob/81048cd1fb63da9fdca567aa863c7d1d43156ed3/sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py#L147

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I did not put it in PortMapping class since this class is more of tied to a single port and the corresponding mapping for it such as its physical/logical port number or asic ID.
However, the function get_asic_to_logical_ports_dict is tied to a ASIC and it maps to a set of ports. Hence, I added this in xcvrd and not as a member of PortMapping class.

Junchao-Mellanox:
> > One question, maybe not relevant: how could CmisManagerTask re-initialize relevant logical port if a CMIS cable is replaced by another CMIS cable (plug out old one and plug in new one)?
> 
> In such case, CmisManagerTask will be receiving Sfp insertion/removal event due to a change in TRANSCEIVER_INFO table. Following is the code flow for it.
> 
> https://github.com/sonic-net/sonic-platform-daemons/blob/81048cd1fb63da9fdca567aa863c7d1d43156ed3/sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py#L118
> 
> https://github.com/sonic-net/sonic-platform-daemons/blob/81048cd1fb63da9fdca567aa863c7d1d43156ed3/sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py#L147

Thanks

sonic-xcvrd/xcvrd/xcvrd.py
Junchao-Mellanox: In a normal reboot/config reload flow, is it possible to run into this case:
1. xcvrd has started and called self.init (xcvrd has initialized PROC_INFO table)
2. swss starts after and clears PROC_INFO table
3. xcvrd runs into while loop to select PROC_INFO table later, but it does not get the delete event

In such case, PROC_INFO table is empty, but xcvrd has already called notify_media_settings. The state is incorrect.

mihirpat1: In this case, starting of swss will not clear PROC_INFO table. Only if swss crashes, PROC_INFO table will be cleared.

Also, xcvrd will initialize PROC_INFO table after port config is done. The CmisManagerTask and SfpStateUpdateTask threads also check for port config done before performing their respective initialization.
Following is the function call from xcvrd.init() to ensure port config done is set prior to performing further actions.
https://github.com/sonic-net/sonic-platform-daemons/blob/81048cd1fb63da9fdca567aa863c7d1d43156ed3/sonic-xcvrd/xcvrd/xcvrd.py#L2429

prgeor:
> > > I just have another thought: why not fix the issue from portsorch.cpp side? Here is my proposal:
> > > In portsorch.cpp, we shall cache all the serdes attributes. We shall compare the cached serdes attributes values with incoming values while doPortTask is called. If the value equals, no need apply the value to SAI.
> > > Any comment? @prgeor @mihirpat1
> > 
> > 
> > > I just have another thought: why not fix the issue from portsorch.cpp side? Here is my proposal:
> > > In portsorch.cpp, we shall cache all the serdes attributes. We shall compare the cached serdes attributes values with incoming values while doPortTask is called. If the value equals, no need apply the value to SAI.
> > > Any comment? @prgeor @mihirpat1
> > 
> > 
> > @Junchao-Mellanox its not a good idea for OA to decide when to apply the media setting because consider a case where same cable is removed and plugged in, it should re-apply same media setting (as previous values) to keep various scenarios where Xcvrd need to init the CMIS state machine during various cases like config reload, swss/syncd crash
> 
> Thanks for the reply. For cable re-plug case, I suppose xcvrd should remove the media settings from APPL DB and re-apply it explicitly. The current solution looks a litter bit complicated to me.

@Junchao-Mellanox agreed. @mihirpat1 ?

mihirpat1:
> > > > I just have another thought: why not fix the issue from portsorch.cpp side? Here is my proposal:
> > > > In portsorch.cpp, we shall cache all the serdes attributes. We shall compare the cached serdes attributes values with incoming values while doPortTask is called. If the value equals, no need apply the value to SAI.
> > > > Any comment? @prgeor @mihirpat1
> > > 
> > > 
> > > > I just have another thought: why not fix the issue from portsorch.cpp side? Here is my proposal:
> > > > In portsorch.cpp, we shall cache all the serdes attributes. We shall compare the cached serdes attributes values with incoming values while doPortTask is called. If the value equals, no need apply the value to SAI.
> > > > Any comment? @prgeor @mihirpat1
> > > 
> > > 
> > > @Junchao-Mellanox its not a good idea for OA to decide when to apply the media setting because consider a case where same cable is removed and plugged in, it should re-apply same media setting (as previous values) to keep various scenarios where Xcvrd need to init the CMIS state machine during various cases like config reload, swss/syncd crash
> > 
> > 
> > Thanks for the reply. For cable re-plug case, I suppose xcvrd should remove the media settings from APPL DB and re-apply it explicitly. The current solution looks a litter bit complicated to me.
> 
> @Junchao-Mellanox agreed. @mihirpat1 ?

Agreed

Junchao-Mellanox (approved):

mihirpat1:
Closing the current PR since planning to open a new PR once the changes inline to the HLD https://github.com/sonic-net/SONiC/pull/1432 are completed.

PR #383, opened 2023-06-26
Add deterministic link bring-up feature for SFF compliant modules

longhuan-cisco:
HLD update is in different PR (link: https://github.com/sonic-net/SONiC/pull/1371)

#### Description
According to [Interface-Link-bring-up-sequence.md](https://github.com/sonic-net/SONiC/blob/fde55e4f3f32fa3b91d42574c31773ab6610e954/doc/sfp-cmis/Interface-Link-bring-up-sequence.md#plan), add a new thread sff_mgr under xcvrd to provide deterministic link bringup feature for SFF compliant modules (100G/40G).

By default sff_mgr is disabled, to enable sff_mgr on a platform, add 'enable_xcvrd_sff_mgr' and set it to true in pmon_daemon_ctrl json file.

#### Motivation and Context

- **Scope of sff_mgr:** 100G/40G optics (copper/aoc not in the scope)

- **Why sff_mgr**
The goal of sff_mgr is to make sure SFF compliant modules are brought up in a deterministc way, meaning TX is enabled only after host_tx_ready becomes True, and TX will be disabled when host_tx_ready becomes False. This will help eliminate link stability issue and potential interface flap, also turning off TX reduces the power consumption and avoid any lab hazard for admin shut interface.

- **What sff_mgr does**
sff_mgr is a new thread inside Xcvrd. It will only turn on module Tx if both admin_status=up and host_tx_ready=true.
sff_mgr will skip if neither of these below events happens: `transceiver insertion event` (including bootup and hot plugin), `host_tx_ready change event` and `admin_status change event`. All other cases are ignored.
To detect these events, sff_mgr listens to below DB tables:
  - STATE_DB PORT_TABLE's `host_tx_ready` field for host_tx_ready change event.
  - STATE_DB TRANSCEIVER_INFO table's `type` field for insesrtion event. 
  - CONFIG_DB PORT_TABLE's `admin_status` field for admin_status change event, and info such as `index`/`channel`/etc.

- **For platforms/vendors:**
        This feature is enabled on per platform basis. There could be cases where vendor(s)/platform(s) may take time to shift from existing codebase to the model (work-flows). 
By default this feature is disabled. No impact for the platforms in current deployment. To enable it, a flag needs to be added to pmon_daemon_control.json and set to true.
  - **pre-requisite to enable this feature**: platform needs to keep TX in disabled state after module coming out-of-reset, in either module insertion or bootup cases. This is to make sure the module is not transmitting with TX enabled before host_tx_ready becomes True. 

1. On platform with sff_mgr enabled which has 100G and 40G optics, verified TX is enabled/disabled properly based on host_tx_ready and in proper order, under cases of module insertion/system bootup.  And no-op for the cases of xcvrd process restart. 
4. Verified TX is enabled/disabled properly along with "config interface shutdown/startup" (which changes admin_status and host_tx_ready value)
5. Verified no impact on platform without sff_mgr enabled.


longhuan-cisco:
@prgeor  @mihirpat1 could you please take a look at the PR?

CC @shyam77git @jaganbal-a

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,393 @@
...
+                    if not api.get_tx_disable_support():
+                        self.log_notice(
+                            "{}: skipping sff_mgr due to tx_disable not supported".format(
+                                lport))
+                        continue
+                except AttributeError:
+                    # Skip if these essential routines are not available
+                    continue
+
+                target_tx_disable_flag = data['host_tx_ready'] == 'false'
```
longhuan-cisco, line 364: Prince suggested: _subscribes to admin_status as well, to make sure only turn on TX if both admin_status and host_tx_ready are true_

longhuan-cisco: added admin_status check

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2456,6 +2459,46 @@ def deinit(self):
 
         del globals()['platform_chassis']
 
+    def load_feature_flags(self):
```
keboliu, line 2462: why not use a similar solution as the way to control the disable/enable of cmis manager task?
https://github.com/sonic-net/sonic-buildimage/blob/9b08fe4eb2fdff9d9e9ebed640100abb3b9aeef8/dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2#L99

longhuan-cisco: The reasons are: 
1.  With this method, user can easily enable/disable this feature at runtime for dev/debug/workaround/etc purposes. Just change the flag in pmon json, and restart xcvrd. 
2.  With this method, now xcvrd directly gets the flag from pmon json file, which is more simplified than the old method. Old method needs to first get the flag from json file, generate the supervisord.conf at build time, pass it as xcvrd process startup argument, and finally xcvrd parse the process startup argument to get the flag.

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,407 @@
...
+                    # all the interested channels, to try best-effort TX disable/enable.
+                    cur_tx_disable_array = [not target_tx_disable_flag] * self.DEFAULT_NUM_CHANNELS
+                # Get an array of bool, where it's True only on the channels that need change.
+                delta_array = self.calculate_tx_disable_delta_array(cur_tx_disable_array,
+                                                                    target_tx_disable_flag, channel)
+                mask = self.convert_bool_array_to_bit_mask(delta_array)
+                if mask == 0:
+                    self.log_notice("{}: No change is needed for tx_disable value".format(lport))
+                    continue
+                if api.tx_disable_channel(mask, target_tx_disable_flag):
```
eddyk-nvidia, line 397: Should we bring a module to high-power before enabling TX ? In CMIS State machine it is done as a part of CMIS Data Path SM

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,407 @@
...
+                    self.log_notice("{}: TX was {} with channel mask: {}".format(
+                        lport, "disabled" if target_tx_disable_flag else "enabled", bin(mask)))
+                else:
+                    self.log_error("{}: Failed to {} TX with channel mask: {}".format(
+                        lport, "disable" if target_tx_disable_flag else "enable", bin(mask)))
+
```
eddyk-nvidia, line 403: Do you plan to add the Custom module SI parameters here ?

sonic-xcvrd/xcvrd/sff_mgr.py
longhuan-cisco: For now, non-cmis/ccmis optics are brought to High Power mode by default by platform.
The missing part is: enabling high power class by setting page 0 byte 93 bit 2(if advertised power class >=5). Yes, this handling will be added to sff_mgr in separate PR.

sonic-xcvrd/xcvrd/sff_mgr.py
longhuan-cisco: Yes, custom optics SI is supposed to be handled by sff_mgr. The detail is still in discussion. This will be added via separate PR also.

eddyk-nvidia: Don't we need also custom SI settings for copper modules managed with SFF ? I know that as of now the SFF manager handles optics only

prgeor:
@mihirpat1 could you review?

prgeor:
@longhuan-cisco could you check the conflict

keboliu (approved):

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,407 @@
...
+                    lport,
+                    xcvr_inserted,
+                    data['host_tx_ready'], host_tx_ready_changed,
+                    data['admin_status'], admin_status_changed))
+
+                # double-check the HW presence before moving forward
+                sfp = self.platform_chassis.get_sfp(pport)
+                if not sfp.get_presence():
+                    self.log_error("{}: module not present!".format(lport))
+                    del self.port_dict[lport]
```
Junchao-Mellanox, line 345: maybe only remove "type" field from the self.port_dict is enough?

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,407 @@
...
+        if port_change_event.event_type == port_change_event.PORT_SET:
+            if lport not in self.port_dict:
+                self.port_dict[lport] = {}
+            if pport >= 0:
+                self.port_dict[lport]['index'] = pport
+            # This field comes from CONFIG_DB PORT_TABLE. This is the channel
+            # that blongs to this logical port, 0 means all channels. tx_disable
+            # API needs to know which channels to disable/enable for a
+            # particular physical port.
+            if 'channel' in port_change_event.port_dict:
```
Junchao-Mellanox, line 103: Is channel already defined in CONFIG DB PORT table? Or, are you going to add it? I don't find such field from port yang model: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-port.yang . CMIS manager is using field "lanes".

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2306,6 +2309,46 @@ def deinit(self):
...
+        The method generates a list of potential file paths, prioritizing the container platform path.
+        It then checks these paths in order to find an existing file.
+
+        Returns:
+            str: The first found platform PMON control file path.
+            None: If no file path exists.
+        """
+        platform_env_conf_path_candidates = []
+
+        platform_env_conf_path_candidates.append(
```
Junchao-Mellanox, line 2338: why not use device_info.get_path_to_platform_dir?

Junchao-Mellanox (changes requested):

Junchao-Mellanox:
Does it support dynamic port break out?

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,407 @@
...
+            # created/deleted. Thus this filed can used to determine
+            # insertion/removal event.
+            if 'type' in port_change_event.port_dict:
+                self.port_dict[lport]['type'] = port_change_event.port_dict['type']
+            self.port_dict[lport]['asic_id'] = asic_id
+        # CONFIG_DB PORT_TABLE DEL case:
+        elif port_change_event.db_name and \
+                port_change_event.db_name == 'CONFIG_DB':
+            # Only when port is removed from CONFIG, we consider this entry as deleted.
+            if lport in self.port_dict:
```
Junchao-Mellanox, line 130: do you need do tx disable here in case the module is still present? According the HLD, tx disable would save power in such case.

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,407 @@
...
+                host_tx_ready_changed = False
+                admin_status_changed = False
+                if pport < 0 or channel < 0:
+                    continue
+
+                if xcvr_type is None:
+                    # TRANSCEIVER_INFO table's 'type' is not ready, meaning xcvr is not present
+                    continue
+
+                # Procced only for 100G/40G
```
eddyk-nvidia, line 286: May be better to remove the reference to speeds since with these FF's also additional speeds can be used - 25/50/200 ...
The code below checks the form factor only

sonic-xcvrd/xcvrd/sff_mgr.py
eddyk-nvidia: So, when high-power and custom SI are supported the following sequence should be ensured for bringing a module up:

1. rx_output_disable=1
2. high-power
3. Set customer SI (per lane speed)
4. rx_output_disable=0
5. tx_disable=0

Agree ?

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,407 @@
...
+            if pport >= 0:
+                self.port_dict[lport]['index'] = pport
+            # This field comes from CONFIG_DB PORT_TABLE. This is the channel
+            # that blongs to this logical port, 0 means all channels. tx_disable
+            # API needs to know which channels to disable/enable for a
+            # particular physical port.
+            if 'channel' in port_change_event.port_dict:
+                self.port_dict[lport]['channel'] = port_change_event.port_dict['channel']
+            # This field comes from STATE_DB PORT_TABLE
+            if 'host_tx_ready' in port_change_event.port_dict:
```
tshalvi, line 106: Why not use constants for 'host_tx_ready' and 'admin_status' instead of string literals? These strings are commonly used in sff_mgr.py.

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,407 @@
...
+                if self.task_stopping_event.is_set():
+                    break
+                data = self.port_dict[lport]
+                pport = int(data.get('index', '-1'))
+                channel = int(data.get('channel', '0'))
+                xcvr_type = data.get('type', None)
+                xcvr_inserted = False
+                host_tx_ready_changed = False
+                admin_status_changed = False
+                if pport < 0 or channel < 0:
```
tshalvi, line 279: Is the condition 'channel < 0' truly necessary? I am not sure under which scenario this condition might evaluate to true. It appears that 'channel < 0' holds false in both cases, whether it is present in the 'data' dictionary or not.

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,407 @@
...
+                if 'admin_status' not in data:
+                    # Fetch admin_status from CONFIG_DB (if not present in DB,
+                    # treat it as false), and update self.port_dict
+                    data['admin_status'] = self.get_admin_status(lport, data['asic_id'])
+                    self.log_notice("{}: fetched DB and updated admin_status={} locally".format(
+                        lport, data['admin_status']))
+
+                # Check if there's a diff between current and previous 'type'
+                # It's a xcvr insertion case if TRANSCEIVER_INFO 'type' doesn't exist
+                # in previous port_dict sanpshot
```
tshalvi, line 309: Typo: 'sanpshot' --> 'snapshot'

mihirpat1 (commented):
Can you please manually cause the SffManagerTask thread to crash due to an exception and then ensure that the traceback is printed in the log followed by xcvrd restarting?

longhuan-cisco:
> Does it support dynamic port break out?

Yes, it would be compatible with DPB, because sff_mgr listens to config update, and always keeps the local port_dict data updated.

longhuan-cisco:
> Can you please manually cause the SffManagerTask thread to crash due to an exception and then ensure that the traceback is printed in the log followed by xcvrd restarting?

Yes, trace was printed in syslog followed by restart:
```
Dec 20 06:33:21.528163 sonic ERR pmon#xcvrd[25125]: Traceback (most recent call last):
Dec 20 06:33:21.528178 sonic ERR pmon#xcvrd[25125]:   File "/usr/local/lib/python3.9/dist-packages/xcvrd/sff_mgr.py", line 213, in run
Dec 20 06:33:21.528178 sonic ERR pmon#xcvrd[25125]:     self.task_worker()
Dec 20 06:33:21.528178 sonic ERR pmon#xcvrd[25125]:   File "/usr/local/lib/python3.9/dist-packages/xcvrd/sff_mgr.py", line 302, in task_worker
Dec 20 06:33:21.528202 sonic ERR pmon#xcvrd[25125]:     if not port_mapping.handle_port_update_event(
Dec 20 06:33:21.528202 sonic ERR pmon#xcvrd[25125]:   File "/usr/local/lib/python3.9/dist-packages/xcvrd/xcvrd_utilities/port_mapping.py", line 237, in handle_port_update_event
Dec 20 06:33:21.528211 sonic ERR pmon#xcvrd[25125]:     port_change_event_handler(port_change_event)
Dec 20 06:33:21.528211 sonic ERR pmon#xcvrd[25125]:   File "/usr/local/lib/python3.9/dist-packages/xcvrd/sff_mgr.py", line 150, in on_port_update_event
Dec 20 06:33:21.528221 sonic ERR pmon#xcvrd[25125]:     active_lanes = self.get_active_lanes_for_lport(lport, subport_idx,
Dec 20 06:33:21.528221 sonic ERR pmon#xcvrd[25125]:   File "/usr/local/lib/python3.9/dist-packages/xcvrd/sff_mgr.py", line 106, in get_active_lanes_for_lport
Dec 20 06:33:21.528228 sonic ERR pmon#xcvrd[25125]:     if subport_idx < 0 or subport_idx > self.DEFAULT_NUM_LANES_PER_PPORT:
Dec 20 06:33:21.528228 sonic ERR pmon#xcvrd[25125]: TypeError: '<' not supported between instances of 'str' and 'int'
Dec 20 06:33:21.528314 sonic ERR pmon#xcvrd[25125]: Xcvrd: exception found at child thread SffManagerTask due to TypeError("'<' not supported between instances of 'str' and 'int'")
Dec 20 06:33:21.528314 sonic ERR pmon#xcvrd[25125]: Exiting main loop as child thread raised exception!
Dec 20 06:33:21.580976 sonic INFO pmon#supervisord 2023-12-20 06:33:21,580 INFO exited: xcvrd (terminated by SIGKILL; not expected)
Dec 20 06:33:23.238659 sonic INFO dhclient[3081]: XMT: Solicit on eth0, interval 131410ms.
Dec 20 06:33:24.585825 sonic INFO pmon#supervisord 2023-12-20 06:33:24,585 INFO spawned: 'xcvrd' with pid 25177
Dec 20 06:33:24.700706 sonic NOTICE pmon#xcvrd[25177]: Starting up...
Dec 20 06:33:24.700706 sonic NOTICE pmon#xcvrd[25177]: XCVRD INIT: Start daemon init...
```

sonic-xcvrd/xcvrd/sff_mgr.py
longhuan-cisco: Raised issue to track SI setting: https://github.com/sonic-net/sonic-platform-daemons/issues/415

sonic-xcvrd/xcvrd/sff_mgr.py
longhuan-cisco: raised issue to track high power handling https://github.com/sonic-net/sonic-platform-daemons/issues/414

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvr_table_helper.py
prgeor, line 1: @longhuan-cisco thanks for this new file change

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2306,6 +2306,28 @@ def deinit(self):
...
+                if ENABLE_SFF_MGR_FLAG_NAME in data:
+                    enable_sff_mgr = data[ENABLE_SFF_MGR_FLAG_NAME]
+                    self.enable_sff_mgr = isinstance(enable_sff_mgr, bool) and enable_sff_mgr
+
+    def get_platform_pmon_ctrl_file_path(self):
+        """
+        Get the platform PMON control file path.
+
+        Returns:
+            str: The platform PMON control file path.
```
prgeor, line 2327: @longhuan-cisco we don't need to repeat code for parsing again. reuse the existing code hence the reason i suggested to use the same approach alreay used to skip CMIS mgr.We can revisit this later for future enhancement for both cmis and sff mgr or any new addition

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -92,6 +88,9 @@
 POWER_UNIT = 'dBm'
 BIAS_UNIT = 'mA'
 
+PLATFORM_PMON_CTRL_FILENAME = "pmon_daemon_control.json"
+ENABLE_SFF_MGR_FLAG_NAME = "enable_xcvrd_sff_mgr"
+
```
prgeor, line 93: @longhuan-cisco not required if we use cmis mgr skip approach

prgeor:
@longhuan-cisco why copper/AOC is not in the scope of SFF manager as per PR description?

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,449 @@
...
+
+
+class SffManagerTask(threading.Thread):
+
+    # CONFIG_DB port table fields:
+
+    ADMIN_STATUS = 'admin_status'
+    # This is the subport index for this logical port, starting from 1, 0 means
+    # all lanes are taken.
+    SUBPORT = 'subport'
```
prgeor, line 31: @longhuan-cisco what is subport?

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,449 @@
...
+    def log_notice(self, message):
+        self.helper_logger.log_notice("SFF: {}".format(message))
+
+    def log_warning(self, message):
+        self.helper_logger.log_warning("SFF: {}".format(message))
+
+    def log_error(self, message):
+        self.helper_logger.log_error("SFF: {}".format(message))
+
+    def get_active_lanes_for_lport(self, lport, subport_idx, num_lanes_per_lport, num_lanes_per_pport):
```
prgeor, line 92: @longhuan-cisco what are active lanes? is it the lanes mapped to the logical port? if so, can't we dervice from config DB's port table?

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,449 @@
...
+
+        if port_change_event.event_type == port_change_event.PORT_SET:
+            if lport not in self.port_dict:
+                self.port_dict[lport] = {}
+            if pport >= 0:
+                self.port_dict[lport]['index'] = pport
+
+            if self.SUBPORT in port_change_event.port_dict and \
+                self.LANES_LIST in port_change_event.port_dict:
+                subport_idx = int(port_change_event.port_dict[self.SUBPORT])
```
prgeor, line 152: @longhuan-cisco where is this defined: subport?

sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
prgeor, line 1: @longhuan-cisco enhancement or bug fix to this file should be taken separate PR.

prgeor (commented):
@

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,462 @@
...
+        cfg_port_tbl = self.xcvr_table_helper.get_cfg_port_tbl(asic_index)
+
+        found, value = cfg_port_tbl.hget(lport, self.ADMIN_STATUS)
+        if found:
+            admin_status = value
+        return admin_status
+
+    def run(self):
+        if self.platform_chassis is None:
+            self.log_notice("Platform chassis is not available, stopping...")
```
mihirpat1, line 211:
```suggestion
            self.log_error("Platform chassis is not available, stopping...")
```

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,407 @@
...
+            # API needs to know which channels to disable/enable for a
+            # particular physical port.
+            if 'channel' in port_change_event.port_dict:
+                self.port_dict[lport]['channel'] = port_change_event.port_dict['channel']
+            # This field comes from STATE_DB PORT_TABLE
+            if 'host_tx_ready' in port_change_event.port_dict:
+                self.port_dict[lport]['host_tx_ready'] = \
+                        port_change_event.port_dict['host_tx_ready']
+            # This field comes from CONFIG_DB PORT_TABLE
+            if 'admin_status' in port_change_event.port_dict and \
```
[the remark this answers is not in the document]
mihirpat1: @longhuan-cisco, I think removing the filter on STATE_DB would actually mean that SFF manager is interested in all the fields of STATE_DB and hence, that would be a calculated move.
Agreed that the check is more robust but I think from code perspective, it will be a redundant check since the FILTER option of the STATE_DB is primarily meant to facilitate this.
Also, this might cause unnecessary confusion since other DB field options (such as host_tx_ready) doesn't have DB relevant checks and rather are assumed to be source of truth.I think the case of removing the filter on STATE_DB would actually mean that SFF manager is interested in all the fields of STATE_DB and hence, that would be a calculated move.
Agreed that the check is more robust but I think from code perspective, it will be a redundant check since the FILTER option of the STATE_DB is primarily meant to facilitate this.
Also, this might cause unnecessary confusion since other DB field options (such as host_tx_ready) don't have DB relevant checks and rather are assumed to be source of truth.

We can also add a FILTER for CONFIG_DB so that only the relevant fields are processed by SFF manager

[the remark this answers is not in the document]
longhuan-cisco: Agreed. removed the checks.
And added one comment at the FILTER definition to keep people cautious, since not all people are aware that `admin_status` can also come from STATE_DB.

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,449 @@
...
+            subport_idx (int): Subport index, starting from 1. 0 means all lanes are taken.
+            num_lanes_per_lport (int): Number of lanes per logical port.
+            num_lanes_per_pport (int): Number of lanes per physical port.
+
+        Returns:
+            list: A list of boolean values, where True means the corresponding
+                  lane is active.
+        """
+        if subport_idx < 0 or subport_idx > num_lanes_per_pport // num_lanes_per_lport:
+            self.log_error(
```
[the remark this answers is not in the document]
longhuan-cisco: changed to format

sonic-xcvrd/xcvrd/sff_mgr.py
longhuan-cisco: changed to error

sonic-xcvrd/xcvrd/sff_mgr.py
Junchao-Mellanox: Hi, is `subport` only used for dynamic port breakout?  I don't see dynamic port breakout handle this field, so, I am not sure that the field is fully supported by sonic.  Or is `subport` used for locating the lanes mapped to logical port? In that case, I think field `lanes` can do that.  So, I am not fully understand why we need subport here. Could you please kindly elaborate?

longhuan-cisco: 'subport' is used for breakout situation,`subport` field determines which lanes belongs to which subport, based on which tx_disable/enable API (which takes lane mask as argument) can be called per subport.
The `lanes` field in CONFIG_DB are just serdes lane numbers (e.g. 2304). Those are just some big HW level number, and it can't derive the lane mask.

today's cmis_mgr is already relying on this `subport`. today's DPB HLD has a section for `subport` also.
For now, it's supported only for static breakout config via port_config.ini.

example:
Assuming, we have a physical port with serdes `lanes` (2304,2305,2306,2307), then we do a 4x breakout as below,
  - `subport` 1 (serdes `lanes`=2304): 1st lane   (lane mask 0b0001 for tx_disable/enable API)
  - `subport` 2 (serdes `lanes`=2305): 2nd lane (lane mask 0b0010)
  - `subport` 3 (serdes `lanes`=2306): 3rd lane  (lane mask 0b0100)
  - `subport` 4 (serdes `lanes`=2307): 4th lane.   (lane mask 0b1000)

Junchao-Mellanox: Thanks for the explanation. So, dynamic port breakout is not fully supported, but static port breakout is supported by this field. @keboliu , @eddyk-nvidia for awareness.

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,449 @@
...
+    # TRANSCEIVER_INFO has the same life cycle as a transceiver, if transceiver
+    # is inserted/removed, TRANSCEIVER_INFO is also created/deleted.
+    XCVR_TYPE = 'type'
+
+
+    # STATE_DB PORT_TABLE fields:
+
+    HOST_TX_READY = 'host_tx_ready'
+
+    # Default number of lanes per physical port for QSFP28/QSFP+ transceiver
```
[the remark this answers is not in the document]
prgeor: @longhuan-cisco why do we need to skip QSFP+

sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
```diff
@@ -173,42 +189,54 @@ def handle_port_update_event(sel, asic_context, stop_event, logger, port_change_
...
+            if key in PortChangeEvent.thread_local_data.PORT_EVENT:
+                # Compare current event with last event on this key, to see if
+                # there's really a change on the contents.
+                diff = set(fvp.items()) - set(PortChangeEvent.thread_local_data.PORT_EVENT[key].items())
+                # Ignore duplicate events
+                if not diff:
+                   PortChangeEvent.thread_local_data.PORT_EVENT[key] = fvp
+                   continue
+            # Update the latest event to the cache
+            PortChangeEvent.thread_local_data.PORT_EVENT[key] = fvp
```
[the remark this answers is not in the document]
prgeor: @longhuan-cisco separate PR for this should work for CMIS mgr as well, so I don't see any reason it cannot be a separate PR. It will less of a work for release mgr to revert in case of smaller change to be reverted that a feature + enhancement/fix

sonic-xcvrd/xcvrd/xcvrd.py
prgeor: @longhuan-cisco to keep the code clean and to NOT repeat the parsing code for every newly introduced parameter to xcvrd daemon, suggest to follow the same approach as skip_cmis_mgr. It just unnecessary to duplicate code for every new parameter introduced here and future

tshalvi:
Any updates on this PR? Haven't seen recent activity and would appreciate an ETA if possible.

abdosi:
@longhuan-cisco : Can you provide latest update on this.

sonic-xcvrd/xcvrd/sff_mgr.py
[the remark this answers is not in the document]
longhuan-cisco: As the code`if not (xcvr_type.startswith('QSFP28') or xcvr_type.startswith('QSFP+'))`, QSFP+ was not skipped

sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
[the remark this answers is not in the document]
longhuan-cisco: sure, created separate PR https://github.com/sonic-net/sonic-platform-daemons/pull/430

sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
longhuan-cisco: separate PR was raised for it.

sonic-xcvrd/xcvrd/xcvrd.py
longhuan-cisco: raised another PR to use the old approach:  https://github.com/sonic-net/sonic-buildimage/pull/18054

sonic-xcvrd/xcvrd/xcvrd.py
longhuan-cisco: updated to use the old approach of going via docker-pmon.supervisord.conf, corresponding PR was raised https://github.com/sonic-net/sonic-buildimage/pull/18054

longhuan-cisco:
> @longhuan-cisco why copper/AOC is not in the scope of SFF manager as per PR description?

AFAIK, those doesn't have ability of doing tx_disable/enable

longhuan-cisco:
> Any updates on this PR? Haven't seen recent activity and would appreciate an ETA if possible.

> @longhuan-cisco : Can you provide latest update on this.

Updated the diff and raised two other PR's based on latest comments from Prince's.

the other PRs:
https://github.com/sonic-net/sonic-platform-daemons/pull/430
https://github.com/sonic-net/sonic-buildimage/pull/18054

sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
longhuan-cisco: raised separate PR.

rlhui:
please resolve conflicts

longhuan-cisco:
> please resolve conflicts

conflicts resolved.

eddyk-nvidia (approved):

mihirpat1 (approved):

tshalvi (approved):

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -0,0 +1,478 @@
...
+                # corresponding lane belongs to this logical port.
+                active_lanes = data.get('active_lanes', None)
+                xcvr_type = data.get(self.XCVR_TYPE, None)
+                xcvr_inserted = False
+                host_tx_ready_changed = False
+                admin_status_changed = False
+                if pport < 0 or lanes_list is None:
+                    continue
+
+                if xcvr_type is None:
```
prgeor, line 342: @longhuan-cisco please use [_wrapper_get_presence()](https://github.com/sonic-net/sonic-platform-daemons/blob/master/sonic-xcvrd/xcvrd/xcvrd.py#L229). There are cases where SFF manager would be busy doing stuff and may miss the event to process Xcvr removal event.

prgeor: @longhuan-cisco or better use  sfp.get_presence()

longhuan-cisco: @prgeor  It's already checked at below:
https://github.com/sonic-net/sonic-platform-daemons/blob/83f7bc3b2dfbbc359380f3b7fbc64127000a1c22/sonic-xcvrd/xcvrd/sff_mgr.py#L401-L406

judyjoseph:
@prgeor Please merge if the review is completed.

prgeor:
@longhuan-cisco why AOC is outside scope?

`Scope of sff_mgr: 100G/40G optics (copper/aoc not in the scope)`

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvr_table_helper.py
```diff
@@ -0,0 +1,64 @@
+try:
```
prgeor, line 1: @longhuan-cisco thanks for moving to this new file

sonic-xcvrd/xcvrd/sff_mgr.py
prgeor: @longhuan-cisco can we remove this redundant check here then because it will be checked further down? The comment says:-

`# TRANSCEIVER_INFO table's XCVR_TYPE is not ready, meaning xcvr is not present`

longhuan-cisco:
> @longhuan-cisco why AOC is outside scope?
> 
> `Scope of sff_mgr: 100G/40G optics (copper/aoc not in the scope)`

@prgeor This was answered at https://github.com/sonic-net/sonic-platform-daemons/pull/383#issuecomment-1931141283
> AFAIK, those doesn't have ability of doing tx_disable/enable

sonic-xcvrd/xcvrd/sff_mgr.py
longhuan-cisco: `if xcvr_type is None` is not redundant, because: it is just a simple checking on value already obtained from DB, much more efficient, and in most of time (probably at least >95% of sfp deletion cases, sff_mgr gets to know it via TRANSCEIVER_INFO entry deletion) it can skip the processing earlier than checking of `sfp.get_presence` at later point.

prgeor (approved):

prgeor:
@longhuan-cisco there is some conflict, can you rebase?

longhuan-cisco:
> @longhuan-cisco there is some conflict, can you rebase?

@prgeor I didn't see conflicts, it just said branch "out-of-date". Anyway I updated it to latest master. please check.

PR #353, opened 2023-05-03
Use tx_disable_channel with media_lanes_mask

rajann:
#### Description
    
    Invoke tx_disable_channel() api using the correct media_lanes_mask in case of subport shut/no shut operations

     (i.e) existing tx_disable_channel( ) api is being used with host_lanes_mask as of now.  
            This is not aligning with the subport concept where media_lane start wont match with host_lanes start.

    As part of the fix, introduced an api to retrieve media_lanes_mask and use it as part of the tx_disable_channel() api

#### How Has This Been Tested?

   Using 4x100G and 2x100G breakout modules 

-   create 4 and 2 subports on 4x100G and 2x100G modules respectively.
-   Check whether all the subports are operationally up 
-   Verify the upper page 10 for outputdisable values and page 11 for the data path state
-   Shut each subports one at a time and verify the operational status of every subports in this module
-   Unshut each subport one at a time and verify the operational status of every subports in this module
-   At each step verify the values in page10h and page 11h
-   Verified the subports with traffic

PR #386, opened 2023-07-13
Fix index out of range in the error log of invalid media lane mask re…

MichaelWangSmci:
…ceived

In the error log format string, there are 6 placeholders but only 5 values are provided.

#### Description

Add lport at the beginning of the values and remove the redundant lport in the format.
So there are 5 placeholders and 5 values.

#### Motivation and Context

There will be an IndexError exception like below.
```
Exception occured at CmisManagerTask thread due to IndexError('Replacement index 5 out of range for positional args tuple')
  File "/usr/local/lib/python3.9/dist-packages/xcvrd/xcvrd.py", line 1878, in run
    self.task_worker()
  File "/usr/local/lib/python3.9/dist-packages/xcvrd/xcvrd.py", line 1672, in task_worker
    self.log_error("{}: Invalid media lane mask received - media_lane_count {} "
IndexError: Replacement index 5 out of range for positional args tuple
Xcvrd: exception found at child thread CmisManagerTask due to IndexError('Replacement index 5 out of range for positional args tuple')
Exiting main loop as child thread raised exception!
```

Fake the error to check if the error log can be printed as expected.
```
CMIS: Ethernet162: Invalid media lane mask received - media_lane_count 1 media_lane_assignment_options 85 subport 0 appl 2!
```

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1530,8 +1530,8 @@ class CmisManagerTask(threading.Thread):
                                                                         appl, lport, subport)
                         if self.port_dict[lport]['media_lanes_mask'] <= 0:
                             self.log_error("{}: Invalid media lane mask received - media_lane_count {} "
-                                            "media_lane_assignment_options {} lport{} subport {}"
-                                            " appl {}!".format(media_lane_count,media_lane_assignment_options,lport,subport,appl))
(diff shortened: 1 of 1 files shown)


MichaelWangSmci:
The pull request introduced this is #353.
And it is included in master, 202211 and 202205.

prgeor (approved):

MichaelWangSmci:
Is it possible this pull request can be also merged into 202305, 202211 and 202205?

Issue #384, opened 2023-06-29
thermalctld adds nonpresent PSUs sensors to the TEMPERATURE_INFO table


cytsao1:
For any nonpresent PSU, thermalctld will add its sensors to the redis database, with all values N/A.
thermalctld should check for PSU/Fan presence before adding their sensors to the db.

show platform temperature
```
            PSU0.0 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:04
            PSU0.1 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:04
            PSU0.2 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:04
            PSU1.0 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:04
            PSU1.1 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:04
            PSU1.2 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:04
            PSU2.0 Outlet_Temp         46.019       97.0      -5.0           102.0          -10.0      False  20230214 14:05:05
            PSU2.1 Outlet_Temp         46.019       97.0      -5.0           102.0          -10.0      False  20230214 14:05:05
            PSU2.2 Outlet_Temp         46.019       97.0      -5.0           102.0          -10.0      False  20230214 14:05:05
            PSU3.0 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:05
            PSU3.1 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:05
            PSU3.2 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:05
            PSU4.0 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:05
            PSU4.1 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:05
            PSU4.2 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:05
            PSU5.0 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:05
            PSU5.1 Outlet_Temp            N/A        N/A       N/A             N/A            N/A      False  20230214 14:05:05
[...]

PR #387, opened 2023-07-13
[thermalctld] Add check for psu presence to thermalctld when updating PSU thermals

cytsao1:
#### Description

issue https://github.com/sonic-net/sonic-platform-daemons/issues/384

For any nonpresent PSU, thermalctld will add its sensors to the redis database, with all values N/A.
thermalctld should check for PSU presence before adding their sensors to the db.

#### Motivation and Context

The empty non-present sensor entries are not necessary to be tracked, and causes test case failures as values are expected to be retrieved from the sensors.

With the update to thermalctld, ran relevant CLI and checked the db.

show plat inv
```
Power Supplies
    psu0                8800-HV-TRAY    1.0             FOC2417N4X0     Cisco 8800 Power Tray for AC/HVAC/HVDC Power Supply
        psu0.0 -- not present
        psu0.1 -- not present
        psu0.2 -- not present
    psu1                8800-HV-TRAY    1.0             FOC2417N4VJ     Cisco 8800 Power Tray for AC/HVAC/HVDC Power Supply
        psu1.0 -- not present
        psu1.1 -- not present
        psu1.2 -- not present
    psu2                8800-HV-TRAY    1.0             FOC2417N4U6     Cisco 8800 Power Tray for AC/HVAC/HVDC Power Supply
        psu2.0          PSU6.3KW-20A-HV 1.0             DTM243501T9     6.3KW AC/HVAC/HVDC Power Supply-20A
        psu2.1          PSU6.3KW-20A-HV 1.0             DTM243501UV     6.3KW AC/HVAC/HVDC Power Supply-20A
        psu2.2          PSU6.3KW-20A-HV 1.0             DTM2435022V     6.3KW AC/HVAC/HVDC Power Supply-20A
    psu3 -- not present
    psu4 -- not present
    psu5 -- not present
```
show plat temp (no empty entries)
```
            PSU2.0 Outlet_Temp         48.992         97        -5             102            -10      False  20230228 13:58:20
            PSU2.1 Outlet_Temp         48.992         97        -5             102            -10      False  20230228 13:58:21
            PSU2.2 Outlet_Temp         48.992         97        -5             102            -10      False  20230228 13:58:21
```
db entries (no non present PSUs)
```
root@sonic:/home/cisco# redis-cli -n 6
127.0.0.1:6379[6]> keys TEMPERATURE_INFO|PSU*
1) "TEMPERATURE_INFO|PSU2.1 Outlet_Temp"
2) "TEMPERATURE_INFO|PSU2.2 Outlet_Temp"
3) "TEMPERATURE_INFO|PSU2.0 Outlet_Temp"
127.0.0.1:6379[6]>
```

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -592,6 +592,8 @@ class TemperatureUpdater(logger.Logger):
             self._refresh_temperature_status(CHASSIS_INFO_KEY, thermal, index)
 
         for psu_index, psu in enumerate(self.chassis.get_all_psus()):
+            if not psu.get_presence():
+                continue
             parent_name = 'PSU {}'.format(psu_index + 1)
             for thermal_index, thermal in enumerate(psu.get_all_thermals()):
                 if self.task_stopping_event.is_set():


sonic-thermalctld/scripts/thermalctld
```diff
@@ -592,6 +592,8 @@ class TemperatureUpdater(logger.Logger):
             self._refresh_temperature_status(CHASSIS_INFO_KEY, thermal, index)
 
         for psu_index, psu in enumerate(self.chassis.get_all_psus()):
+            if not psu.get_presence():
```
Junchao-Mellanox, line 595: I am not sure that we should check presence here. "N/A" is good enough to tell user that the PSU thermal is not available at the moment, no matter it is because of PSU absence or hardware error. Maybe you should fix your test expectation. But if you insit the change, please consider following items:
1. Do we need check psu.get_power_good_status here? If PSU is present but not powered on, thermal sensor should not be available.
2. Do we need remove previous DB entry here?
3. For `show platform fan` and `show platform psu`, non presence PSU will be displayed, should we align it with this design?
4. For existing sonic-mgmt test, can we guarantee no test is affected?
5. Unit test is needed to cover the new change.

Junchao-Mellanox (changes requested):

Junchao-Mellanox:
Hi @cytsao1 and @prgeor , I personally prefer to keeping existing code and requesting to change their test logic. Let me know if you have concern.

sonic-thermalctld/scripts/thermalctld
prgeor: @cytsao1 @Junchao-Mellanox how about get_all_psus() return only those PSUs which are present? In that case, "N/A" would indicate some issue with physical present PSU?

spilkey-cisco: Only returning present PSUs from `get_all_psus` would cause issues with inconsistent indexing throughout the code; when `enumerate` is used, numbering will always be sequential from `0`, and inserting a new PSU may cause numbering of existing PSUs to change.

gechiang:
> Hi @cytsao1 and @prgeor , I personally prefer to keeping existing code and requesting to change their test logic. Let me know if you have concern.

@Junchao-Mellanox , I still don't get why you need to keep sensors info for PSU that are not present in redis-db?  what is the purpose of it?  It has absolutely no useful info so why are we keeping this?  I think not including it is the correct way of handling this.

gechiang (approved):

Junchao-Mellanox:
> > Hi @cytsao1 and @prgeor , I personally prefer to keeping existing code and requesting to change their test logic. Let me know if you have concern.
> 
> @Junchao-Mellanox , I still don't get why you need to keep sensors info for PSU that are not present in redis-db? what is the purpose of it? It has absolutely no useful info so why are we keeping this? I think not including it is the correct way of handling this.

It tells user that there are 2 PSU sensors, and 1 is not available, I don't see an issue here. To my understanding, this change will cause sonic-mgmt `test_show_platform_temperature_mocked` case failure. So, why should we choose a solution that would affect other vendors with no apparent benefit? How about change your own test case?

gechiang:
> > > Hi @cytsao1 and @prgeor , I personally prefer to keeping existing code and requesting to change their test logic. Let me know if you have concern.
> > 
> > 
> > @Junchao-Mellanox , I still don't get why you need to keep sensors info for PSU that are not present in redis-db? what is the purpose of it? It has absolutely no useful info so why are we keeping this? I think not including it is the correct way of handling this.
> 
> It tells user that there are 2 PSU sensors, and 1 is not available, I don't see an issue here. To my understanding, this change will cause sonic-mgmt `test_show_platform_temperature_mocked` case failure. So, why should we choose a solution that would affect other vendors with no obviously benefit? How about change your own test case? How about change your platform API psu.get_all_thermals to return empty list while the PSU is not present?

Is N/A in redis-db indicate Not Available Or it is bad sensor data?  how can one tell? Also, for your example, there is only 1 PSU sensor.  there is really no "2 PSU sensors" since one PSU is not even present??  How do you differentiate the case where the PSU is really present but not providing proper sensor data which resulted to N/A?

Junchao-Mellanox:
> > > > Hi @cytsao1 and @prgeor , I personally prefer to keeping existing code and requesting to change their test logic. Let me know if you have concern.
> > > 
> > > 
> > > @Junchao-Mellanox , I still don't get why you need to keep sensors info for PSU that are not present in redis-db? what is the purpose of it? It has absolutely no useful info so why are we keeping this? I think not including it is the correct way of handling this.
> > 
> > 
> > It tells user that there are 2 PSU sensors, and 1 is not available, I don't see an issue here. To my understanding, this change will cause sonic-mgmt `test_show_platform_temperature_mocked` case failure. So, why should we choose a solution that would affect other vendors with no obviously benefit? How about change your own test case? How about change your platform API psu.get_all_thermals to return empty list while the PSU is not present?
> 
> Is N/A in redis-db indicate Not Available Or it is bad sensor data? how can one tell? Also, for your example, there is only 1 PSU sensor. there is really no "2 PSU sensors" since one PSU is not even present?? How do you differentiate the case where the PSU is really present but not providing proper sensor data which resulted to N/A?

We don't have mechanism to check bad sensor data. So, with or without this change, we cannot tell it. For PSU presence, you can check `show platform psustatus`. Like I said, it will cause sonic-mgmt test case failure. The change should not break exist test case, agree? Why don't you change the platform API psu.get_all_thermals?

spilkey-cisco:
There will be a new PR raised to hopefully address this more completely. One issue I haven't seen mentioned here is that without some sort of presence check, sensors from PSUs that were previously present will not be removed from redis when the PSU is removed, and stale data from those sensors will remain in redis (this also affects snmp).

gechiang:
@spilkey-cisco , 
I am not sure what is the final decision and what we are still waiting for on this PR?  Based on your last comment reply, are you making additional changes? raising a new PR and abandoning this PR?  BTW, you need to remerge as this PR is out-of-sync with base branch already...

spilkey-cisco:
> @spilkey-cisco , I am not sure what is the final decision and what we are still waiting for on this PR? Based on your last comment reply, are you making additional changes? raising a new PR and abandoning this PR? BTW, you need to remerge as this PR is out-of-sync with base branch already...

New PR will be created, this PR will not be merged and can be closed.

prgeor:
closing @spilkey-cisco @cytsao1

PR #400, opened 2023-10-12
Add Port SI Configuration Per Speed 

tshalvi:
Description
I have updated the lookup in media_settings.json. With the updated parser, media_settings.json now supports different SI values for various lane speeds. Previously, get_media_settings_key() returned two types of keys: vendor_key and media_key. With these code changes, three keys are now returned: vendor_key, media_key, and lane_speed_key. The parser, implemented in get_media_settings_value(), intelligently utilizes the lane_speed_key only for JSON files that support per-speed port SI values. In cases where the JSON does not support this feature, the parser's functionality remains unchanged.

Motivation and Context
To achieve optimal configuration of their SerDes, some vendors wish to have the option to configure SerDes SI based on lane speeds. Therefore, we want to enhance the current ASIC configuration to provide support for configurations that take into account lane speed.

How Has This Been Tested?
I was running the existing unit tests to test these changes, and updated them if needed. They all passed.
Soon I will add new unit test to cover all my changes.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -29,9 +29,7 @@ try:
     from .xcvrd_utilities import port_mapping
     from .xcvrd_utilities import optics_si_parser
     
-
     from sonic_platform_base.sonic_xcvr.api.public.c_cmis import CmisApi
-
 
 except ImportError as e:
     raise ImportError(str(e) + " - required module not found")
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -132,23 +132,15 @@ def get_cmis_application_desired(api, host_lane_count, speed):
     if speed == 0 or host_lane_count == 0:
         return None
 
-    appl_code = None
-    app_found = False
-    if is_cmis_api(api):
-        appl_dict = api.get_application_advertisement()
-        for index in appl_dict.keys():
-            app_info = appl_dict[index]
-            if app_info.get('host_lane_count') != host_lane_count:
-                continue
-            if get_interface_speed(app_info.get('host_electrical_interface_id')) != speed:
-                continue
-            appl_code = index
-            app_found = True
-            break
+    if not is_cmis_api(api):
+        return None
 
-        if app_found:
-            return (appl_code & 0xf)
-    
+    appl_dict = api.get_application_advertisement()
+    for index, app_info in appl_dict.items():
+        if (app_info.get('host_lane_count') == host_lane_count and
+        get_interface_speed(app_info.get('host_electrical_interface_id')) == speed):
+            return (index & 0xf)
+
     return None
 
 
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -109,7 +109,6 @@ helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
 #
 
 
-#TODO: This function is duplicated from the CMIS thread. Need to have one merged function out of the CMIS context
 def get_cmis_application_desired(api, host_lane_count, speed):
     """
     Get the CMIS application code that matches the specified host side configurations
@@ -150,7 +149,6 @@ def get_cmis_application_desired(api, host_lane_count, speed):
     return None
 
 
-#TODO: This function is duplicated from the CMIS thread. Need to have one merged function out of the CMIS context
 def get_interface_speed(ifname):
     """
     Get the port speed from the host interface name
@@ -752,7 +750,6 @@ def get_media_settings_value(physical_port, key):
     return {}
 
 
-#TODO: update the implementation of this method after getting answers from Prince (probably replacing CONFIG_DB with APP_DB)
 def get_speed_and_lane_count(port, cfg_port_tbl):
     port_speed, lane_count = -1, -1
     found, port_info = cfg_port_tbl.get(port)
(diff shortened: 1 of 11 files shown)


sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
```diff
@@ -0,0 +1,317 @@
...
+g_dict = {}
+
+LANE_SPEED_KEY_PREFIX = "speed:"
+VENDOR_KEY = 'vendor_key'
+MEDIA_KEY = 'media_key'
+LANE_SPEED_KEY = 'lane_speed_key'
+SYSLOG_IDENTIFIER = "xcvrd"
+helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
+
+
```
prgeor, line 18: @tshalvi add a comment, this parser parses the NPU or ASIC side SerDes custom SI settings.

prgeor:
@tshalvi please fix the code coverage

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -103,6 +107,75 @@
 # Helper functions =============================================================
 #
 
+
+def is_cmis_api(api):
+   return type(api) == CmisApi
```
mihirpat1, line 112: Wondering if we should include CCmisApi here?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1625,7 +1420,13 @@ def task_worker(self):
                             # Apply module SI settings if applicable
                             lane_speed = int(speed/1000)//host_lane_count
                             optics_si_dict = optics_si_parser.fetch_optics_si_setting(pport, lane_speed, sfp)
-
+                            
+                            helper_logger.log_debug("Read SI parameters for port {} from optics_si_settings.json vendor file:".format(lport))
```
mihirpat1, line 1424: Please define log_debug (similar to log_error and log_notice) in CmisManagerTask and use it in CMIS related functions.

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
```diff
@@ -0,0 +1,325 @@
...
+        if physical_port not in transceiver_dict:
+            helper_logger.log_error("Media {} eeprom not populated in transceiver dict".format(physical_port))
+            continue
+
+        port_name = xcvrd.get_physical_port_name(logical_port_name,
+                                           ganged_member_num, ganged_port)
+        
+        ganged_member_num += 1
+        key = get_media_settings_key(physical_port, transceiver_dict, port_speed, lane_count)
+        helper_logger.log_debug("Retrieving media settings for port {} using the following keys: {}".format(logical_port_name, key))
```
prgeor, line 303: @tshalvi please also log the port speed and lane count

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
```diff
@@ -0,0 +1,325 @@
...
+        media_dict = get_media_settings_value(physical_port, key)
+
+        if len(media_dict) == 0:
+            helper_logger.log_error("Error in obtaining media setting for {}".format(logical_port_name))
+            return
+
+        fvs = swsscommon.FieldValuePairs(len(media_dict))
+
+        index = 0
+        helper_logger.log_debug("Storing in Application DB the following media settings for port {}:".format(logical_port_name))
```
prgeor, line 313: @tshalvi Suggest to log as "Publishing ASIC-side SI setting for port"

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -103,6 +107,75 @@
...
+
+    Returns:
+        Integer, the transceiver-specific application code
+    """
+
+    if speed == 0 or host_lane_count == 0:
+        return None
+
+    if not is_cmis_api(api):
+        return None
```
prgeor, line 135: @tshalvi the caller for this is CMIS task manager which calls this API only for CMIS compliant transceivers. So we don't need this check again

tshalvi: The is_cmis_api() method is called not only within the CmisManagerTask, but is also invoked from the get_lane_speed_key() method in the media_settings_parser during the notify_media_setings() process (within SfpStateUpdateTask). When it's called from this method, there's no guarantee that the module being used is a CMIS module.

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
```diff
@@ -0,0 +1,325 @@
...
+        g_dict = json.load(media_file)
+
+
+def media_settings_present():
+    if g_dict:
+        return True
+    return False
+
+
+def get_lane_speed_key(physical_port, port_speed, lane_count):
```
prgeor, line 42: @tshalvi why not calculate the lane speed as `port_speed / lane_count` ?

sonic-xcvrd/xcvrd/xcvrd.py
tshalvi: This should be done as part of a separate PR because we don't have the necessary modules to test it.

Junchao-Mellanox (approved):

prgeor (approved):

prgeor:
@tshalvi @liat-grozovik please see cherry pick conflict so may need separate pr for 202311

keboliu:
@yxieca @liushilongbuaa  would you please remove the tag "request for 202311 branch" as it is already included in 202311?

PR #401, opened 2023-10-17
Add dynamic sensor logic for fixed and psu presence/state checking in thermalctld

gregoryboudreau:
…hecking for PSUs

#### Description

Extends the existing dynamic sensor logic of modular chassis to fixed and adds a check for PSU presence/status in thermalctld
#### Motivation and Context

In current form, if a PSU or other device with sensors is removed from the device or disabled during runtime on a fixed chassis, they will remain in the temperature sensor DB which should not be correct. This change brings the existing logic checks that modular chassis' have to fixed platform as well as checks for PSU presence/status to account for scenario where PSU may be disabled during runtime.

On both fixed and modular platforms (Cisco 8800RP and Cisco 8101_32FH), I confirmed presence of all thermal sensors. Then disable the PSU and confirmed its associated sensors were removed from the DB while it was either inactive and present or not present. Reenabled PSU and confirmed its associated sensors were updated back into the DB.

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -520,10 +520,10 @@ class TemperatureUpdater(logger.Logger):
         self.table = swsscommon.Table(state_db, TemperatureUpdater.TEMPER_INFO_TABLE_NAME)
         self.phy_entity_table = swsscommon.Table(state_db, PHYSICAL_ENTITY_INFO_TABLE)
         self.chassis_table = None
+        self.all_thermals = set()
 
         self.is_chassis_system = chassis.is_modular_chassis()
         if self.is_chassis_system:
-            self.module_thermals = set()
             my_slot = try_get(chassis.get_my_slot, INVALID_SLOT)
             if my_slot != INVALID_SLOT:
                 try:
@@ -572,13 +572,23 @@ class TemperatureUpdater(logger.Logger):
 
             self._refresh_temperature_status(CHASSIS_INFO_KEY, thermal, index)
 
+        available_thermals = set()
+        for index, thermal in enumerate(self.chassis.get_all_thermals()):
+            if self.task_stopping_event.is_set():
+                return
+
+            available_thermals.add((thermal, CHASSIS_INFO_KEY, index))
+            self._refresh_temperature_status(CHASSIS_INFO_KEY, thermal, index)
+
         for psu_index, psu in enumerate(self.chassis.get_all_psus()):
             parent_name = 'PSU {}'.format(psu_index + 1)
-            for thermal_index, thermal in enumerate(psu.get_all_thermals()):
-                if self.task_stopping_event.is_set():
-                    return
+            if psu.get_presence() and psu.get_powergood_status():
+                for thermal_index, thermal in enumerate(psu.get_all_thermals()):
+                    if self.task_stopping_event.is_set():
+                        return
 
-                self._refresh_temperature_status(parent_name, thermal, thermal_index)
+                    available_thermals.add((thermal, parent_name, thermal_index))
+                    self._refresh_temperature_status(parent_name, thermal, thermal_index)
 
         for sfp_index, sfp in enumerate(self.chassis.get_all_sfps()):
             parent_name = 'SFP {}'.format(sfp_index + 1)
@@ -586,10 +596,9 @@ class TemperatureUpdater(logger.Logger):
                 if self.task_stopping_event.is_set():
                     return
 
+                available_thermals.add((thermal, parent_name, thermal_index))
                 self._refresh_temperature_status(parent_name, thermal, thermal_index)
-
         if self.is_chassis_system:
-            available_thermals = set()
             for module_index, module in enumerate(self.chassis.get_all_modules()):
                 module_name = try_get(module.get_name, 'Module {}'.format(module_index + 1))
 
@@ -608,21 +617,19 @@ class TemperatureUpdater(logger.Logger):
 
                         available_thermals.add((thermal, sfp_name, thermal_index))
                         self._refresh_temperature_status(sfp_name, thermal, thermal_index)
-
+                        
                 for psu_index, psu in enumerate(module.get_all_psus()):
-                    psu_name = '{} PSU {}'.format(module_name, psu_index + 1)
-                    for thermal_index, thermal in enumerate(psu.get_all_thermals()):
-                        if self.task_stopping_event.is_set():
-                            return
+                    if psu.get_presence() and psu.get_powergood_status():
+                        for thermal_index, thermal in enumerate(psu.get_all_thermals()):
+                            if self.task_stopping_event.is_set():
+                                return
+                            available_thermals.add((thermal, psu_name, thermal_index))
+                            self._refresh_temperature_status(psu_name, thermal, thermal_index)
 
-                        available_thermals.add((thermal, psu_name, thermal_index))
-                        self._refresh_temperature_status(psu_name, thermal, thermal_index)
-
-
-            thermals_to_remove = self.module_thermals - available_thermals
-            self.module_thermals = available_thermals
-            for thermal, parent_name, thermal_index in thermals_to_remove:
-                self._remove_thermal_from_db(thermal, parent_name, thermal_index)
+        thermals_to_remove = self.all_thermals - available_thermals
+        self.all_thermals = available_thermals
+        for thermal, parent_name, thermal_index in thermals_to_remove:
+            self._remove_thermal_from_db(thermal, parent_name, thermal_index)
 
         self.log_debug("End temperature updating")
 


gregoryboudreau:
This PR is being introduced as was described in https://github.com/sonic-net/sonic-platform-daemons/pull/387

sonic-thermalctld/scripts/thermalctld
```diff
@@ -608,21 +612,20 @@ class TemperatureUpdater(logger.Logger):
 
                         available_thermals.add((thermal, sfp_name, thermal_index))
                         self._refresh_temperature_status(sfp_name, thermal, thermal_index)
-
+
```
spilkey-cisco, line 615: Please revert whitespace change here.

bmridul:
@gechiang , Can u review this PR.

sonic-thermalctld/scripts/thermalctld
gregoryboudreau: Updated.

gechiang (approved):
LGTM.

gregoryboudreau:
This definitely needs to go in 202305 and should also probably go into 202205 as well. Fixes bug in both where stale thermal sensors will be left in temp db and reported on until device power cycle as well as undefined behavior on fixed platforms where dependent on the sensors, a device that is not on but present can still report sensor information which is not behavior we would expect.

gechiang:
@Junchao-Mellanox , please help review this PR. Thanks!

spilkey-cisco (approved):

keboliu:
If my understanding is correct, this PR will change the output of the "show platform psustatus" in some cases(when some PSU is not present), some sonic-mgmt test cases is relying on analyzing the CLI output. I would like suggest providing a test result on those cases against this PR, to make sure no case is broken.  If some test case is broken, as a full solution,  PRs shall also raised to fix those test cases according to the new logic. 
I would suggest at least running the following test case to verify the change.
https://github.com/sonic-net/sonic-mgmt/blob/master/tests/platform_tests/cli/test_show_platform.py

Junchao-Mellanox:
Same concern as Kebo mentioned.

gregoryboudreau:
@keboliu @Junchao-Mellanox 
To my understanding, this has no impact on `show platform psustatus`. This change should only effect `show platform temperature` as the modification changes how temperatures are stored within the temperature DB. Specifically making it 1) so that if a device is removed during runtime (in a fixed platform) that the temperature sensors associated with it are not left stale within the DB, in the exact same manner as is done already within modular chassis and 2) so that if a PSU is disabled/turned off during runtime but is still present in the chassis, that it's sensors are not included in the temperature DB either.

I have confirmed that this DOES pass `platform_tests/cli/test_show_platform.py::test_show_platform_temperature[mth64-m5-2] PASSED ` and the behavior in the test case for `test_show_platform_psustatus` is also unchanged from before changes are applied. Additionally, this change should not warrant/require any changes to the CLI reference as the overall functionality and display format of `show platform temperature` remains completely untouched. To reiterate from above, all this change does is change the behavior of when sensors are removed/included from the temperature database. All other components (especially its display and formatting) of `show platform temperature` remain unchanged.

gechiang:
@prgeor , when you have a chance, please help review this PR.
Thanks!

gechiang:
MSFT ADO: 25872447

bmridul (dismissed):

bmridul:
@Junchao-Mellanox, @keboliu , Pls check Greg's response to your comments.

sonic-thermalctld/scripts/thermalctld
```diff
@@ -610,19 +614,18 @@ class TemperatureUpdater(logger.Logger):
...
-
-                        available_thermals.add((thermal, psu_name, thermal_index))
-                        self._refresh_temperature_status(psu_name, thermal, thermal_index)
-
-
-            thermals_to_remove = self.module_thermals - available_thermals
-            self.module_thermals = available_thermals
-            for thermal, parent_name, thermal_index in thermals_to_remove:
-                self._remove_thermal_from_db(thermal, parent_name, thermal_index)
+                    if psu.get_presence() and psu.get_powergood_status():
```
prgeor, line 617: @gregoryboudreau why don't we want to add PSU which is present but does not have good power status to the DB?

prgeor:
@gechiang LGTM. Please wait for @keboliu or @Junchao-Mellanox to also approve if no concern.

sonic-thermalctld/scripts/thermalctld
gregoryboudreau: This is two fold, first is to remove stale sensors from the db in the case where a psu is present and active at first but during runtime is disabled while second is that adding sensors for devices that are not in good power status introduces potentially undefined behavior of introducing sensors to the db that are from a not active or faulting PSU.

keboliu:
> @keboliu @Junchao-Mellanox To my understanding, this has no impact on `show platform psustatus`. This change should only effect `show platform temperature` as the modification changes how temperatures are stored within the temperature DB. Specifically making it 1) so that if a device is removed during runtime (in a fixed platform) that the temperature sensors associated with it are not left stale within the DB, in the exact same manner as is done already within modular chassis and 2) so that if a PSU is disabled/turned off during runtime but is still present in the chassis, that it's sensors are not included in the temperature DB either.
> 
> I have confirmed that this DOES pass `platform_tests/cli/test_show_platform.py::test_show_platform_temperature[mth64-m5-2] PASSED ` and the behavior in the test case for `test_show_platform_psustatus` is also unchanged from before changes are applied. Additionally, this change should not warrant/require any changes to the CLI reference as the overall functionality and display format of `show platform temperature` remains completely untouched. To reiterate from above, all this change does is change the behavior of when sensors are removed/included from the temperature database. All other components (especially its display and formatting) of `show platform temperature` remain unchanged.

Hi @gregoryboudreau  you are correct that the only impacted CLI will be "show platform temperature" in the case that one PSU is not present.

original output:
```
admin@sonic# show platform temperature 
                Sensor    Temperature    High TH    Low TH    Crit High TH    Crit Low TH    Warning          Timestamp
----------------------  -------------  ---------  --------  --------------  -------------  ---------  -----------------
                  ASIC           44.0      105.0       N/A           120.0            N/A      False  20231117 03:59:11
         CPU Pack Temp           70.0       82.0       N/A           104.0            N/A      False  20231117 03:59:11
            PSU-1 Temp            N/A        N/A       N/A             N/A            N/A      False  20231117 03:59:11
            PSU-2 Temp          26.25       60.0       N/A             N/A            N/A      False  20231117 03:59:11
    xSFP module 1 Temp            N/A        N/A       N/A             N/A            N/A      False  20231117 03:59:11
    xSFP module 2 Temp           44.0       70.0       N/A            80.0            N/A      False  20231117 03:59:11
    xSFP module 3 Temp           44.0       70.0       N/A            80.0            N/A      False  20231117 03:59:11
.......
```

output after the change, will see that the not present PSU removed from the output.

```
admin@sonic:/tmp# show platform temperature 
                Sensor    Temperature    High TH    Low TH    Crit High TH    Crit Low TH    Warning          Timestamp
----------------------  -------------  ---------  --------  --------------  -------------  ---------  -----------------
                  ASIC           43.0      105.0       N/A           120.0            N/A      False  20231117 05:00:36
         CPU Pack Temp           72.0       82.0       N/A           104.0            N/A      False  20231117 05:00:36
            PSU-2 Temp           26.0       60.0       N/A             N/A            N/A      False  20231117 05:00:36
    xSFP module 1 Temp            N/A        N/A       N/A             N/A            N/A      False  20231117 05:00:36
    xSFP module 2 Temp           45.0       70.0       N/A            80.0            N/A      False  20231117 05:00:36
    xSFP module 3 Temp           45.0       70.0       N/A            80.0            N/A      False  20231117 05:00:36
......
```

Keeping the not-present PSU in the system gives us a full picture of all the expected thermal sensors, including those that are not supported or that have issues.
This is not just for the PSU - other sensors, like cable temperature sensors, may also show up as 'N/A,' which could mean that they are either not supported or that something has gone wrong.
Removing the not-present PSU would only obscure information, and go against the original design intention.

sonic-thermalctld/scripts/thermalctld
prgeor: @gregoryboudreau doesn't the user want to see the status of the sensor if its not in good power status? Why do you want to hide it from the user/CLI output? I think it would be better to add another column in the output for "Status" which will indicate the current status of the sensor: "Present", "Absent", "Error", "Disabled", "Not Supported" etc..

gechiang:
> > @keboliu @Junchao-Mellanox To my understanding, this has no impact on `show platform psustatus`. This change should only effect `show platform temperature` as the modification changes how temperatures are stored within the temperature DB. Specifically making it 1) so that if a device is removed during runtime (in a fixed platform) that the temperature sensors associated with it are not left stale within the DB, in the exact same manner as is done already within modular chassis and 2) so that if a PSU is disabled/turned off during runtime but is still present in the chassis, that it's sensors are not included in the temperature DB either.
> > I have confirmed that this DOES pass `platform_tests/cli/test_show_platform.py::test_show_platform_temperature[mth64-m5-2] PASSED ` and the behavior in the test case for `test_show_platform_psustatus` is also unchanged from before changes are applied. Additionally, this change should not warrant/require any changes to the CLI reference as the overall functionality and display format of `show platform temperature` remains completely untouched. To reiterate from above, all this change does is change the behavior of when sensors are removed/included from the temperature database. All other components (especially its display and formatting) of `show platform temperature` remain unchanged.
> 
> Hi @gregoryboudreau you are correct that the only impacted CLI will be "show platform temperature" in the case that one PSU is not present.
> 
> original output:
> 
> ```
> admin@sonic# show platform temperature 
>                 Sensor    Temperature    High TH    Low TH    Crit High TH    Crit Low TH    Warning          Timestamp
> ----------------------  -------------  ---------  --------  --------------  -------------  ---------  -----------------
>                   ASIC           44.0      105.0       N/A           120.0            N/A      False  20231117 03:59:11
>          CPU Pack Temp           70.0       82.0       N/A           104.0            N/A      False  20231117 03:59:11
>             PSU-1 Temp            N/A        N/A       N/A             N/A            N/A      False  20231117 03:59:11
>             PSU-2 Temp          26.25       60.0       N/A             N/A            N/A      False  20231117 03:59:11
>     xSFP module 1 Temp            N/A        N/A       N/A             N/A            N/A      False  20231117 03:59:11
>     xSFP module 2 Temp           44.0       70.0       N/A            80.0            N/A      False  20231117 03:59:11
>     xSFP module 3 Temp           44.0       70.0       N/A            80.0            N/A      False  20231117 03:59:11
> .......
> ```
> 
> output after the change, will see that the not present PSU removed from the output.
> 
> ```
> admin@sonic:/tmp# show platform temperature 
>                 Sensor    Temperature    High TH    Low TH    Crit High TH    Crit Low TH    Warning          Timestamp
> ----------------------  -------------  ---------  --------  --------------  -------------  ---------  -----------------
>                   ASIC           43.0      105.0       N/A           120.0            N/A      False  20231117 05:00:36
>          CPU Pack Temp           72.0       82.0       N/A           104.0            N/A      False  20231117 05:00:36
>             PSU-2 Temp           26.0       60.0       N/A             N/A            N/A      False  20231117 05:00:36
>     xSFP module 1 Temp            N/A        N/A       N/A             N/A            N/A      False  20231117 05:00:36
>     xSFP module 2 Temp           45.0       70.0       N/A            80.0            N/A      False  20231117 05:00:36
>     xSFP module 3 Temp           45.0       70.0       N/A            80.0            N/A      False  20231117 05:00:36
> ......
> ```
> 
> Keeping the not-present PSU in the system gives us a full picture of all the expected thermal sensors, including those that are not supported or that have issues. This is not just for the PSU - other sensors, like cable temperature sensors, may also show up as 'N/A,' which could mean that they are either not supported/plugged in or that something has gone wrong. Removing the not-present PSU would only obscure information, and go against the original design intention.

@gregoryboudreau please help clarify:
If the PSU is present but went bad, what would this new change show on the "show platform temperature"? Will it not be shown as well?  I think for this case we should still show the output "NA" as long as the PSU is "present".  Even if the temperature sensor are not readable.  being "present" qualify it as "Expected" to be there. and any readout including "NA" is correct and further indicate there is some issue with the PSU which is present but not functioning.

@keboliu,  If the PSU is NOT Present, I don't see why we think it is "expected".  Let's say we have slots for 2 PSUs but based on current device requirement 1 is enough to power up the device and the second one is for backup/redundancy purpose.  Then what if we only have one PSU and another one was removed for RMA?  should we display the one that was removed for RMA in "show platform temperature" if it is not present? How do I differentiate this case to the case where the PSU is present and went bad? Probably not much help to see NA for it since it is not present, it will always show N/A... Would show platform psustatus show the "not present" PSU? If it does, then we have mechanism to tell it is not present and therefore no need to see it as "NA" under "show platform temperature".  Having slot available for PSUs does not mean all are expected to be inserted. But as long as we have ways to show accurately which slot is missing PSU then it should be good enough and I think that is the purpose for "show platform psustatus".  The purpose of "show platform temperature" is to display temperature sensors that are available for read out and they must be "present" to be read out.

gregoryboudreau:
> > > @keboliu @Junchao-Mellanox To my understanding, this has no impact on `show platform psustatus`. This change should only effect `show platform temperature` as the modification changes how temperatures are stored within the temperature DB. Specifically making it 1) so that if a device is removed during runtime (in a fixed platform) that the temperature sensors associated with it are not left stale within the DB, in the exact same manner as is done already within modular chassis and 2) so that if a PSU is disabled/turned off during runtime but is still present in the chassis, that it's sensors are not included in the temperature DB either.
> > > I have confirmed that this DOES pass `platform_tests/cli/test_show_platform.py::test_show_platform_temperature[mth64-m5-2] PASSED ` and the behavior in the test case for `test_show_platform_psustatus` is also unchanged from before changes are applied. Additionally, this change should not warrant/require any changes to the CLI reference as the overall functionality and display format of `show platform temperature` remains completely untouched. To reiterate from above, all this change does is change the behavior of when sensors are removed/included from the temperature database. All other components (especially its display and formatting) of `show platform temperature` remain unchanged.
> > 
> > 
> > Hi @gregoryboudreau you are correct that the only impacted CLI will be "show platform temperature" in the case that one PSU is not present.
> > original output:
> > ```
> > admin@sonic# show platform temperature 
> >                 Sensor    Temperature    High TH    Low TH    Crit High TH    Crit Low TH    Warning          Timestamp
> > ----------------------  -------------  ---------  --------  --------------  -------------  ---------  -----------------
> >                   ASIC           44.0      105.0       N/A           120.0            N/A      False  20231117 03:59:11
> >          CPU Pack Temp           70.0       82.0       N/A           104.0            N/A      False  20231117 03:59:11
> >             PSU-1 Temp            N/A        N/A       N/A             N/A            N/A      False  20231117 03:59:11
> >             PSU-2 Temp          26.25       60.0       N/A             N/A            N/A      False  20231117 03:59:11
> >     xSFP module 1 Temp            N/A        N/A       N/A             N/A            N/A      False  20231117 03:59:11
> >     xSFP module 2 Temp           44.0       70.0       N/A            80.0            N/A      False  20231117 03:59:11
> >     xSFP module 3 Temp           44.0       70.0       N/A            80.0            N/A      False  20231117 03:59:11
> > .......
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > output after the change, will see that the not present PSU removed from the output.
> > ```
> > admin@sonic:/tmp# show platform temperature 
> >                 Sensor    Temperature    High TH    Low TH    Crit High TH    Crit Low TH    Warning          Timestamp
> > ----------------------  -------------  ---------  --------  --------------  -------------  ---------  -----------------
> >                   ASIC           43.0      105.0       N/A           120.0            N/A      False  20231117 05:00:36
> >          CPU Pack Temp           72.0       82.0       N/A           104.0            N/A      False  20231117 05:00:36
> >             PSU-2 Temp           26.0       60.0       N/A             N/A            N/A      False  20231117 05:00:36
> >     xSFP module 1 Temp            N/A        N/A       N/A             N/A            N/A      False  20231117 05:00:36
> >     xSFP module 2 Temp           45.0       70.0       N/A            80.0            N/A      False  20231117 05:00:36
> >     xSFP module 3 Temp           45.0       70.0       N/A            80.0            N/A      False  20231117 05:00:36
> > ......
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Keeping the not-present PSU in the system gives us a full picture of all the expected thermal sensors, including those that are not supported or that have issues. This is not just for the PSU - other sensors, like cable temperature sensors, may also show up as 'N/A,' which could mean that they are either not supported/plugged in or that something has gone wrong. Removing the not-present PSU would only obscure information, and go against the original design intention.
> 
> @gregoryboudreau please help clarify: If the PSU is present but went bad, what would this new change show on the "show platform temperature"? Will it not be shown as well? I think for this case we should still show the output "NA" as long as the PSU is "present". Even if the temperature sensor are not readable. being "present" qualify it as "Expected" to be there. and any readout including "NA" is correct and further indicate there is some issue with the PSU which is present but not functioning.
> 
> @keboliu, If the PSU is NOT Present, I don't see why we think it is "expected". Let's say we have slots for 2 PSUs but based on current device requirement 1 is enough to power up the device and the second one is for backup/redundancy purpose. Then what if we only have one PSU and another one was removed for RMA? should we display the one that was removed for RMA in "show platform temperature" if it is not present? How do I differentiate this case to the case where the PSU is present and went bad? Probably not much help to see NA for it since it is not present, it will always show N/A... Would show platform psustatus show the "not present" PSU? If it does, then we have mechanism to tell it is not present and therefore no need to see it as "NA" under "show platform temperature". Having slot available for PSUs does not mean all are expected to be inserted. But as long as we have ways to show accurately which slot is missing PSU then it should be good enough and I think that is the purpose for "show platform psustatus". The purpose of "show platform temperature" is to display temperature sensors that are available for read out and they must be "present" to be read out.

@gechiang
If the PSU is present but went bad, its thermal sensors would be removed from the temperature db with the changes in this PR. If a temperature sensor is from a device that is bad then there is very little reason to include in the temperature db as it provides the user with minimally useful information while muddling the information stored in the db and causing issues in some tests (to my knowledge tests like platform_tests/test_thermal_state_db.py require the sensors to be readable). My issue with keeping the sensors in the db for devices that are present is that it does NOT tell us if the issues are with the sensor or the PSU and could provide the user with misleading information (especially with N/A values as, in my opinion with this scope of info that the cli provides, they really don't tell the user much useful information at all because of how vague it is). Also, in regards to helping "to indicate there is some issue with the psu", while these temperature sensors could potentially provide additional context, I am of the opinion that that is something that should really only fall under the purview of `show platform psustatus`, the temperature cli is defined as focusing on the status of the temperature sensors, if a device  is not in good operation, should the user blanket trust the values its temperature sensors are reporting (some PSUs could have temperature sensors that would still work while others wouldn't)?

@prgeor I think this comes down to what we want to pursue as the intent of the `show platform temperature` command, should it show all 'expected' sensors (all thermal sensors from devices that are present) or should it only show 'expected' sensors from devices that are operational as its definition does not clarify this... 

@keboliu 
There is no reason to display the sensors of devices that are not present. It only serves to confuse the user. No meaningful data can be gleamed by populating this db with what is practically trash information. While a device may have (and expect) multiple PSUs for example, if it can function with one PSU then what purpose does providing empty information points actually serve? I don't think removing this info from the db obscures anything, it just ensures that the platform provides the db with information the platform actually has about its' devices rather than also outlining what would be akin to ghost information...

keboliu:
Hi @gechiang @gregoryboudreau  I understand your point,  but
1. To my knowledge, on a modular system, the second PSU is a must for redundancy. though with one PSU it still can work, redundant PSUs are standard practice in a production environment. While it's true that some systems may not require all PSUs to be present, we shouldn't make changes to the UI(CLI output) based on exceptional cases.
2. Your point is reasonable from some angle, but please keep in mind this PR is changing the current behavior of the system. The current output of the CLI provides a clear and consistent view of the system's temperature sensors, and any change could cause confusion for users. In short, the benefits of the change should outweigh the potential negative impact on the UI.

gregoryboudreau:
> Hi @gechiang @gregoryboudreau I understand your point, but
> 
> 1. To my knowledge, on a modular system, the second PSU is a must for redundancy. though with one PSU it still can work, redundant PSUs are standard practice in a production environment. While it's true that some systems may not require all PSUs to be present, we shouldn't make changes to the UI(CLI output) based on exceptional cases.
> 2. Your point is reasonable from some angle, but please keep in mind this PR is changing the existing CLI output. The current output of the CLI provides a clear and consistent view of the system's temperature sensors, and any change could cause confusion for users. In short, the benefits of the change should outweigh the potential negative impact on the UI.
> 3. A possible alternative solution to address this issue on the platform that does not require all the PSUs to be present:  inside the platform API implementation, ignore the not required PSUs, and don't create a PSU entity for them. Does it sound ok?

@keboliu 
I really don't understand including the data from not present PSUs especially. There is an argument to be made for keeping the information from the thermal sensors of PSUs in not good power status (though I still don't think they should personally as the provide potentially confusing and incorrect information). But for not present PSUs, there is zero information being provided by zero readable sensors. `show platform temperature` has no reason to show non readable sensors from non present devices. If a device requires two PSUs then `show platform temperature` is not the place to enforce that. This CLI should only show actual information regarding the state/information of sensors on devices that are actually present. The purpose of this change is that the information provided by the CLI is unclear as it does not have a way to designate N/A values meaning. It could be anything from not present to not on to not functioning.

gechiang:
@keboliu , I think the purpose of "show platform temperature" is to display all temperatures from components that are "present".  It absolutely makes no sense to display components that are not even present.  It is up to the user to decide whether they want to populate all the PSU slots or just keep it at minimum (to save cost perhaps, or in the middle of RMA replacement). Redundancy is not a must especially if the device purpose is not mission critical for the user.  
So for components that are not present we should make the change and not keep them in the stateDB.  I don't think the current behavior is correct and we should fix it with this PR.

@gregoryboudreau, I also do not agree to remove BAD PSUs that are "present" from the stateDB from being displayed by "show platform temperature". We must trust the "Presence" signal and still attempt to read out the sensors. We don't know if we can trust the sensor data if the PSU is bad.  But it does provide some useful info with readout that raise concerns in case it went bad or in the process of becoming bad with temperature swings read out so that it allows us to detect and perform troubleshooting/RMA before it totally goes dead.  It could be that it got declared as "BAD" way too early or falsely and in that case as long as the PSU is "present", the temperature data can provide some usefulness. Temperature changes is a good way to provide alert signals.  Also, this provides one additional signal that we can use besides totally depended on "show platform psustatus" to detect the problem.  I do not think all bad PSU cases will lead to no temperature sensor can be read out.  As long as the component is "present" we should try to read out the temperatures from it.  Can you change the PR to restore the readout for PSUs that are "present"?

Thanks!

sonic-thermalctld/scripts/thermalctld
gechiang, line 617: Please change this to trust the "present" signal only. PSU good or bad is not a criteria for temperature sensor readout.

gechiang (changes requested):
Please remove the criteria of PSU must be good. as long as it is present, it should be good enough for sensor readout.

sonic-thermalctld/scripts/thermalctld
gregoryboudreau: Updated.

gregoryboudreau:
As requested, have removed the powergood check, presence check remains for PSUs.

gechiang (approved):

keboliu:
I think I made my ponit clear, and I don't prefer to change the current CLI output since it does provide comprehensive information to the user.  @gregoryboudreau if you don't want to have those PSU info in the CLI, could you please consider the alternative proposal in my previous comment:

> A possible alternative solution to address this issue on the platform that does not require all the PSUs to be present: inside the platform API implementation, ignore the not required PSUs, and don't create a PSU entity for them. Does it sound ok?

gregoryboudreau:
@keboliu To directly address your comment, yes this could be done at the platform level. However, this does not change that this is an issue within the daemon. To reiterate on what @gechiang said, it does not make sense to display sensors for devices that are not present. This does **not** provide the user with any usable information. While a device may have (and expect) multiple PSUs for example, if it can function with one PSU then what purpose does providing empty information points actually serve? The purpose of this CLI is to "displays the status of the device's thermal sensors." If a device is not present, then there are no thermal sensors to be read, so I don't see any reason to include them. Please explain why this information about the sensors on a device that is not present are important to this CLI.

spilkey-cisco:
> I think I made my ponit clear, and I don't prefer to change the current CLI output since it does provide comprehensive information to the user. @gregoryboudreau if you don't want to have those PSU info in the CLI, could you please consider the alternative proposal in my previous comment:
> 
> > A possible alternative solution to address this issue on the platform that does not require all the PSUs to be present: inside the platform API implementation, ignore the not required PSUs, and don't create a PSU entity for them. Does it sound ok?

There is no requirement about which PSUs are required on not for the available slots on the chassis. @keboliu can you please comment on the most recent reply tagging you from @gechiang? `show platform temperature` does not include any kind of `status` or `presence` column, so displayed temperature data from a non-present PSU is not well defined. `show platform psustatus` is designed to show status and presence, so all PSUs (present or not) should be in the DB. Displaying 'N/A' for temperature sensors in `show platform temperature` that are not present makes the CLI ambiguous as to whether they are faulty or just not present.

What is the intended behavior for `get_all_thermals` when called on a non-present module? If that case should not return any thermals, it should be clearly documented in the base class and vendor code can be adjusted to comply. That would negate the need for a presence check in this PR.

prgeor:
@gregoryboudreau latest changes LGTM. @keboliu please specify what is blocking and why.

prgeor:
> @gregoryboudreau latest changes LGTM. @keboliu please specify what is blocking and why.

@keboliu doesn't have anything blocking.

prgeor:
@gregoryboudreau I am OK to have this change in master. Why we need this in other branches? Which functionality is broken?

gregoryboudreau:
@prgeor 
Needed for other branches as it fixes bug where stale thermal sensors will be left in temp db and reported on until device power cycle on fixed platform. (extension of the dynamic sensor logic from modular platform to fixed platforms)

prgeor:
> @prgeor Needed for other branches as it fixes bug where stale thermal sensors will be left in temp db and reported on until device power cycle on fixed platform. (extension of the dynamic sensor logic from modular platform to fixed platforms)

@gregoryboudreau  CLI/DB part I understand. Which functionality is broken? ... like what is broken that mandates this to be ported to other stable branches 202205?

prgeor (approved):

gregoryboudreau:
Because of the stale sensors, if a device is removed and tests are run without a power cycle, it can lead to the failure of mgmt tests such as platform_tests/test_thermal_state_db.py. It is also incorrect functionality for these stale sensors to remain in the first place (though not sure how pertinent that is to the branch discussion). These changes were begun due to the above described test failure being reported to us by MSFT on 202205 devices. 

If this is unable to get into 202205, is it still up for 202305?

gregoryboudreau:
@prgeor Pending the decision for potential inclusion in the past release branches, can this be merged into master?

gechiang:
@yxieca , @StormLiangMS , Please help review/approve this for 202205 and 202305.
MSFT ADO: 25872447
Thanks!

gechiang:
@gregoryboudreau please raise a separate PR under 202205 to resolve the conflict as this POR cannot be cherry-picked directly...
Thanks!

StormLiangMS:
@gregoryboudreau @gechiang could you pls have a test on top of 202305 with this PR? to avoid regression at this stage.

gregoryboudreau:
This PR doesn't have that label?

gechiang:
@gregoryboudreau This PR is not possible to be picked cleanly into 202311 directly.  Please create a PR on top of the 202311 branch. Thanks!

gregoryboudreau:
@gechiang I might be missing something, what part of this PR isn't in 202311? https://github.com/sonic-net/sonic-platform-daemons/blob/202311/sonic-thermalctld/scripts/thermalctld#L569, it looks like its there?

gechiang:
> @gechiang I might be missing something, what part of this PR isn't in 202311? https://github.com/sonic-net/sonic-platform-daemons/blob/202311/sonic-thermalctld/scripts/thermalctld#L569, it looks like its there?

You are absolutely correct.  So it looks like this PR got merged at the border line when the 202311 got branched out which picked it up already.  Nice!  One less thing to worry about.
I will go ahead remove the request for 202311 as it is already there.
Thanks for responding promptly!

PR #403, opened 2023-11-02
Changes for port reinitialization in case of syncd/swss/oa crash  and NPU SI settings update for CMIS transceivers

mihirpat1:
#### Description

Code changes for implementing the HLD https://github.com/sonic-net/SONiC/pull/1432
This is needed to ensure that XCVRD re-initializes all ports if syncd/swss/orchagent crashes. 

#### Motivation and Context

In addition to the implementation mention in the HLD, following are the changes being done through this PR
	1. Remove wait for port config completion from CMIS thread
	2. Add print for module and DP states
	3. Change "media settings" to NPU SI settings to clearly differentiate optics and NPU SI settings

In progress

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -48,6 +48,8 @@ TRANSCEIVER_PM_TABLE = 'TRANSCEIVER_PM'
 
 TRANSCEIVER_STATUS_TABLE_SW_FIELDS = ["status", "error"]
 
+CMIS_MODULE_TYPES    = ['QSFP-DD', 'QSFP_DD', 'OSFP', 'QSFP+C']
+
 # Mgminit time required as per CMIS spec
 MGMT_INIT_TIME_DELAY_SECS = 2
 
@@ -631,7 +633,7 @@ def get_media_settings_value(physical_port, key):
             if len(default_dict) != 0:
                 return default_dict
             else:
-                helper_logger.log_error("Error: No values for physical port '{}'".format(physical_port))
+                helper_logger.log_info("Error: No values for physical port '{}'".format(physical_port))
             return {}
 
         if key[0] in media_dict:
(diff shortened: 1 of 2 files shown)


mihirpat1:
@prgeor @shyam77git @jaganbal-a - It will be great if you can help in reviewing this.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -48,6 +48,8 @@
 
 TRANSCEIVER_STATUS_TABLE_SW_FIELDS = ["status", "error"]
 
+CMIS_MODULE_TYPES    = ['QSFP-DD', 'QSFP_DD', 'OSFP', 'QSFP+C']
```
keboliu, line 51: can we add a new API to detect whether the cable is CMIS? otherwise, we have to update this list when new cables are supported.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -631,7 +633,7 @@ def get_media_settings_value(physical_port, key):
             if len(default_dict) != 0:
                 return default_dict
             else:
-                helper_logger.log_error("Error: No values for physical port '{}'".format(physical_port))
+                helper_logger.log_info("Error: No values for physical port '{}'".format(physical_port))
```
keboliu, line 636: why not a Error level log?

mihirpat1:
@tshalvi - It will be great if you can review this PR

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: On SKUs which have media_settings.json file present, but have ports with transceivers not requiring NPU SI settings, we see this error.
Hence, to handle such scenario, I am planning to change this to info.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2434,6 +2668,43 @@ def load_media_settings(self):
...
+            if "NPU_SI_SETTINGS_SYNC_STATUS" not in app_port_table_fvs_dict:
+                appl_db.set(appl_db.APPL_DB, 'PORT_TABLE:{}'.format(lport), "NPU_SI_SETTINGS_SYNC_STATUS", "NPU_SI_SETTINGS_DEFAULT")
+                self.log_notice("Added NPU_SI_SETTINGS_SYNC_STATUS for lport {}".format(lport))
+
+        self.log_notice("XCVRD INIT: Port init fields initialized in APP_PORT_TABLE")
+
+    """
+    Subscribes APPL_DB PORT_TABLE table events
+    """
+    def subscribe_appl_port_table(self):
```
keboliu, line 2697: there is an existing function  `wait_for_port_config_done(self, namespace)` that has the same logic, suggest reducing the duplicated code by calling this new function in that one.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2421,6 +2654,7 @@ def wait_for_port_config_done(self, namespace):
             (key, op, fvp) = port_tbl.pop()
             if key in ["PortConfigDone", "PortInitDone"]:
                 break
+        sel.removeSelectable(port_tbl)
```
keboliu, line 2657: in the newly added code it also subscribes to the same table, is possible to keep this subscription and reuse it?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2542,9 +2823,34 @@ def run(self):
...
-        self.stop_event.wait()
+        sel, asic_context  = self.subscribe_appl_port_table()
+        self.log_notice("Subscribed to appl {} update".format(swsscommon.APP_PORT_TABLE_NAME))
 
-        self.log_info("Stop daemon main loop")
+        generate_sigabrt = False
+
+        while not self.stop_event.is_set() and not generate_sigabrt:
+            (state, _) = sel.select(port_mapping.SELECT_TIMEOUT_MSECS)
+            if state == swsscommon.Select.TIMEOUT:
```
keboliu, line 2833: seems to me following condition `if state != swsscommon.Select.OBJECT:` is enough and this one is not needed?

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have removed this now and added a API to identify CMIS based transceivers using "isinstance" API.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have now removed the duplicate code and xcvrd is now subscribing only once for APPL_DB updates

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have addressed this now.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have kept this check to avoid flooding of logs when select times out

prgeor:
@mihirpat1 can you check the build failure and coverage

mihirpat1:
> @mihirpat1 can you check the build failure and coverage

I have resolved it now.

kevinskwang:
@mihirpat1  could you resolve the conflicts?

keboliu (approved):

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1502,6 +1688,20 @@ def task_worker(self):
...
+                        # update through post_port_active_apsel_to_db
+                        asic_id = self.port_mapping.get_asic_id_for_logical_port(lport)
+                        cmis_reinit_rqd_val = get_state_db_port_table_val_by_key(
+                                        self.xcvr_table_helper, self.port_mapping, lport, "CMIS_REINIT_REQUIRED")
+                        if cmis_reinit_rqd_val is None:
+                            self.log_error("{}: CMIS_STATE_DP_ACTIVATE state: cmis_reinit_rqd_val is None".format(lport))
+                        else:
+                            if cmis_reinit_rqd_val == "true":
+                                state_port_table = self.xcvr_table_helper.get_state_port_tbl(asic_id)
+                                state_port_table.set(lport, [("CMIS_REINIT_REQUIRED", "false")])
```
longhuan-cisco, line 1702: Not sure if `CMIS_REINIT_REQUIRED` is the best flag/naming, it seems to be ambiguous since re-init can also be required in cases such as error/timeout/host_tx_ready_change/etc. 
Moreover, `force_cmis_reinit()` can get triggered at any time if the listened DB contents get touched/`SET` , please refer to below code snippet (though not sure if the code is doing the correct thing by expanding the coverage of re-init to so many cases): https://github.com/sonic-net/sonic-platform-daemons/blob/70e875e5f21d408bedc4251164f1898c48c205ea/sonic-xcvrd/xcvrd/xcvrd.py#L912-L930


There might be a better flag/naming. (maybe `CMIS_INIT_DONE`/`CMIS_DONE`/`XXX_DONE`/etc is less ambiguous? considering `DONE` can only be `True` when state machine goes to certain final state)

liat-grozovik:
@mihirpat1 @prgeor any plan to handle the conflict and merge this PR?

PR #409, opened 2023-12-04
Add dynamic sensor logic for fixed and psu presence/state checking in thermalctld, 202205 merge conflicts addressed

gregoryboudreau:
Add dynamic sensor logic for fixed and psu presence/state checking in thermalctld. As requested, creating separate double commit for 202205 to address merge conflict.

Description

Extends the existing dynamic sensor logic of modular chassis to fixed and adds a check for PSU presence/status in thermalctld

Motivation and Context

In current form, if a PSU or other device with sensors is removed from the device or disabled during runtime on a fixed chassis, they will become stale and remain in the temperature sensor DB which should not be correct. This change brings the existing logic checks that modular chassis' have to fixed platform as well as checks for PSU presence/status to account for scenario where PSU may be disabled during runtime.

How Has This Been Tested?

On both fixed and modular platforms (Cisco 8800RP and Cisco 8101_32FH), I confirmed presence of all thermal sensors. Then disable the PSU and confirmed its associated sensors were removed from the DB while it was either inactive and present or not present. Reenabled PSU and confirmed its associated sensors were updated back into the DB. `test_show_platform.py::test_show_platform_temperature` passes.

Additional Information (Optional)
For more context, please see https://github.com/sonic-net/sonic-platform-daemons/pull/401

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -541,10 +541,10 @@ class TemperatureUpdater(logger.Logger):
         self.table = swsscommon.Table(state_db, TemperatureUpdater.TEMPER_INFO_TABLE_NAME)
         self.phy_entity_table = swsscommon.Table(state_db, PHYSICAL_ENTITY_INFO_TABLE)
         self.chassis_table = None
+        self.all_thermals = set()
 
         self.is_chassis_system = chassis.is_modular_chassis()
         if self.is_chassis_system:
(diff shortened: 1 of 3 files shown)


gechiang (approved):

gechiang:
@yxieca please help approve/merge this PR that @gregoryboudreau has made to resolve the cherry-pick issue.
copied the MSFT ADO here for your reference: MSFT ADO: 25872447
Thanks!

PR #430, opened 2024-02-07
Enhance port_mapping to be thread-safe

longhuan-cisco:
#### Description
1. Make class variable `PORT_EVENT ` to be thread-safe.
2. Add new argument to allow users to decide what tables and fields to subscribe via subscribe_port_update_event()
3. Record table_name and db_name for each event to avoid ambiguity in certain cases. (e.g. for `DEL` events from different tables/DBs, user may have different handlings.)

#### Motivation and Context

Today's port_mapping has a class variable `PORT_EVENT `, which is not thread-safe: different threads could share this class level global variable PORT_EVENT, and overwriting each other's data.

It's not causing issue in today's code, because today only one thread (cmis_mgr) relying on port_mapping.handle_port_update_event (and PORT_EVENT).
But in the future, it will run into issues if multiple threads (e.g. cmis_mgr, sff_mgr, PM) rely on function port_mapping.handle_port_update_event (where PORT_EVENT is getting used) 

This PR is a dependency of #383

Verified cmis_mgr/sff_mgr(new thread) to be able to work normally to bring up/down optics without intervention with each other.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
+++ b/sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
@@ -1,19 +1,31 @@
+import threading
 from sonic_py_common import daemon_base
 from sonic_py_common import multi_asic
 from sonic_py_common.interface import backplane_prefix, inband_prefix, recirc_prefix
 from swsscommon import swsscommon
 
 SELECT_TIMEOUT_MSECS = 1000
+DEFAULT_PORT_TBL_MAP = [
+    {'CONFIG_DB': swsscommon.CFG_PORT_TABLE_NAME},
+    {'STATE_DB': 'TRANSCEIVER_INFO'},
+    {'STATE_DB': 'PORT_TABLE', 'FILTER': ['host_tx_ready']},
+]
 
 
 class PortChangeEvent:
+    # thread_local_data will be used to hold class-scope thread-safe variables
+    # (i.e. global variable to the local thread), such as
+    # thread_local_data.PORT_EVENT. thread_local_data.PORT_EVENT dict will be
+    # initialized in subscribe_port_update_event, and used to store the latest
+    # port change event for each key, to avoid duplicate event processing.
+    thread_local_data = threading.local()
     PORT_ADD = 0
     PORT_REMOVE = 1
     PORT_SET = 2
     PORT_DEL = 3
-    PORT_EVENT = {}
 
-    def __init__(self, port_name, port_index, asic_id, event_type, port_dict=None):
+    def __init__(self, port_name, port_index, asic_id, event_type, port_dict=None,
+                 db_name=None, table_name=None):
         # Logical port name, e.g. Ethernet0
         self.port_name = port_name
         # Physical port index, equals to "index" field of PORT table in CONFIG_DB
@@ -24,6 +36,8 @@ class PortChangeEvent:
         self.event_type = event_type
         # Port config dict
         self.port_dict = port_dict
+        self.db_name = db_name
+        self.table_name = table_name
 
     def __str__(self):
         return '{} - name={} index={} asic_id={}'.format('Add' if self.event_type == self.PORT_ADD else 'Remove',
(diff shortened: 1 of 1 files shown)


sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
```diff
@@ -1,19 +1,31 @@
...
+]
 
 
 class PortChangeEvent:
+    # thread_local_data will be used to hold class-scope thread-safe variables
+    # (i.e. global variable to the local thread), such as
+    # thread_local_data.PORT_EVENT. thread_local_data.PORT_EVENT dict will be
+    # initialized in subscribe_port_update_event, and used to store the latest
+    # port change event for each key, to avoid duplicate event processing.
+    thread_local_data = threading.local()
```
prgeor, line 21: @longhuan-cisco can we have a much simpler solution? If you instantiate `PortChangeEvent` class per thread, it will be much easier as each instance will be accessed by their own thread context.

prgeor: @longhuan-cisco the problem is PortChangeEvent is being used like a global variable which is a bad choice. We should use object oriented approach which will scale better

longhuan-cisco: Yes, agreed. When I was writing the code for main PR (sff_mgr), I was thinking the same to go with object approach by change the global variable to object scope. But this would be a wider refactor (touching more code), including changing the caller side (cmis_mgr in xcvrd.py) and migrating those global functions (`subscribe_port_update_event` and `handle_port_update_event`) to class method/etc. Thus, at that time, I chose to go with the more self-contained and less-code-change solution, by simply changing the one single global variable to thread-safe without having to touching other part of the code.

Since now, we split the main PR and tracking this enhancement/fix as a separate PR, I think it's a good time to do this refactor. Will update accordingly.

longhuan-cisco: @prgeor I updated the diff to make the logic object oriented and get rid of all the global variables/functions.

prgeor (approved):

mihirpat1 (approved):

sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
prgeor, line 1: @longhuan-cisco can this file now be renamed appropriately to port_observer.py or port_event.py ?

sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
```diff
@@ -1,19 +1,25 @@
+import threading
 from sonic_py_common import daemon_base
 from sonic_py_common import multi_asic
 from sonic_py_common.interface import backplane_prefix, inband_prefix, recirc_prefix
 from swsscommon import swsscommon
 
 SELECT_TIMEOUT_MSECS = 1000
+DEFAULT_PORT_TBL_MAP = [
```
prgeor, line 8: @longhuan-cisco Thanks this looks much better than previous proposed solution. May I have one more suggestion. Currently this map is fixed to static filtering if the subscriber threads could pass this filter object while subscribing, then this will make the solution more generic like multiple threads could subscribe to different combination of events: DB/Table/Fields

sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
longhuan-cisco: renamed to `port_event_helper.py`  (appended suffix `_helper` to `port_event` to make it aligned with today's other helper scripts in this folder, such as `sfp_status_helper.py`)

sonic-xcvrd/xcvrd/xcvrd_utilities/port_mapping.py
longhuan-cisco: The code had already been made generic by allowing custom port_tbl_map argument (DB/table/fields)

https://github.com/sonic-net/sonic-platform-daemons/blob/abcf4c425a81befdce316317cd0c74e46a10bc18/sonic-xcvrd/xcvrd/xcvrd_utilities/port_event_helper.py#L48-L51

prgeor (approved):

PR #440, opened 2024-02-26
Update for supported CMIS module types in xcvrd to include a new module: OSFP-8X

tshalvi:
A new cable was added to the list of supported CMIS modules

#### Description
In xcvrd, within the CmisManagerTask thread, there is a list of supported CMIS modules. The CMIS state machine only
starts for modules found in this list. Currently, a certain type of CMIS module, OSFP-8X, is missing from this list. Hence, the
configuration for such modules will be skipped by the CmisManagerTask.

#### Motivation and Context
Without this change, xcvrd wouldn't be able to configure OSFP-8X modules.

I confirmed that when an OSFP-8X module is plugged in, the CMIS state machine is triggered, and the module SI
parameters are retrieved from optics_si_settings.json and configured successfully.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -759,7 +759,7 @@ class CmisManagerTask(threading.Thread):
 
     CMIS_MAX_RETRIES     = 3
     CMIS_DEF_EXPIRED     = 60 # seconds, default expiration time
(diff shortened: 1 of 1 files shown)


mihirpat1 (approved):

prgeor (approved):

dprital:
@yxieca , Can you please cherry pick to 202311 ?

PR #443, opened 2024-03-03
Enable periodic polling of TRANSCEIVER_FIRMWARE_INFO table in DomInfoUpdateTask

mihirpat1:
MSFT ADO - 26801871

#### Description

We need to enable periodic update of TRANSCEIVER_FIRMWARE_INFO table so that the active_firmware and inactive_firmware fields are periodically updated in the redis-db. Periodic polling of firmware version will enable XCVRD to update firmware version for all breakout ports of a physical port after performing firmware upgrade using any one of the breakout port.

Also, a new CLI (`config interface transceiver dom PORT_NAME enable/disable`) has been created to allow user enable/disable DOM monitoring. XCVRD needs to handle this CLI so that periodic polling through DomInfoUpdateTask thread is enabled/disabled for a particular port.
The new CLI will set `dom_status` field in PORT|\<logical_port\> table of CONFIG_DB. The possible values of this field are `enabled` and `disabled`
Please refer to https://github.com/sonic-net/sonic-utilities/pull/3187 for further details.

DOM monitoring should be disabled before executing `sfputil show fwversion PORT_NAME` to ensure that DomInfoUpdateTask thread does not send a CDB command parallelly to retrieve the FW version.

Following are the testcases which were tested and passed successfully
```
1	Device with non-breakout ports
1.1	Disable DOM monitoring for port
1.2	Enable DOM monitoring for port
2	Device with breakout ports
2.1	Disable DOM monitoring for subport 1
2.2	Enable DOM monitoring for subport 1
2.3	Disable DOM monitoring for subport 2
2.4	Enable DOM monitoring for subport 2
3	Multi-asic device - The CLI handler in xcvrd was not tested since all the available multi-asic devices currently support 202205 image and the TRANSCEIVER_FIRMWARE_INFO table is supported 202305 image onwards.
4	Firmware upgrade
5	Help options for DOM CLI
sonic:/home/admin# config interface transceiver dom --help
Usage: config interface transceiver dom [OPTIONS] <interface_name>
                                        (enable|disable)

  Enable/disable DOM monitoring for SFP transceiver module

Options:
  -?, -h, --help  Show this message and exit.
```


prgeor:
@keboliu can you review

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -530,8 +531,15 @@ def post_port_sfp_firmware_info_to_db(logical_port_name, port_mapping, table,
             continue
 
         try:
-            transceiver_firmware_info_dict = _wrapper_get_transceiver_firmware_info(physical_port)
-            if transceiver_firmware_info_dict is not None:
+            if firmware_info_cache is not None and physical_port in firmware_info_cache:
```
prgeor, line 534: @mihirpat1 what is this optimization? My understanding is the dom is going to read the firmware version from the EEPROM every polling interval

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1580,34 +1588,82 @@ def __init__(self, namespaces, port_mapping, main_thread_stop_event):
         self.port_mapping = copy.deepcopy(port_mapping)
         self.namespaces = namespaces
 
+    def get_dom_status_from_config_db(self, lport):
```
prgeor, line 1591: @mihirpat1 suggest dom_status -> dom_polling?

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: @prgeor - Upon further looking into the code after our discussion, I found that the current code reads data from EERPOM for the first subport for every poll. However, the remaining subports of that physical port in that iteration have their data retrieved from the relevant cache.
In the current case, the firmware_info_cache is a dictionary with `physical_port` as the key. Hence, DomInfoUpdateTask will clear the entire firmware_info_cache dictionary (line 1642) before it starts retrieving data for all ports. While handling the first subport of a physical port, DomInfoUpdateTask will read data from EEPROM and update firmware_info_cache. All other subports of the physical ports will have the data retrieved from firmware_info_cache in that iteration.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: @prgeor I have addressed this now.

prgeor (approved):

mihirpat1:
@StormLiangMS - I have created the below PR for 202305 cherry-pick. It will be great if you can help in merging it.
https://github.com/sonic-net/sonic-platform-daemons/pull/445

MSFT ADO - 26801871

PR #449, opened 2024-03-18
Disable periodic polling of port in DomInfoUpdateTask thread during CMIS init

mihirpat1:
#### Description

We currently need to disable periodic DOM polling of a port through DomInfoUpdateTask thread during CMIS initialization.

#### Motivation and Context

Disabling of DOM polling during CMIS initialization is primarily needed to disable sending CDB commands to read FW version from DomInfoUpdateTask thread during CMIS initialization.
For transceivers which do not support CDB background mode, any EEPROM access to the module can fail if a CDB command is executed at the same time.

In order to disable DOM polling during CMIS initialization, the `cmis_state` from CmisManagerTask thread is now being updated in the `TRANSCEIVER_STATUS` table of STATE_DB.
If the current `cmis_state` does not belong to `CMIS_TERMINAL_STATES`, DomInfoUpdateTask will disable DOM polling for the port to allow CmisManagerTask thread to complete CMIS initialization if required.
For platforms with CmisManagerTask disabled, the function `is_port_in_cmis_initialization_process` will always return False to allow DOM polling.
In case of device boot-up or transceiver insertion, the DomInfoUpdateTask thread will wait for the port to be in either of the CMIS_TERMINAL_STATES before proceeding with DOM polling. For non-CMIS transceivers, the expected `cmis_state` is `CMIS_STATE_READY`. Hence, once the corresponding port reaches `CMIS_STATE_READY` state, DOM polling will be enabled for such port.

Also, the `cmis_state` is not planned to be modified by DomInfoUpdateTask thread at any time to prevent race condition with CmisManagerTask.

Following is the summary of tests performed
```
1 Ensure DOM thread polls for non-CMIS transceivers
2 Ensure DOM thread polls for platform with CMIS manager disabled
3 Ensure DOM thread waits during CMIS initialization
4 Ensure DOM polling is resumed after transceiver insertion
4.1 After removal
4.2 After insertion
5. Ensured `show interface transceiver status` CLI works with the current changes

Redis-db dump snippet to show `cmis_state` field in TRANSCEIVER_STATUS table
root@sonic:/home/admin# redis-cli -n 6 hgetall "TRANSCEIVER_STATUS|Ethernet0"
  1) "cmis_state"
  2) "READY"
```
MSFT ADO - 26993372

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -776,6 +776,12 @@ class CmisManagerTask(threading.Thread):
     CMIS_STATE_REMOVED   = 'REMOVED'
     CMIS_STATE_FAILED    = 'FAILED'
 
+    CMIS_TERMINAL_STATES = {
+                            CMIS_STATE_FAILED,
+                            CMIS_STATE_READY,
+                            CMIS_STATE_REMOVED
+                            }
+
     def __init__(self, namespaces, port_mapping, main_thread_stop_event, skip_cmis_mgr=False):
         threading.Thread.__init__(self)
         self.name = "CmisManagerTask"
@@ -1170,6 +1176,9 @@ class CmisManagerTask(threading.Thread):
             self.log_error("{} Tuning in progress, subport selection may fail!".format(lport))
         return api.set_laser_freq(freq, grid)
 
+    def get_cmis_state(self, lport):
+        return self.port_dict[lport].get('cmis_state', self.CMIS_STATE_UNKNOWN)
+
     def post_port_active_apsel_to_db(self, api, lport, host_lanes_mask):
         try:
             act_apsel = api.get_active_apsel_hostlane()
@@ -1245,10 +1254,7 @@ class CmisManagerTask(threading.Thread):
                     continue
 
                 state = self.port_dict[lport].get('cmis_state', self.CMIS_STATE_UNKNOWN)
-                if state in [self.CMIS_STATE_UNKNOWN,
-                             self.CMIS_STATE_FAILED,
-                             self.CMIS_STATE_READY,
-                             self.CMIS_STATE_REMOVED]:
+                if state in self.CMIS_TERMINAL_STATES or state == self.CMIS_STATE_UNKNOWN:
                     if state != self.CMIS_STATE_READY:
                         self.port_dict[lport]['appl'] = 0
                         self.port_dict[lport]['host_lanes_mask'] = 0
@@ -1574,7 +1580,7 @@ class CmisManagerTask(threading.Thread):
 
 
 class DomInfoUpdateTask(threading.Thread):
-    def __init__(self, namespaces, port_mapping, main_thread_stop_event):
+    def __init__(self, namespaces, port_mapping, main_thread_stop_event, cmis_manager):
         threading.Thread.__init__(self)
         self.name = "DomInfoUpdateTask"
         self.exc = None
@@ -1582,6 +1588,7 @@ class DomInfoUpdateTask(threading.Thread):
         self.main_thread_stop_event = main_thread_stop_event
         self.port_mapping = copy.deepcopy(port_mapping)
         self.namespaces = namespaces
+        self.cmis_manager = cmis_manager
 
     def get_dom_polling_from_config_db(self, lport):
         """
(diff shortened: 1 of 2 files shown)


sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1619,8 +1626,31 @@ def get_dom_polling_from_config_db(self, lport):
...
         return dom_polling
 
+    """
+    Checks if the port is in CMIS terminal state
+    This API assumes CMIS_STATE_UNKNOWN as a transitional state since any CMIS supported platform will
+    eventually reach to a CMIS terminal state (CMIS_TERMINAL_STATES) irrespective of the transciver type
+    Returns:
+        True if the port is in CMIS initialization process, otherwise False
+    """
+    def is_port_in_cmis_terminal_state(self, logical_port_name):
```
prgeor, line 1636: @mihirpat1 instead of passing cmismgr to dom thread, can we post the CMIS state to `TRANSCEIVER_STATUS` table and use this API to fetch it from that table? this will also help in cases to know the current SW state machine state

mihirpat1: I have now replaced using cmismgr with updating `TRANSCEIVER_STATUS ` table to communicate the `cmis_state` between the threads.

prgeor:
@keboliu @Junchao-Mellanox can you review this PR and see if it does not impact your paltform that skip's CMIS manager

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -697,6 +714,35 @@ def update_port_transceiver_status_table_sw(logical_port_name, status_tbl, statu
     fvs = swsscommon.FieldValuePairs([('status', status), ('error', error_descriptions)])
     status_tbl.set(logical_port_name, fvs)
 
+def update_port_transceiver_status_table_sw_cmis_state(lport, xcvr_table_helper, port_mapping, cmis_state_to_set):
```
prgeor, line 717: @mihirpat1 any specific reason to define outside CmisManagerTask? if not, can we keep this inside CMIS task class

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1574,14 +1605,15 @@ def join(self):
...
-    def __init__(self, namespaces, port_mapping, main_thread_stop_event):
+    def __init__(self, namespaces, port_mapping, main_thread_stop_event, skip_cmis_mgr):
         threading.Thread.__init__(self)
         self.name = "DomInfoUpdateTask"
         self.exc = None
         self.task_stopping_event = threading.Event()
         self.main_thread_stop_event = main_thread_stop_event
         self.port_mapping = copy.deepcopy(port_mapping)
         self.namespaces = namespaces
+        self.skip_cmis_mgr = skip_cmis_mgr
```
prgeor, line 1616: @mihirpat1 I am not sure if this is a must required because platform which skip cmis mgr will return CMIS_STATE_UNKNOWN which means `is_port_in_cmis_initialization_process()` will return False always (even though the fw could be initializing the CMIS module like in SN 2700)

mihirpat1: @prgeor The function `is_port_in_cmis_initialization_process` currently assumes that `CMIS_STATE_UNKNOWN` is a transient state and hence, returns True.
For platforms which have CMIS manager enabled, we assume that `CMIS_STATE_UNKNOWN` is the first state after starting CMIS state machine post xcvrd boot-up. Hence, the DomInfoUpdateTask thread holds off polling on the corresponding port to allow the port go through CMIS initialization if needed.

sonic-xcvrd/xcvrd/xcvrd.py
mihirpat1: I have moved this to CmisManagerTask now.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1252,6 +1252,10 @@ def task_worker(self):
         for namespace in self.namespaces:
             self.wait_for_port_config_done(namespace)
 
+        logical_port_list = self.port_mapping.logical_port_list
+        for lport in logical_port_list:
+            self.update_port_transceiver_status_table_sw_cmis_state(lport, CMIS_STATE_UNKNOWN)
```
prgeor, line 1257: @mihirpat1 do we need to take care of Xcvrd restart?

mihirpat1: We don't need to explicitly handle xcvrd restart case since the below code will set the `cmis_state` to `CMIS_STATE_READY` and prevent link flap if the module is already initialized.
https://github.com/sonic-net/sonic-platform-daemons/blob/d513eca54f27c15fa384c927a5fd0f01be1353e3/sonic-xcvrd/xcvrd/xcvrd.py#L1390

Also, the current nightly behavior is to clear `cmis_state` after xcvrd restart. Hence, this should not cause additional change in behavior.

PR #467, opened 2024-04-15
Added SmartSwitch support in chassisd and enabling chassisd 

rameshraghupathy:
Added SmartSwitch support in chassisd and enabling chassisd for fixed SmartSwitches

#### Description
chassisd is enabled only for modular chassis.  Smartswitch is a fixed chassis.  However it has been treated like a modular chassis to manage the DPU cards just like the line-cards of a modular chassis.  chassisd will be enabled only on the smartswitch NPU and hence the scope of these changes are limited only to NPU and not applicable to DPU.

#### Motivation and Context
chassisd is enabled only for modular chassis.  Smartswitch is a fixed chassis.  However it has been treated like a modular chassis to manage the DPU cards just like the line-cards of a modular chassis.  Hence, chassid is needed for SmartSwitch and enabled here.  Also, some of the table updates and clean up are not required for SmartSwitch platform and hence using the is_smartswitch API to selectively enable it. the text "fixes #xxxx", "closes #xxxx" or "resolves #xxxx" here

1. Enabled and built and image and checked if the chassisd is running all the time without crashing
2. Verify if the config change handler is working as expected by issuing config CLIs to startup and shutdown DPUs


sonic-chassisd/scripts/chassisd
```diff
@@ -158,7 +161,20 @@ class ModuleConfigUpdater(logger.Logger):
         if (admin_state == MODULE_ADMIN_DOWN) or (admin_state == MODULE_ADMIN_UP):
             # Setting the module to administratively up/down state
             self.log_info("Changing module {} to admin {} state".format(key, 'DOWN' if admin_state == MODULE_ADMIN_DOWN else 'UP'))
-            try_get(self.chassis.get_module(module_index).set_admin_state, admin_state, default=False)
+            # Acquire the lock before submitting the callback function
+            with self.lock:
```
oleksandrivantsiv, line 165: Can you please explain the logic here? Why the thread is needed? 
The lock is taken in the parent thread and is unlocked immediately after the thread is started. It is not available for the child threads and can't protect data.

oleksandrivantsiv (changes requested):

sonic-chassisd/scripts/chassisd
```diff
@@ -568,27 +600,23 @@ class ChassisdDaemon(daemon_base.DaemonBase):
...
-            self.log_error("Chassisd not supported for this platform")
-            sys.exit(CHASSIS_NOT_SUPPORTED)
-
-        # Start configuration manager task on supervisor module
-        if self.module_updater.supervisor_slot == self.module_updater.my_slot:
-            config_manager = ConfigManagerTask()
-            config_manager.task_run()
-        else:
-            config_manager = None
+        config_manager = ConfigManagerTask()
```
oleksandrivantsiv, line 617: ConfigManagerTask is derived from the ProcessTaskBase class. This class runs the task_worker in a separate process. Why is the process replaced with the thread? In Python due to the GIL limitation, it is preferable to use processes instead of threads.

sonic-chassisd/scripts/chassisd
rameshraghupathy: This is not required. Reverted the change.

sonic-chassisd/scripts/chassisd
rameshraghupathy: This is also not required.  Reverted the change.

sonic-chassisd/scripts/chassisd
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
```
oleksandrivantsiv, line 1: Why was this changed? #!/usr/bin/env python3 is the proper way to execute Python.

sonic-chassisd/scripts/chassisd
```diff
@@ -138,6 +138,7 @@ class ModuleConfigUpdater(logger.Logger):
         super(ModuleConfigUpdater, self).__init__(log_identifier)
 
         self.chassis = chassis
+        self.lock = threading.Lock()
```
oleksandrivantsiv, line 141: Is this needed?

sonic-chassisd/scripts/chassisd
```diff
@@ -296,37 +305,34 @@ class ModuleUpdater(logger.Logger):
...
-                fvs = self.hostname_table.get(key)
-                if isinstance(fvs, list) and fvs[0] is True:
-                    fvs = dict(fvs[-1])
-                    hostname = fvs[CHASSIS_MODULE_INFO_HOSTNAME_FIELD]
-                    down_module_key = key+'|'+hostname
-                else:
-                    down_module_key = key+'|'
-
-                if module_info_dict[CHASSIS_MODULE_INFO_OPERSTATUS_FIELD] != str(ModuleBase.MODULE_STATUS_ONLINE):
-                    if prev_status == ModuleBase.MODULE_STATUS_ONLINE:
```
oleksandrivantsiv, line 309: Why was prev_status removed? This is legacy code that is not executed for the smart switch.

oleksandrivantsiv:
What are the states supported by the DPUs in the Smart Switch?

sonic-chassisd/scripts/chassisd
rameshraghupathy: This not removed, but the indentation has changed as we are checking for is_smartswitch().

sonic-chassisd/scripts/chassisd
rameshraghupathy: Good catch! Looks like I forgot to clean it up.

sonic-chassisd/scripts/chassisd
rameshraghupathy: Not supposed to be.  Restored.

rameshraghupathy:
> What are the states supported by the DPUs in the Smart Switch?

”dpu_midplane_link_state”
”dpu_control_plane_state"
"dpu_data_plane_state"

sonic-chassisd/scripts/chassisd
vvolam: Even I think prev_status seems to be removed and not used.

sonic-chassisd/scripts/chassisd
```diff
@@ -296,37 +304,34 @@ class ModuleUpdater(logger.Logger):
...
                             self.down_modules[down_module_key] = {}
                             self.down_modules[down_module_key]['down_time'] = time.time()
                             self.down_modules[down_module_key]['cleaned'] = False
-                    continue
-                else:
-                    # Module is operational. Remove it from down time tracking.
-                    if down_module_key in self.down_modules:
-                        self.log_notice("Module {} recovered on-line!".format(key))
-                        del self.down_modules[down_module_key]
-                    elif prev_status != ModuleBase.MODULE_STATUS_ONLINE:
```
vvolam, line 328: Both these lines also got removed looks like.

rameshraghupathy: Yes, Good catch!

sonic-chassisd/tests/test_chassisd.py
```diff
@@ -703,6 +706,8 @@ def test_daemon_run_linecard():
        daemon_chassisd.run()
 
 def test_chassis_db_cleanup():
+    # Fix Me
+    return
```
oleksandrivantsiv, line 710: ?

sonic-chassisd/scripts/chassisd
```diff
@@ -184,7 +188,7 @@ class ModuleUpdater(logger.Logger):
         self.chassis = chassis
```
oleksandrivantsiv, line 188: I think that instead of modifying the existing `ModuleUpdater` class, we should implement `SmartSwitchModuleUpdater(ModuleUpdater)`. `SmartSwitchModuleUpdater` should be derived from `ModuleUpdater` and overwrite the methods that should behave differently for the Smart Switch. This approach allows us to keep the original implementation untouched and guarantees full backward compatibility with the chassisd.

sonic-chassisd/tests/test_chassisd.py
```diff
@@ -686,13 +686,16 @@ def test_signal_handler():
...
     daemon_chassisd = ChassisdDaemon(SYSLOG_IDENTIFIER)
     daemon_chassisd.stop = MagicMock()
     daemon_chassisd.stop.wait.return_value = True
     daemon_chassisd.run()
 
 def test_daemon_run_linecard():
     # Test the chassisd run
+    # Fix Me
+    return
```
oleksandrivantsiv, line 698: ?

oleksandrivantsiv (changes requested):
as commented

sonic-chassisd/tests/test_chassisd.py
rameshraghupathy: This has dependency on is_smartswitch  defined in "https://github.com/sonic-net/sonic-platform-common/pull/454" to be committed.  Will unto this once this PR is committed.

sonic-chassisd/tests/test_chassisd.py
```diff
@@ -686,13 +686,16 @@ def test_signal_handler():
 
 def test_daemon_run_supervisor():
     # Test the chassisd run
+    return
```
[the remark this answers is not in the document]
rameshraghupathy: This has dependency on is_smartswitch  defined in "https://github.com/sonic-net/sonic-platform-common/pull/454" to be committed.  Will unto this once this PR is committed.

sonic-chassisd/tests/test_chassisd.py
rameshraghupathy: This has dependency on is_smartswitch  defined in "https://github.com/sonic-net/sonic-platform-common/pull/454" to be committed.  Will unto this once this PR is committed.

sonic-chassisd/scripts/chassisd
rameshraghupathy: Will consider this

sonic-chassisd/tests/test_chassisd.py
rameshraghupathy: restored

sonic-chassisd/tests/test_chassisd.py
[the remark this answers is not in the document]
rameshraghupathy: restored

sonic-chassisd/tests/test_chassisd.py
rameshraghupathy: restored

sonic-chassisd/scripts/chassisd
rameshraghupathy: Implemented this change.

sonic-chassisd/scripts/chassisd
oleksandrivantsiv: This is still not resolved...

rameshraghupathy: It works without that.  However, the config change notification is sent sequentially to each DPU and takes severalX time depending on the number of DPUs.  As we use this mainly for time consuming startup/shutdown it is very slow.  Having this concurrent triggering solves the issue.  If you see any issue we can fix it.  Let me know if there is a better way to solve this.  I'm open to your suggestion.

sonic-chassisd/scripts/chassisd
rameshraghupathy: @oleksandrivantsiv

prgeor:
@rameshraghupathy can you point to the Yang model change for `admin_status` for DPU in config DB table?

prgeor (commented):
@rameshraghupathy can we add monitoring of the DPU midplane in Chassisd?

prgeor:
@rameshraghupathy Please specify in the PR description if this chassisd changes is limited to NPU or DPU or both

sonic-chassisd/scripts/chassisd
```diff
@@ -559,6 +612,161 @@ class ModuleUpdater(logger.Logger):
...
                     self.down_modules[module]['cleaned'] = True
 
+#
+# Module Updater ==============================================================
+#
+
+
+class SmartSwitchModuleUpdater(ModuleUpdater):
+
+    def __init__(self, log_identifier, chassis, my_slot, supervisor_slot):
```
prgeor, line 622: @rameshraghupath not sure what is the use of slot in SMART switch context. can we remove?

sonic-chassisd/scripts/chassisd
```diff
@@ -559,6 +612,161 @@ class ModuleUpdater(logger.Logger):
...
+        if not self.midplane_initialized:
+            return
+
+        index = -1
+        for module in self.chassis.get_all_modules():
+            index += 1
+
+            module_key = try_get(module.get_name, default='MODULE {}'.format(index))
+            midplane_ip = try_get(module.get_midplane_ip, default=INVALID_IP)
+            midplane_access = try_get(module.is_midplane_reachable, default=False)
```
prgeor, line 737: @rameshraghupathy I remember midplane IP is captured in NPU config_db.json as per this 
 HLD https://github.com/Pterosaur/SONiC/blob/master/doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md

image

Can we use one source of truth for DPU midplane IP. ? i.e fetch the DPU midplane ip from the config_db.json? Since the IP address is fixed , chassisd can fetch the IP once during init

sonic-chassisd/scripts/chassisd
```diff
@@ -559,6 +612,161 @@ class ModuleUpdater(logger.Logger):
...
+        self.chassis_app_db_clean_sha = None
+
+        self.linecard_reboot_timeout = DEFAULT_LINECARD_REBOOT_TIMEOUT
+        if os.path.isfile(PLATFORM_ENV_CONF_FILE):
+            with open(PLATFORM_ENV_CONF_FILE, 'r') as file:
+                for line in file:
+                    field = line.split('=')[0].strip()
+                    if field == "linecard_reboot_timeout":
+                        self.linecard_reboot_timeout = int(line.split('=')[1].strip())
+
```
prgeor, line 658: @rameshraghupathy line card reboot .? I am not aware of this in smart switch HLD

sonic-chassisd/scripts/chassisd
```diff
@@ -559,6 +612,161 @@ class ModuleUpdater(logger.Logger):
...
+        self.chassis_state_db = daemon_base.db_connect("CHASSIS_STATE_DB")
+
+        self.hostname_table = swsscommon.Table(self.chassis_state_db, CHASSIS_MODULE_HOSTNAME_TABLE)
+        self.module_reboot_table = swsscommon.Table(self.chassis_state_db, CHASSIS_MODULE_REBOOT_INFO_TABLE)
+        self.down_modules = {}
+        self.chassis_app_db_clean_sha = None
+
+        self.linecard_reboot_timeout = DEFAULT_LINECARD_REBOOT_TIMEOUT
+        if os.path.isfile(PLATFORM_ENV_CONF_FILE):
```
prgeor, line 652: @rameshraghupathy not aligned to HLD. What is the use case for this file?

sonic-chassisd/scripts/chassisd
```diff
@@ -559,6 +612,161 @@ class ModuleUpdater(logger.Logger):
...
+                    self.log_error("Incorrect module-name {}. Should start with {} ".format(key,
+                                                            ModuleBase.MODULE_TYPE_DPU))
+                    continue
+
+                fvs = swsscommon.FieldValuePairs([(CHASSIS_MODULE_INFO_DESC_FIELD, module_info_dict[CHASSIS_MODULE_INFO_DESC_FIELD]),
+                                                  (CHASSIS_MODULE_INFO_SLOT_FIELD,
+                                                   module_info_dict[CHASSIS_MODULE_INFO_SLOT_FIELD]),
+                                                  (CHASSIS_MODULE_INFO_OPERSTATUS_FIELD, module_info_dict[CHASSIS_MODULE_INFO_OPERSTATUS_FIELD]),
+                                                  (CHASSIS_MODULE_INFO_NUM_ASICS_FIELD, str(len(module_info_dict[CHASSIS_MODULE_INFO_ASICS]))),
+                                                  (CHASSIS_MODULE_INFO_SERIAL_FIELD, module_info_dict[CHASSIS_MODULE_INFO_SERIAL_FIELD])])
```
prgeor, line 696: @rameshraghupathy can you point to which schema its follows in HLD? if not can we remove?

sonic-chassisd/scripts/chassisd
```diff
@@ -559,6 +612,161 @@ class ModuleUpdater(logger.Logger):
...
+        asics = try_get(self.chassis.get_module(module_index).get_all_asics,
+                        default=[])
+        serial = try_get(self.chassis.get_module(module_index).get_serial)
+
+        module_info_dict[CHASSIS_MODULE_INFO_NAME_FIELD] = name
+        module_info_dict[CHASSIS_MODULE_INFO_DESC_FIELD] = str(desc)
+        module_info_dict[CHASSIS_MODULE_INFO_SLOT_FIELD] = str(slot)
+        module_info_dict[CHASSIS_MODULE_INFO_OPERSTATUS_FIELD] = str(status)
+        module_info_dict[CHASSIS_MODULE_INFO_ASICS] = asics
+        module_info_dict[CHASSIS_MODULE_INFO_SERIAL_FIELD] = str(serial)
```
prgeor, line 720: @rameshraghupathy where are we planning to use this? Is this captured in HLD?

rameshraghupathy:
> @rameshraghupathy can we add monitoring of the DPU midplane in Chassisd?

This is already part of chassisd  -> check_midplane_reachability

rameshraghupathy:
> @rameshraghupathy can you point to the Yang model change for `admin_status` for DPU in config DB table? https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-chassis-module.yang#L23 needs update for DPU

I will create a PR for this and send it out soon

rameshraghupathy:
> @rameshraghupathy Please specify in the PR description if this chassisd changes is limited to NPU or DPU or both

Done
@prgeor

sonic-chassisd/scripts/chassisd
rameshraghupathy: > @rameshraghupath not sure what is the use of slot in SMART switch context. can we remove?

Resolved
@prgeor

sonic-chassisd/scripts/chassisd
rameshraghupathy: Always config_db.json is the single source of truth.  The module also gets it from there.  That's why I didn't change this code.
@prgeor

sonic-chassisd/scripts/chassisd
rameshraghupathy: Good catch!  Cleaned it!
@pr

sonic-chassisd/scripts/chassisd
rameshraghupathy: The "CHASSIS_MODULE_INFO_SERIAL_FIELD" and "CHASSIS_MODULE_INFO_OPERSTATUS_FIELD" are all used in "show chassis modules status" and this CLI output is referred in the HLD
root@sonic:~#show chassis modules status                                                                                      
Name        Description         ID         Oper-Status     Admin-Status     Serial

DPU0        SS-DPU0             1           Online          up              SN20240105
DPU1        SS-DPU1             2           Online          up              SN20240102

sonic-chassisd/scripts/chassisd
rameshraghupathy: > @rameshraghupathy where are we planning to use this? Is this captured in HLD?

The "CHASSIS_MODULE_INFO_SERIAL_FIELD" and "CHASSIS_MODULE_INFO_OPERSTATUS_FIELD" are all used in "show chassis modules status" and this CLI output is referred in the HLD
root@sonic:~#show chassis modules status                                                                                      
Name        Description         ID         Oper-Status     Admin-Status     Serial

DPU0        SS-DPU0             1           Online          up              SN20240105
DPU1        SS-DPU1             2           Online          up              SN20240102

I will remove "CHASSIS_MODULE_INFO_ASICS"

sonic-chassisd/scripts/chassisd
```diff
@@ -649,19 +955,27 @@ class ChassisdDaemon(daemon_base.DaemonBase):
...
-            sys.exit(CHASSIS_NOT_SUPPORTED)
+        if not self.smartswitch:
+            if ((self.module_updater.my_slot == INVALID_SLOT) or
+                    (self.module_updater.supervisor_slot == INVALID_SLOT)):
+                self.log_error("Chassisd not supported for this platform")
+                sys.exit(CHASSIS_NOT_SUPPORTED)
 
         # Start configuration manager task on supervisor module
-        if self.module_updater.supervisor_slot == self.module_updater.my_slot:
+        if self.smartswitch:
```
prgeor, line 975: @rameshraghupathy comment above is incorrect

sonic-chassisd/scripts/chassisd
```diff
@@ -603,6 +808,52 @@ class ConfigManagerTask(ProcessTaskBase):
...
+            if state == swsscommon.Select.TIMEOUT:
+                # Do not flood log when select times out
+                continue
+            if state != swsscommon.Select.OBJECT:
+                self.logger.log_warning("sel.select() did not return swsscommon.Select.OBJECT")
+                continue
+
+            (key, op, fvp) = sst.pop()
+
+            if op == 'SET':
```
prgeor, line 848: @rameshraghupathy SET operation can set ADMIN down or UP. please parse the key to know if the user is setting admin UP/DOWN

sonic-chassisd/scripts/chassisd
```diff
@@ -603,6 +808,52 @@ class ConfigManagerTask(ProcessTaskBase):
...
+                admin_state = MODULE_ADMIN_DOWN
+            elif op == 'DEL':
```
prgeor, line 850: @rameshraghupathy on DEL of the admin state why do module ADMIN UP?

sonic-chassisd/scripts/chassisd
rameshraghupathy: @prgeor Fixed

sonic-chassisd/scripts/chassisd
rameshraghupathy: > @rameshraghupathy on DEL of the admin state why do module ADMIN UP?

@prgeor The original behavior is: The default is no config - meaning admin up   when the user wants to shutdown a module adds a config "admins_status: down".  So, the SET operation has been used for "admins_status: down" and DEL has been used for setting the module up.   If you want we can change this for Smartswitch.  Please let me know.

sonic-chassisd/scripts/chassisd
rameshraghupathy: > @rameshraghupathy SET operation can set ADMIN down or UP. please parse the key to know if the user is setting admin UP/DOWN

@prgeor Good question.  The original behavior is: The default is no config - meaning admin up   when the user wants to shutdown a module adds a config "admins_status: down".  So, the SET operation has been used for "admins_status: down" and DEL has been used for setting the module up.   If you want we can change this for Smartswitch.  Please let me know.

sonic-chassisd/scripts/chassisd
```diff
@@ -73,6 +73,7 @@ CHASSIS_MODULE_REBOOT_INFO_TABLE = 'CHASSIS_MODULE_REBOOT_INFO_TABLE'
 CHASSIS_MODULE_REBOOT_TIMESTAMP_FIELD = 'timestamp'
 CHASSIS_MODULE_REBOOT_REBOOT_FIELD = 'reboot'
 DEFAULT_LINECARD_REBOOT_TIMEOUT = 180
+DEFAULT_DPU_REBOOT_TIMEOUT = 180
```
vvolam, line 76: Is 180 secs sufficient for entire DPU reboot time?

vvolam: Also, should we add this to platform.json as vendor configurable timeout?

sonic-chassisd/scripts/chassisd
```diff
@@ -167,6 +168,59 @@ class ModuleConfigUpdater(logger.Logger):
...
+        # on multiple DPUs for faster bring-up and shutdown.
+        self.lock = threading.Lock()
+
+    def deinit(self):
+        """
+        Destructor of SmartSwitchModuleConfigUpdater
+        :return:
+        """
+
+    def module_config_update(self, key, admin_state):
```
vvolam, line 196: @rameshraghupathy could you please add a quick summary of function here and also remaining functions for future clarity?

sonic-chassisd/scripts/chassisd
```diff
@@ -167,6 +168,59 @@ class ModuleConfigUpdater(logger.Logger):
...
+        Constructor for SmartSwitchModuleConfigUpdater
+        :param chassis: Object representing a platform chassis
+        """
+        super(SmartSwitchModuleConfigUpdater, self).__init__(log_identifier)
+
+        self.chassis = chassis
+        # The threading.Lock is required for concurrent setting of admin_state
+        # on multiple DPUs for faster bring-up and shutdown.
+        self.lock = threading.Lock()
+
```
prgeor, line 189: @rameshraghupathy which data is be protected against multiple thread access?

sonic-chassisd/scripts/chassisd
prgeor: @rameshraghupathy we need to handle following scenarios:-

1. At run time if user configures module up/down then process the event
2. When chassisd is booting up and if the admin status is missing in config DB then process to keep the DPU in dark mode. In this case there will not be any event triggered. Its a DB read. 
3. While chassisd bootup if the config DB already has admin UP/DOWN then process accordingly

@rameshraghupathy please make sure if chasssid crashes at run time it does not bring down the DPU if DPU was already admin up.

sonic-chassisd/scripts/chassisd
prgeor: @rameshraghupathy didn't we agree to keep the default behavior on missing of the config DB admin status to have dark mode enabled as default? @vvolam is that not captured in the HLD?

sonic-chassisd/scripts/chassisd
```diff
@@ -630,6 +881,61 @@ class ChassisdDaemon(daemon_base.DaemonBase):
         else:
             self.log_warning("Caught unhandled signal '{}' - ignoring...".format(SIGNALS_TO_NAMES_DICT[sig]))
 
+    def set_initial_dpu_admin_state(self):
+        """Send admin_state trigger once to modules those are powered up"""
+        lock = threading.Lock()
+        threads = []
+        for module_index in range(0, self.module_updater.num_modules):
```
prgeor, line 888: @rameshraghupathy I am not sure why we cannot handle all DPUs in one single thread.

rameshraghupathy: > @rameshraghupathy I am not sure why we cannot handle all DPUs in one single thread.
@prgeor It is not that we can't.  We can but the call goes sequentially to each DPU and power cycling is a time consuming process and each DPU requires roughly 3 mins. So to power cycle a chassis we need 24+ mins.  When we fork all 8 finishes in 3 mins

sonic-chassisd/scripts/chassisd
rameshraghupathy: > @rameshraghupathy we need to handle following scenarios:-
> 
> 1. At run time if user configures module up/down then process the event
> 2. When chassisd is booting up and if the admin status is missing in config DB then process to keep the DPU in dark mode. In this case there will not be any event triggered. Its a DB read.
> 3. While chassisd bootup if the config DB already has admin UP/DOWN then process accordingly
> 
> @rameshraghupathy please make sure if chasssid crashes at run time it does not bring down the DPU if DPU was already admin up.

@prgeor Support for (1) exists in the original implementation. (2) and (3) are handled by  "def set_initial_dpu_admin_state(self):"

rameshraghupathy: > @rameshraghupathy please make sure if chasssid crashes at run time it does not bring down the DPU if DPU was already admin up.

@prgeor Already verified this case

sonic-chassisd/scripts/chassisd
oleksandrivantsiv: @rameshraghupathy this comment is still not resolved. Please do not mark it as resolved.

The lock is taken before the thread is stated. The thread doesn't take the lock at all.  Why do we need to use the lock here at all? What resource are you trying to protect?

sonic-chassisd/scripts/chassisd
```diff
@@ -167,6 +168,59 @@ class ModuleConfigUpdater(logger.Logger):
...
+            self.log_error("Unable to get module-index for key {} to set admin-state {}". format(key, admin_state))
+            return
+
+        if (admin_state == MODULE_ADMIN_DOWN) or (admin_state == MODULE_ADMIN_UP):
+            self.log_info("Changing module {} to admin {} state".format(key, 'DOWN' if admin_state == MODULE_ADMIN_DOWN else 'UP'))
+            # Acquire the lock before submitting the callback function
+            with self.lock:
+                # Submit the callback function as a separate thread. It is the responsibility
+                # of the module to handle multiple b2b requests to the same module
+                t = threading.Thread(target=self.submit_callback, args=(module_index, admin_state))
```
oleksandrivantsiv, line 215: Duplicating the comment since it was marked as resolved:
image

rameshraghupathy: The lock is valid until the start of the thread. It is safe to have that as multiple threads may access common resources in chassis and modules causing data corruption. Once the thread is created it has its own stack. In fact it is good to have a lock inside the submit_callback.  It is hard to find what common resources are accessed inside each vendor API. So, it is better to have a lock even in submit_callback.

sonic-chassisd/scripts/chassisd
rameshraghupathy: @prgeor It is hard to predict and say what vendor API would do under the "submit_callback". When multiple threads are forked chassis/modules resources can be corrupted. One thread can increase the chassis.number_of_modules and the other can decrease it causing data corruption and unpredictable states and behavior.  In fact it is good have another one in submit_callback as well.  But I haven't tested putting a lock in submit_callback.

sonic-chassisd/scripts/chassisd
rameshraghupathy: This may not be sufficient for a reboot as I told you.  This is good for shutdown. 
Here is what may work:
shutdown: pre_shutdown_time + graceful shutdown time + 180s for proper HW shutdown (240s is safe)
reboot: shutdown (240s) + hw off-on  requirement if any (say 30s) + 180s for power on

rameshraghupathy: A vendor configurable option is a good thing.

rameshraghupathy: We can pick a number now and optimize it as we move.

sonic-chassisd/scripts/chassisd
```diff
@@ -630,6 +913,61 @@ class ChassisdDaemon(daemon_base.DaemonBase):
...
+                    if admin_state != 'down' and midplane_state == 'False':
+                        # startup DPU
+                        op = MODULE_ADMIN_UP
+
+                    if op is not None:
+                        # Spawn a thread to handle the DPU logic
+                        def handle_dpu_logic():
+                            try:
+                                with lock:
+                                    module = self.module_updater.chassis.get_module(module_index)
```
gpunathilell, line 950: The `module_index` and `op` variables here are module specific, so these variables should be function arguments for `handle_dpu_logic`. With the current implementation there is a good probability that the `module_index` and `op` are already updated by the time a thread has started for the previous module, Example scenario:
`DPU1 : op = MODULE_ADMIN_UP and module_index = 1` thread.start() is called (but thread has still not started yet or lock is not yet obtained - > this is thread1)
We move to the next element in the for loop in `for module_index in range(0, self.module_updater.num_modules):`
`DPU2 : op = MODULE_ADMIN_DOWN and module_index = 2` thread.start() is called -> this is thread2.
At this point if thread1 starts executing the DPU2 related admin down operation is handled while the DPU1 admin up operation is skipped (Since `op` and `module_index` is already updated.

sonic-chassisd/scripts/chassisd
```diff
@@ -630,6 +913,61 @@ class ChassisdDaemon(daemon_base.DaemonBase):
...
+            midplane_state = 'False'
+            if isinstance(fvs, list) and fvs[0] is True:
+                fvs = dict(fvs[-1])
+                midplane_state = fvs[CHASSIS_MIDPLANE_INFO_ACCESS_FIELD]
+
+            try:
+                if module_info_dict is not None:
+                    # Get admin state of DPU
+                    key = module_info_dict[CHASSIS_MODULE_INFO_NAME_FIELD]
+                    admin_state = self.module_updater.get_module_admin_status(key)
```
gpunathilell, line 937: As per [this comment](https://github.com/sonic-net/sonic-platform-daemons/pull/467/files#r1796139055) and [this comment](https://github.com/sonic-net/sonic-platform-daemons/pull/467/files#r1796155286) the `set_initial_dpu_admin_state` is supposed to handle the dark mode requirement (If there is no CONFIG_DB `CHASSIS_MODULE` entries for the DPU then the admin state for the DPU should be down) but in this function, the `get_module_admin_status` [would return `'up'`](https://github.com/sonic-net/sonic-platform-daemons/blob/2ae658dbc20fe7d8862fd0319e1cf78151aec105/sonic-chassisd/scripts/chassisd#L357) by default if there is no CONFIG_DB entry, meaning `op = 'up'` which would power on the dpu, but by default the DPU should be powered off by default if there is no CONFIG_DB entry

sonic-chassisd/scripts/chassisd
oleksandrivantsiv: @rameshraghupathy during the HDL review sessions we agreed that this value is vendor specific and should be taken from `platform.json`. Please change the implementation based on our agreement

rameshraghupathy: Yes, agree, but the existing modular-chassis LC use "platform_env.conf". Shall we be consistent with that?
PLATFORM_ENV_CONF_FILE = "/usr/share/sonic/platform/platform_env.conf"
        self.linecard_reboot_timeout = DEFAULT_LINECARD_REBOOT_TIMEOUT
        if os.path.isfile(PLATFORM_ENV_CONF_FILE):
            with open(PLATFORM_ENV_CONF_FILE, 'r') as file:
                for line in file:
                    field = line.split('=')[0].strip()
                    if field == "linecard_reboot_timeout":
                        self.linecard_reboot_timeout = int(line.split('=')[1].strip())

sonic-chassisd/tests/mock_platform.py
```diff
@@ -75,13 +75,25 @@ def set_midplane_reachable(self, up):
     def get_all_asics(self):
         return self.asic_list
 
+    def get_reboot_cause(self):
+        reboot_cause = {
```
oleksandrivantsiv, line 79: Where does this schema come from? This not what is written in the HLD:
```
        Retrieves the cause of the previous reboot of the DPU module

        Returns:
            A tuple (string, string) where the first element is a string containing the cause of the previous reboot. This string must be one of the predefined strings in this class. If the first string is "REBOOT_CAUSE_HARDWARE_OTHER", the second string can be used to pass a description of the reboot cause.
```

oleksandrivantsiv: https://github.com/sonic-net/SONiC/blob/master/doc/smart-switch/pmon/smartswitch-pmon.md#get-dpu-reboot-cause

sonic-chassisd/scripts/chassisd
```diff
@@ -559,6 +608,184 @@ class ModuleUpdater(logger.Logger):
...
+                                                  (CHASSIS_MODULE_INFO_SLOT_FIELD,
+                                                   module_info_dict[CHASSIS_MODULE_INFO_SLOT_FIELD]),
+                                                  (CHASSIS_MODULE_INFO_OPERSTATUS_FIELD, module_info_dict[CHASSIS_MODULE_INFO_OPERSTATUS_FIELD]),
+                                                  (CHASSIS_MODULE_INFO_SERIAL_FIELD, module_info_dict[CHASSIS_MODULE_INFO_SERIAL_FIELD])])
+                current_status = self.get_module_current_status(key)
+                if self.prev_status[module_index] == ModuleBase.MODULE_STATUS_OFFLINE:
+                    if current_status != ModuleBase.MODULE_STATUS_OFFLINE:
+                        # DPU module has been turned on, update the reboot-cause
+                        # If the reboot-cause file not already present save it
+                        # If the STATE_DB REBOOT_CAUSE table already doesn't have
```
oleksandrivantsiv, line 703: Why STATE_DB? Should it be CHASSIS_STATE_DB?

sonic-chassisd/scripts/chassisd
```diff
@@ -559,6 +608,184 @@ class ModuleUpdater(logger.Logger):
...
+                        # an entry for the recent reboot update the DB
+                        reboot_cause = try_get(self.chassis.get_module(module_index).get_reboot_cause)
```
oleksandrivantsiv, line 705: Where is the reboot cause populated in the CHASSIS_DB? I don't see the implementation for this. The reboot cause is read into a local variable and that is it. Nothing else is done with the variable

oleksandrivantsiv:
Please implement a unit test for reboot_cause implementation

oleksandrivantsiv:
@rameshraghupathy, @prgeor please extend the HLD with the flows for reboot_case and reboot_case_history. From this implementation, it is not clear how it works for the platforms that support the `get_reboot_causes` API without information in the CHASSIS_STATE_DB

sonic-chassisd/scripts/chassisd
```diff
@@ -746,26 +1228,84 @@ class ChassisdDaemon(daemon_base.DaemonBase):
...
+                        # startup DPU
+                        op = MODULE_ADMIN_UP
+
+                    elif admin_state != 'up' and midplane_state == 'True':
+                        # shutdown DPU
+                        op = MODULE_ADMIN_DOWN
+
+                    # Initialize DPU_STATE DB table on bootup
+                    dpu_state_key = "DPU_STATE|" + module_name
+                    self.module_updater.update_dpu_state(dpu_state_key, admin_state.upper())
```
gpunathilell, line 1266: Why do we update the midplane state with the configured `admin_state` instead of checking the reachability of the midplane here?

oleksandrivantsiv: We need to query the operation state of the interface connected to the DPU. And if the oper state is up, the we should test the reachability of IP address of the interface on the DPU side

rameshraghupathy: During the initialization of the system we want to make sure the system comes up correctly with the configured state, which is the admin_state.  During boot up the operational state of the DPUs could be anything. So, 
1. Get the admin_state and initialize the DPUs to that state
2. Set the DPU_STATE table as per the admins state
3. Soon the periodic update of the DPU_STATE will kick which will check for the operational state of the DPU and update it

gpunathilell: Assuming before we call the `set_initial_dpu_admin_state` the DPUs are powered off but the CONFIG_DB entry is set to `admin_state=up` we are setting the `dpu_midplane_link_state=up` in the CHASSIS_STATE_DB before we even power on the DPUs (which is done in the `submit_dpu_callback` later in this function), if the DPU power on fails or if we do not achieve midplane reachability we have the wrong data in the CHASSIS_STATE_DB

sonic-chassisd/scripts/chassisd
```diff
@@ -566,6 +618,389 @@ class ModuleUpdater(logger.Logger):
...
+
+            elif midplane_access is True and current_midplane_state == 'False':
+                self.log_notice("Module {} midplane connectivity is up".format(module_key))
+
+                # Update midplane state UP in the chassisStateDB DPU_STATE table
+                key = "DPU_STATE|" + module_key
+                self.update_dpu_state(key, "UP")
+
+            elif midplane_access is False and current_midplane_state == 'False':
+                if self.is_module_reboot_system_up_expired(module_key):
```
gpunathilell, line 980: Do we need this function call?

sonic-chassisd/scripts/chassisd
```diff
@@ -746,26 +1226,84 @@ class ChassisdDaemon(daemon_base.DaemonBase):
...
+            # Get midplane state of DPU
+            module_info_dict = self.module_updater._get_module_info(module_index)
+            module_name = self.platform_chassis.get_module(module_index).get_name()
+            state_db = daemon_base.db_connect("STATE_DB")
+            midplane_table = swsscommon.Table(state_db, CHASSIS_MIDPLANE_INFO_TABLE)
+            fvs = midplane_table.get(module_name)
+            midplane_state = 'False'
+            if isinstance(fvs, list) and fvs[0] is True:
+                fvs = dict(fvs[-1])
+                midplane_state = fvs[CHASSIS_MIDPLANE_INFO_ACCESS_FIELD]
```
gpunathilell, line 1247: We are checking for the `midplane_state` here in the DB (and if it is not present in the DB we are assuming that the  module is not reachable since `midplane_state=False` by default) but in case of Nvidia platform the `midplane_state` is being updated to the DB by Chassisd itself, so chassisd here is checking for midplane state in DB before it is supposed to update the midplane state itself. Either we need to re order code so that we update midplane_state to db before we check the DB for values, or instead of checking the DB we can directly invoke the midplane reachability API

sonic-chassisd/scripts/chassisd
gpunathilell: This code needs to be updated as per discussion, We are writing the admin state data (irrespective of up/down) to the CONFIG_DB from sonic-utilities, but here we are checking to see if data is set or removed (instead of checking the admin state value (as per previous discussion)

sonic-chassisd/scripts/chassisd
rameshraghupathy: @gpunathilell  Using dpu  operational_state and the API to get this as planned.

sonic-chassisd/scripts/chassisd
rameshraghupathy: @gpunathilell No we don't, cleaned it.

sonic-chassisd/scripts/chassisd
rameshraghupathy: @gpunathilell  Fixed

sonic-chassisd/scripts/chassisd
```diff
@@ -566,6 +618,398 @@ class ModuleUpdater(logger.Logger):
...
+
+        file_path = self._get_history_path(module, file_name)
+        try:
+            dt_obj = datetime.strptime(prev_reboot_time, "%Y_%m_%d_%H_%M_%S")
+        except ValueError:
+            dt_obj = datetime.now()
+
+        formatted_time = dt_obj.strftime("%a %b %d %I:%M:%S %p UTC %Y")
+
+        reboot_cause_dict = {
```
gpunathilell, line 870: The `get_reboot_cause` is supposed to return a tuple - a reboot cause and an accompanying description ([from HLD](https://github.com/sonic-net/SONiC/blob/9acaa87252412b3e07d21ec85edcd2c33f2aec3d/doc/smart-switch/pmon/smartswitch-pmon.md?plain=1#L614)) when we perform `json.dump` with a tuple, it is saved as a list can we split here as "cause" and "description" separately? otherwise the `# Publish the reboot cause information to CHASSIS_STATE_DB` step fails in chassisd because we are performing `hset` to chassis_state_db with a list

sonic-chassisd/tests/mock_platform.py
gpunathilell: Please update the return value as per previous comment, the get_reboot_cause just returns a tuple

sonic-chassisd/scripts/chassisd
```diff
@@ -566,6 +618,398 @@ class ModuleUpdater(logger.Logger):
...
+
+            # Fetch the current data for the given key and convert it to a dict
+            current_data = self._convert_to_dict(self.chassis_state_db.hgetall(key))
+
+            if current_data:
+                self.chassis_state_db.delete(key)
+
+            # Prepare the updated data
+            updates = {
+                "dpu_midplane_link_state": state,
```
gpunathilell, line 776: Formatting suggestion, Can we have the midplane state similar to admin state and have in lower case? the `show interfaces status` also displays in lower case, Can be changed later if it takes time, also the time format can be changed to   `"%Y-%m-%d %H:%M:%S"` (here or in sonic-utilities system_health command - since we are displaying this data to user) https://github.com/sonic-net/sonic-platform-daemons/blob/f371df717a261d4ffbebd2e6aedd969faad58b88/sonic-chassisd/scripts/chassisd#L1259

gpunathilell:
The REBOOT_CAUSE_DIR is not accessible from the pmon docker where chassisd is running, (theres a diffierent /host/reboot created in the docker )`REBOOT_CAUSE_DIR = "/host/reboot-cause/module/"` The first-boot file which is created by sonic-host-services is not seen by chassisd, we need to use a different location which is available in the docker, and is persistent across reboots

rameshraghupathy:
> The REBOOT_CAUSE_DIR is not accessible from the pmon docker where chassisd is running, (theres a diffierent /host/reboot created in the docker )`REBOOT_CAUSE_DIR = "/host/reboot-cause/module/"` The first-boot file which is created by sonic-host-services is not seen by chassisd, we need to use a different location which is available in the docker, and is persistent across reboots

@gpunathilell   Have you used the PR "
image

gpunathilell:
> > The REBOOT_CAUSE_DIR is not accessible from the pmon docker where chassisd is running, (theres a diffierent /host/reboot created in the docker )`REBOOT_CAUSE_DIR = "/host/reboot-cause/module/"` The first-boot file which is created by sonic-host-services is not seen by chassisd, we need to use a different location which is available in the docker, and is persistent across reboots
> 
> @gpunathilell   Have you used the PR?  This is needed for pmon docker to mount the volume.
> image
> 
No, I see that the location is mounted in pmon in that PR, will include, please ignore comment

sonic-chassisd/scripts/chassisd
rameshraghupathy: @gpunathilell Done

rameshraghupathy: @gpunathilell Done

sonic-chassisd/scripts/chassisd
rameshraghupathy: @gpunathilell Made dpu_midplane_link_state lowercase.  The time is already in the format "%Y-%m-%d %H:%M:%S" righ?

sonic-chassisd/scripts/chassisd
```diff
@@ -566,6 +618,410 @@ class ModuleUpdater(logger.Logger):
...
+                "dpu_midplane_link_reason": "",
+                "dpu_midplane_link_time": datetime.now().strftime("%Y%m%d %H:%M:%S"),
```
gpunathilell, line 778: @rameshraghupathy  I mean this line 
```suggestion
                "dpu_midplane_link_time": datetime.now().strftime("%Y-%m-%d %H:%M:%S"),
```
Will be displayed as such when we do `show system-health dpu`:
`DPU2    Online         dpu_midplane_link_state  up             2024-11-25 23:38:21`

sonic-chassisd/scripts/chassisd
gpunathilell: Resolved

sonic-chassisd/scripts/chassisd
```diff
@@ -74,7 +76,9 @@ CHASSIS_MODULE_REBOOT_INFO_TABLE = 'CHASSIS_MODULE_REBOOT_INFO_TABLE'
 CHASSIS_MODULE_REBOOT_TIMESTAMP_FIELD = 'timestamp'
 CHASSIS_MODULE_REBOOT_REBOOT_FIELD = 'reboot'
 DEFAULT_LINECARD_REBOOT_TIMEOUT = 180
+DEFAULT_DPU_REBOOT_TIMEOUT = 360
 PLATFORM_ENV_CONF_FILE = "/usr/share/sonic/platform/platform_env.conf"
+PLATFORM_JSON_FILE = "/usr/share/sonic/platform/platform.json"
```
vvolam, line 81: I see platform.json file exists in /usr/share/sonic/devices/<platform>/platform.json, am I missing something?

rameshraghupathy: @vvolam The file exists, we are accessing it here

sonic-chassisd/scripts/chassisd
```diff
@@ -92,6 +96,8 @@ INVALID_IP = '0.0.0.0'
 CHASSIS_MODULE_ADMIN_STATUS = 'admin_status'
 MODULE_ADMIN_DOWN = 0
 MODULE_ADMIN_UP = 1
+REBOOT_CAUSE_DIR = "/host/reboot-cause/module/"
```
prgeor, line 99: @rameshraghupathy rename to `MODULE_REBOOT_CAUSE_DIR` ?

rameshraghupathy: @prgeor Done

sonic-chassisd/scripts/chassisd
oleksandrivantsiv: #closed

sonic-chassisd/scripts/chassisd
rameshraghupathy: @gpunathilell done

oleksandrivantsiv (commented):
LGTM, @prgeor to review and approve

oleksandrivantsiv (approved):

prgeor:
@rameshraghupathy see conflict

kperumalbfn:
@rameshraghupathy could you check the failures

kperumalbfn:
@vvolam could you review and sign-off?

rameshraghupathy:
> @rameshraghupathy see conflict

@prgeor resolved

vvolam:
@rameshraghupathy could you resolve the build issue?

vvolam (approved):
Looks fine to me. Approving based on my knowledge.

prabhataravind:
@prgeor could you please help merge if all comments are addressed?

KrisNey-MSFT:
@prgeor - gentle ping please

PR #468, opened 2024-04-17
[ycabled] remove restriction of at least a single physical  y_cable to be present for  correct start of all  ycable threads

vdahiya12:


prgeor:
@vdahiya12 can you fix the bulid failure

PR #459, opened 2024-04-05
CMIS 'ConfigSuccess" failure while changing default ApSel code for 800G DR8/FR8 modules

AnoopKamath:
#### Description
CMIS 'ConfigSuccess" failure while changing default ApSel code for 800G DR8/FR8 modules
Issue seen with vendors:
Eoptolink
Finisar
Source-Photonics

Ex: If module supports 800G, 400G and 100G app code and has default app mode as 800G, an issue has arisen with the 2x400G target mode which is hitting ConfigRejectedPartailDataPath error and failing. (same as 8x100 App mode)

#### Motivation and Context
Extract from CMIS spec: - 6.2.4.3 Host Rules and Recommendations
 ```The host can change the width of a Data Path only while in the DPDeactivated state, i.e. the host must always transition an existing Data Path to DPDeactivated before selecting an Application with a different lane count. Any lane that becomes unused must be marked as such (AppSel = 0000b) or it must be assigned to a new valid Data Path (remaining in DPDeactivated state until eventually used) ```

Reset AppSel value for all lanes when setting non default app value

#### How Has This Been Tested?
Tested with different vendors changing different modes:

```
root@sonic:/home/cisco# show logging xcvrd | grep Ethernet160
Apr  5 17:11:02.534867 sonic WARNING pmon#xcvrd[29]: $$$ Ethernet160 handle_port_update_event() : op=SET DB:CONFIG_DB Table:PORT fvp {'admin_status': 'up', 'alias': 'etp20a', 'index': '20', 'lanes': '8,9,10,11', 'mtu': '9100', 'speed': '400000', 'subport': '1'}
Apr  5 17:11:02.537697 sonic WARNING pmon#xcvrd[29]: $$$ Ethernet160 handle_port_update_event() : op=SET DB:STATE_DB Table:PORT_TABLE fvp {'state': 'ok', 'netdev_oper_status': 'down', 'admin_status': 'up', 'mtu': '9100', 'supported_speeds': '40000,100000,200000,400000', 'supported_fecs': 'rs', 'host_tx_ready': 'true'}
Apr  5 17:11:02.538174 sonic WARNING pmon#xcvrd[29]: *** Ethernet160CONFIG_DBPORT handle_port_update_event() fvp {'admin_status': 'up', 'alias': 'etp20a', 'index': '20', 'lanes': '8,9,10,11', 'mtu': '9100', 'speed': '400000', 'subport': '1', 'key': 'Ethernet160', 'asic_id': 0, 'op': 'SET'}
[...]

PR #492, opened 2024-05-29
Fix for 'log_notifce()' typo in xcvrd.py

tshalvi:
This PR fixes the following issue: https://github.com/sonic-net/sonic-buildimage/issues/19116

#### Description
     Typo fix: log_notifce() --> log_notice()

#### Motivation and Context
PR https://github.com/sonic-net/sonic-platform-daemons/pull/459 introduces a typo "log_notifce".
Check line: https://github.com/sonic-net/sonic-platform-daemons/blob/9ffce2036d53571bd23b4a630ac285039e383c4f/sonic-xcvrd/xcvrd/xcvrd.py#L1482.

It causes an exception in cmis task manager which results in port down:

May 27 20:38:17.489639 sonic ERR pmon#xcvrd: CMIS: Ethernet96: internal errors due to 'CmisManagerTask' object has no attribute 'log_notifce'


dprital (approved):

tshalvi:
The developer who introduced the typo has already raised a PR to fix it.
Closing this PR.

PR #494, opened 2024-05-30
Don't process CMIS xcvrs on warm-reboot case

bobbymcgonigle:
#### Description
When warm-reboot happens xcvrd gets started it treats xcvrs as going through an insertion event. In the case of optics who on insertion event might need to be reprogrammed, it will force the TX off and disable them.

On warm-reboot, we don't want to do this - the xcvrs are already programmed correctly and it will impact dataplane traffic.

#### Motivation and Context
Interfaces with cmis optics in breakout configurations would stay down after warm-reboot cases in sonic-mgmt test runs. 

I tested this using a sonic-mgmt warm-reboot case where the optics would stay down and fail entire t0 runs. Those tests now pass and the interfaces stay up.

I also checked the flap counters on the peer for those interfaces - the link stayed up and did not flap - showing warm-reboot succeeded as expected.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1423,7 +1423,7 @@ class CmisManagerTask(threading.Thread):
 
                 try:
                     # CMIS state transitions
-                    if state == CMIS_STATE_INSERTED:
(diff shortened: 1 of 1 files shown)


bobbymcgonigle:
Looking into unit test issue

prgeor:
@bobbymcgonigle I think we can clos this. This is already taken care. feel free to open otherwise.

PR #496, opened 2024-06-06
Add a call to apply app0 configuration before DP INIT stage

noaOrMlnx:
depends on: https://github.com/sonic-net/sonic-platform-common/pull/474

#### Description

When NOS configures the CMIS modules, we need to be aware to a case where some of the lanes are disabled. 
According to CMIS spec, in that case, an application 0 is needed to make sure all disabled lanes are zeroed.

#### Motivation and Context

This was done according to CMIS specification document.

This addition to CMIS thread was tested with a CMIS active cable, with only 4 active lanes (and 4 de-active). 
We made sure the link is up, right places are zeroed, and have traffic. 


noaOrMlnx:
Closing since @prgeor already raised a PR to fix the same issue.

PR #500, opened 2024-06-14
[xcvrd] DPB support on platforms with CmisManagerTask enabled

ishidawataru:
#### Description

This commit fixes DPB support with CMIS transceivers.

CmisManagerTask's `port_dict` must be updated according to the port add/remove events.
This commit removes the `port_mapping` field from CmisManagerTask as
`port_mapping` was mostly used just for storing `asic_id` information
and that can be simply done by `port_dict` instead.

Added a helper method `get_asic_id()` method to CmisManagerTask for
getting `asic_id` from `logical_port`.

#### Motivation and Context

fixes https://github.com/sonic-net/sonic-buildimage/issues/18893

I tested on the VS environment.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -681,12 +681,44 @@ class CmisManagerTask(threading.Thread):
 
             self.force_cmis_reinit(lport, 0)
 
-        # PORT_DEL event for the same lport happens 3 times because
-        # we are subscribing to CONFIG_DB, STATE_DB|TRANSCEIVER_INFO, and STATE_DB|PORT_TABLE.
-        # To only handle the first one, check if the lport exists in the port_dict.
-        elif lport in self.port_dict:
-            self.update_port_transceiver_status_table_sw_cmis_state(lport, CMIS_STATE_REMOVED)
-            self.port_dict.pop(lport)
+        elif port_change_event.event_type == port_change_event.PORT_DEL:
+            # In handling the DEL event, the following two scenarios must be considered:
+            # 1. PORT_DEL event due to transceiver plug-out
+            # 2. PORT_DEL event due to Dynamic Port Breakout (DPB)
+            #
+            # Scenario 1 is simple, as only a STATE_DB|TRANSCEIVER_INFO PORT_DEL event occurs,
+            # so we just need to set SW_CMIS_STATE to CMIS_STATE_REMOVED.
+            #
+            # Scenario 2 is a bit more complex. First, for the port(s) before DPB, a CONFIG_DB|PORT PORT_DEL
+            # and a STATE_DB|PORT_TABLE PORT_DEL event occur. Next, for the port(s) after DPB,
+            # a CONFIG_DB|PORT PORT_SET and a STATE_DB|PORT_TABLE PORT_SET event occur.
+            # After that (after a short delay), a STATE_DB|TRANSCEIVER_INFO PORT_DEL event
+            # occurs for the port(s) before DPB, and finally, a STATE_DB|TRANSCEIVER_INFO
+            # PORT_SET event occurs for the port(s) after DPB.
+            #
+            # Below is the event sequence when configuring Ethernet0 from "2x200G" to "1x400G"
+            # (based on actual logs).
+            #
+            # 1.  SET Ethernet0 CONFIG_DB|PORT
+            # 2.  DEL Ethernet2 CONFIG_DB|PORT
+            # 3.  DEL Ethernet0 CONFIG_DB|PORT
+            # 4.  DEL Ethernet0 STATE_DB|PORT_TABLE
+            # 5.  DEL Ethernet2 STATE_DB|PORT_TABLE
+            # 6.  SET Ethernet0 CONFIG_DB|PORT
+            # 7.  SET Ethernet0 STATE_DB|PORT_TABLE
+            # 8.  SET Ethernet0 STATE_DB|PORT_TABLE
+            # 9.  DEL Ethernet2 STATE_DB|TRANSCEIVER_INFO
+            # 10. DEL Ethernet0 STATE_DB|TRANSCEIVER_INFO
+            # 11. SET Ethernet0 STATE_DB|TRANSCEIVER_INFO
+            #
+            # To handle both scenarios, if the lport exists in port_dict for any DEL EVENT,
+            # set SW_CMIS_STATE to REMOVED. Additionally, for DEL EVENTS from CONFIG_DB due to DPB,
+            # remove the lport from port_dict.
+            if lport in self.port_dict:
+                self.update_port_transceiver_status_table_sw_cmis_state(lport, CMIS_STATE_REMOVED)
+
+            if port_change_event.db_name == 'CONFIG_DB' and port_change_event.table_name == 'PORT':
+                self.port_dict.pop(lport)
 
     def get_cmis_dp_init_duration_secs(self, api):
         return api.get_datapath_init_duration()/1000
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -2167,20 +2167,13 @@ class DaemonXcvrd(daemon_base.DaemonBase):
         # Load new platform api class
         try:
             import sonic_platform.platform
-            import sonic_platform_base.sonic_sfp.sfputilhelper
             platform_chassis = sonic_platform.platform.Platform().get_chassis()
             self.log_info("chassis loaded {}".format(platform_chassis))
-            # we have to make use of sfputil for some features
-            # even though when new platform api is used for all vendors.
-            # in this sense, we treat it as a part of new platform api.
-            # we have already moved sfputil to sonic_platform_base
-            # which is the root of new platform api.
-            platform_sfputil = sonic_platform_base.sonic_sfp.sfputilhelper.SfpUtilHelper()
         except Exception as e:
             self.log_warning("Failed to load chassis due to {}".format(repr(e)))
 
         # Load platform specific sfputil class
-        if platform_chassis is None or platform_sfputil is None:
+        if platform_chassis is None:
             try:
                 platform_sfputil = self.load_platform_util(PLATFORM_SPECIFIC_MODULE_NAME, PLATFORM_SPECIFIC_CLASS_NAME)
             except Exception as e:
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -604,12 +604,12 @@ class CmisManagerTask(threading.Thread):
         self.exc = None
         self.task_stopping_event = threading.Event()
         self.main_thread_stop_event = main_thread_stop_event
-        self.port_dict = {}
-        self.port_mapping = copy.deepcopy(port_mapping)
+        self.port_dict = {k: {"asic_id": v} for k, v in port_mapping.logical_to_asic.items()}
         self.isPortInitDone = False
         self.isPortConfigDone = False
         self.skip_cmis_mgr = skip_cmis_mgr
         self.namespaces = namespaces
+        self.xcvr_table_helper = XcvrTableHelper(self.namespaces)
 
     def log_debug(self, message):
         helper_logger.log_debug("CMIS: {}".format(message))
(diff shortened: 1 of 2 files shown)


mihirpat1:
@ishidawataru Can you please fix the code coverage check failure?

ishidawataru:
@mihirpat1 I'm still communicating with the original reporter of this issue to check the fix actually works. After that I'll fix the code coverage failure.

ishidawataru:
@mihirpat1 Confirmed that the fix resolved the issue. Also, I fixed the coverage test failure. Please review.

prgeor (changes requested):
@ishidawataru There is not enough information for me to understand the problem. As long as CONFIG_DB port table has the right number of lanes and Speed specified, Xcvrd must be able to select the best matching application from the list of advertised application by the module. Do you see any issue in DPB port deletion path or DPB port creation path?

ishidawataru:
@prgeor This PR fixes https://github.com/sonic-net/sonic-buildimage/issues/18893. It is not about the application selection. The current `CmisManagerTask` implementation doesn't delete the port info from `port_dict` when ports are deleted due to DPB configuration, which causes a crash.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -896,7 +897,13 @@ def on_port_update_event(self, port_change_event):
 
             self.force_cmis_reinit(lport, 0)
         else:
+            # PORT_DEL event for the same lport happens 3 times because
+            # we are subscribing to CONFIG_DB, STATE_DB|TRANSCEIVER_INFO, and STATE_DB|PORT_TABLE.
+            # We only handle the first one and ignore the rest.
```
longhuan-cisco, line 902: Instead of introducing a new attribute `deleted_ports` and its corresponding update logic, you may be able to simply rely on port_change_event's existing `db_name`(/`table_name`), and only proceed for the case of db_name is CONFG_DB 
https://github.com/sonic-net/sonic-platform-daemons/blob/bf865c6b711833347d3c57e9d84cd366bcd1b776/sonic-xcvrd/xcvrd/xcvrd_utilities/port_event_helper.py#L30-L31

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1621,6 +1623,17 @@ def handle_cmis_state_machine(self, lport, info, is_fast_reboot):
...
 
+    def do_task(self, is_fast_reboot):
+        for lport, info in self.port_dict.items():
+            if self.task_stopping_event.is_set():
+                break
+
+            self.handle_cmis_state_machine(lport, info, is_fast_reboot)
+
+        for event in self.deleted_ports.values():
+            self.port_dict.pop(event.port_name)
```
longhuan-cisco, line 1634: Would it make the flow/logic simpler by just directly deleting `self.port_dict[lport]` entry at the PORT_DEL event of CONFIG_DB in `on_port_update_event()`?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1300,354 +1302,360 @@ def wait_for_port_config_done(self, namespace):
...
-                if lport not in self.port_dict:
-                    continue
-
-                state = get_cmis_state_from_state_db(lport, self.xcvr_table_helper.get_status_tbl(self.port_mapping.get_asic_id_for_logical_port(lport)))
-                if state in CMIS_TERMINAL_STATES or state == CMIS_STATE_UNKNOWN:
-                    if state != CMIS_STATE_READY:
-                        self.port_dict[lport]['appl'] = 0
-                        self.port_dict[lport]['host_lanes_mask'] = 0
-                    continue
+    def handle_cmis_state_machine(self, lport, info, is_fast_reboot):
```
longhuan-cisco, line 1305: The change here seems to be refactoring today's cmis_mgr's `task_worker` function by splitting it into sub functions. I'm assuming the flow/logic of cmis_mgr's task_worker doesn't get changed, it's mostly just splitting functions. But this makes the number of LOC (Line of change) very big. I feel It's a little bit difficult for other people to distinguish the actual change of DPB fix from the entire change set of this PR.

Thus, I feel it might be better to put this refactoring change set in a separate PR. It will also be less of a work for release mgr to revert in case of anything broken. Just my 2 cents, not something must-have.
Not sure how @mihirpat1 or Prince feel about this.

mihirpat1: @longhuan-cisco Agreed that splitting the current changeset into 2 separate PRs will be more helpful.

ishidawataru: @longhuan-cisco @mihirpat1 I'm sorry for the late response. Thanks for the suggestion.

I'll move the refactoring commit to a separate PR and only include the DPB fix in this PR.

sonic-xcvrd/xcvrd/xcvrd.py
ishidawataru: https://github.com/sonic-net/sonic-platform-daemons/pull/500/files#diff-fc70c43977a1ce57f3a6e9d7d09cb7ea5204f52b3adf3162936205e5680efb2fR899-R902

@longhuan-cisco  Thank you for the suggestion. I was able to fix the issue without introducing `deleted_ports`.

I initially added `deleted_ports` because the CmisManagerTask handles port events in two stages.

The first stage is done in `on_port_update_event` via   [`port_change_observer.handle_port_update_event()`](https://github.com/sonic-net/sonic-platform-daemons/pull/500/files#diff-fc70c43977a1ce57f3a6e9d7d09cb7ea5204f52b3adf3162936205e5680efb2fR1313) and the second stage is done after returning from `port_change_observer.handle_port_update_event()` by [`iterating over port_dict`](https://github.com/sonic-net/sonic-platform-daemons/pull/500/files#diff-fc70c43977a1ce57f3a6e9d7d09cb7ea5204f52b3adf3162936205e5680efb2fR1315).

If we delete the port info from `port_dict` in the first stage, the second stage cannot handle the port.
By introducing `deleted_ports`, the origin code was deleting the port info from `port_dict` after finishing the second stage.
However, it looks like currently, the second stage is not doing anything for the removed port. so deleting the port info in the first stage is safe.

Though this works, I think the current two-stage design is bug-prone, because we might want to add some handling(e.g. de-initialize) for the removed port in the second stage.
If you agree, I'd like to refactor the current design in a different PR.

sonic-xcvrd/xcvrd/xcvrd.py
ishidawataru: Done.

I explained the background of this change [here](https://github.com/sonic-net/sonic-platform-daemons/pull/500#discussion_r1762172401)

ishidawataru:
@longhuan-cisco @mihirpat1 I updated the PR based on @longhuan-cisco 's suggestion. Please review.

I used force-push and removed the original commits. I backed up the original branch [here](https://github.com/ishidawataru/sonic-platform-daemons/tree/fix-xcvrd-dpb-bak).

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -896,8 +895,13 @@ def on_port_update_event(self, port_change_event):
...
 
             self.force_cmis_reinit(lport, 0)
-        else:
+
+        # PORT_DEL event for the same lport happens 3 times because
+        # we are subscribing to CONFIG_DB, STATE_DB|TRANSCEIVER_INFO, and STATE_DB|PORT_TABLE.
+        # To only handle the first one, check if the lport exists in the port_dict.
+        elif lport in self.port_dict:
             self.update_port_transceiver_status_table_sw_cmis_state(lport, CMIS_STATE_REMOVED)
+            self.port_dict.pop(lport)
```
chiourung, line 904: If a transceiver is removed (not doing port breakout), the port would be removed from port_dict. When the transceiver is reattached to the port, there would be no port-related information in port_dict. I think the port should only be removed from port_dict after receiving a delete CONFIG_DB|PORT event and update cmis state after receiving a delete STATE_DB|TRANSCEIVER_INFO event.

ishidawataru: Thank you for the comment. You're right. I'll update the code as you suggested.

ishidawataru: @chiourung I fixed the issue with this commit. https://github.com/sonic-net/sonic-platform-daemons/pull/500/commits/954e9cb8f39554c2770c030c796c39feb667c194

Could you take a look?

ishidawataru:
@prgeor Hi, just a gentle ping to check if I need to do something to get this PR merged.

ishidawataru:
@prgeor I fixed the conflicts. Please take a look.

prgeor:
@ishidawataru will you be able to test this on a real hardware with breakout enabled and do insertion and removal of the optics?

ishidawataru:
@prgeor Sorry, I don’t have access to real hardware to test this PR. I performed insertion and removal testing using the xcvr emulator in last October. I can redo those tests and provide the logs if needed.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -681,8 +680,45 @@ def on_port_update_event(self, port_change_event):
...
+            # 10. DEL Ethernet0 STATE_DB|TRANSCEIVER_INFO
+            # 11. SET Ethernet0 STATE_DB|TRANSCEIVER_INFO
+            #
+            # To handle both scenarios, if the lport exists in port_dict for any DEL EVENT,
+            # set SW_CMIS_STATE to REMOVED. Additionally, for DEL EVENTS from CONFIG_DB due to DPB,
+            # remove the lport from port_dict.
+            if lport in self.port_dict:
+                self.update_port_transceiver_status_table_sw_cmis_state(lport, CMIS_STATE_REMOVED)
+
+            if port_change_event.db_name == 'CONFIG_DB' and port_change_event.table_name == 'PORT':
```
longhuan-cisco, line 720: would it make the code more robust if we also put line 719-720 protected by line 717 `if lport in self.port_dict:` to avoid `self.port_dict.pop(lport)` throwing KeyError in the case of `lport` not present in port_dict? Though not sure if this rare case can actually happen.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2138,20 +2167,13 @@ def init(self):
...
             import sonic_platform.platform
-            import sonic_platform_base.sonic_sfp.sfputilhelper
             platform_chassis = sonic_platform.platform.Platform().get_chassis()
             self.log_info("chassis loaded {}".format(platform_chassis))
-            # we have to make use of sfputil for some features
-            # even though when new platform api is used for all vendors.
-            # in this sense, we treat it as a part of new platform api.
-            # we have already moved sfputil to sonic_platform_base
-            # which is the root of new platform api.
-            platform_sfputil = sonic_platform_base.sonic_sfp.sfputilhelper.SfpUtilHelper()
```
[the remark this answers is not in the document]
prgeor: @longhuan-cisco 
1. We want to get rid of _wrapper_xxx  as it was applicable when we needed to support platforms using API < v2.0. All platforms by now should be moving to Platform API v2.0.
2. sonic_platform_base.sonic_sfp is already deprecated.

sonic-xcvrd/xcvrd/xcvrd.py
ishidawataru: Sure, I'll fix it as you mentioned.
I just thought if the case is not expected to happen, it's better to treat it as an error instead of silently ignoreing it.
By putting line 720-721 protected by line 717, we won't be able to see if this case is happening or not.

longhuan-cisco: I see your concern, it makes sense to let the exception be thrown to not ignore error. 

The case might not always be an error case. One non-error case I can think of is: If xcvrd receives duplicate DEL events (though it's not very unlikely using today's [PortChangeObserver](https://github.com/sonic-net/sonic-platform-daemons/blob/ba47670352b1778ad3732d27c646e9a1f4b9c49b/sonic-xcvrd/xcvrd/xcvrd_utilities/port_event_helper.py#L155)), protecting it will help xcvrd survive.

It's your choice whether to protect it or leave it as is. I'm fine with either option.

longhuan-cisco:
@prgeor 

I have done OIR tests with the latest diff. 

Tried below sequence on 400G AOC:
1. Move from non-breakout mode (1x400G) to 4x100G breakout mode via DPB CLI
2. Perform OIR
3. Move back to non-breakout mode (1x400G) via DPB CLI
4. Perform OIR

No xcvrd crash. link comes up fine with module ready state. `show interface` CLI shows correct interface entries.

sonic-xcvrd/xcvrd/xcvrd.py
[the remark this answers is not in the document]
longhuan-cisco: got it, thanks for the info.

longhuan-cisco (approved):

prgeor (approved):

PR #501, opened 2024-06-14
[xcvrd] Fix issue: do not print error log when module support non-ethernet application

Junchao-Mellanox:
#### Description

In function `get_interface_speed`, it will log an error if the application name does not contain a valid speed. However, many modules support both ethernet and non-ethernet (e.g. IB) application, non-ethernet application would not contain a valid speed. For such module, the function `get_interface_speed` logs an error to syslog:

```
Jun 12 20:09:57.315556 sonic ERR pmon#xcvrd: No interface speed found for: 'IB HDR (Arch.Spec.Vol.2)'
``` 

Actually, this is not an error because the module will finally find a proper application via the for loop in function `get_cmis_application_desired`

#### Motivation and Context

Remove the unexpected error log since it could be a false alarm. Also, log the module applications if it cannot find a proper application.

Manual test

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -160,6 +160,7 @@ def get_cmis_application_desired(api, host_lane_count, speed):
         get_interface_speed(app_info.get('host_electrical_interface_id')) == speed):
             return (index & 0xf)
 
+    helper_logger.log_error(f'Failed to get desired application from {appl_dict}')
     return None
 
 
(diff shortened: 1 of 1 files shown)


keboliu (approved):

mihirpat1 (approved):

tshalvi (approved):

Junchao-Mellanox:
Backport PR for 202405: https://github.com/sonic-net/sonic-platform-daemons/pull/532 
@bingwang-ms , could you please help merge that?

PR #502, opened 2024-06-17
Avoid loading media_settings.json on mellanox platforms if the host cmis management feature is disabled

tshalvi:
#### Description
I added a check to ensure that the media settings JSON is not loaded when the host CMIS management feature is disabled. This prevents unnecessary port-side configurations and potential errors when the feature is not in use.

#### Motivation and Context
Currently, the xcvrd code always posts the SI parameters from media_settings.json to APP_DB. On Mellanox platforms, this process is required only if the Host CMIS Management feature is enabled. If the feature is disabled when media_settings.json is present on the platform and notify_media_settings() is triggered, it can cause containers to crash or links to flap. This is why I added a check to ensure that on Mellanox platforms, media_settings.json is not loaded if the Host CMIS Management feature is disabled.

Manual tests to ensure that containers do not crash when the feature is disabled and media_settings.json exists on the platform.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
+++ b/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
@@ -6,6 +6,7 @@ import json
 import os
 import ast
 
+from sonic_platform.device_data import DeviceDataManager
 from sonic_py_common import device_info, logger
 from swsscommon import swsscommon
 from xcvrd import xcvrd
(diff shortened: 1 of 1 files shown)


dprital (approved):

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
```diff
@@ -22,6 +22,13 @@
 
 def load_media_settings():
     global g_dict
+
+    version_info = device_info.get_sonic_version_info()
+    if version_info and version_info['asic_type'] == 'mellanox':
```
prgeor, line 27: @tshalvi NO platform specific checks allowed . Xcvrd code should be platform independent.

PR #507, opened 2024-06-26
[xcvrd] Fix CMIS application selection to allow an app whose host lane count is less than the actual value

ishidawataru:
#### Description

`get_cmis_application_desired()` was choosing an app that matches both speed and host_lane_count of the port.
This PR relaxes the constraint so that an app whose required host lane count is less than the actual value can be selected.

#### Motivation and Context

fixes https://github.com/sonic-net/sonic-buildimage/issues/19336

No E2E done yet. UT is added.

#500 needs to be merged before this PR.


prgeor (changes requested):
@ishidawataru I don't understand the constraint when you say Xcvrd is selecting the application based upon number of host lane count and speed. It should be like that. Why you think otherwise?

PR #508, opened 2024-07-03
Fix CMIS and preemphasis lane mapping issues

jemifdo:
- CMIS port db is updated dynamically on DPB
- check_datapath_init_pending is disabled as it need to be done for different lanes sequentially instead of concurrent
- Preemphasis to lane map now works when not all lanes of the port are used
- get_physical_to_logical is now sorted to generated the correct preemphasis values


prgeor (changes requested):
`Preemphasis to lane map now works when not all lanes of the port are used` what do you mean by not all lanes are used?

jemifdo:
@prgeor I mean the case where 1x100G breakout mode is used in an 8 lane port (with speed 25G per lane) and for this the preemphasis dict must have values only the first four lanes. Remaining 4 lanes are unused by the port.

PR #510, opened 2024-07-04
Time set early issue: When time set early, cannot use non-monotonic c…

inspurSDN:
-What I did
    When time set early, cannot use non-monotonic clock. It will make the timer stop. 

-How I did it
    Use monotonic clock to implement thread timer.

-How to verify it
    Set the time earlier and check if it is correct.


prgeor:
@inspurSDN can you explain what was the impact for not having this fix? What was the problem to begin with?

PR #512, opened 2024-07-09
Add speed media key for non CMIS cable

Junchao-Mellanox:
#### Description

There is a function `get_media_settings_key` in `media_settings_parser.py` which is used to extract media setting keys for a given module. The key contains 3 parts: vendor key, media key, speed lane key. 

1. Currently, speed lane key for non CMIS module is always None. For a given module, it could apply different media setting values when it is working on different speed. So, we need to add speed lane key for non CMIS module. 

2. Currently, media key for non CMIS module does not consider the "Extended Specification Compliance". So, the media key could be something like "QSFP+-Extended-255M" or "QSFP+-Unknown-255M", this media key is not good enough. This PR also extends the media key.

#### Motivation and Context

Add media key and speed key for non CMIS cable.

unit test

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
+++ b/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
@@ -58,6 +58,12 @@ def get_lane_speed_key(physical_port, port_speed, lane_count):
             host_electrical_interface_id = appl_adv_dict[app_id].get('host_electrical_interface_id')
             if host_electrical_interface_id:
                 lane_speed_key = LANE_SPEED_KEY_PREFIX + host_electrical_interface_id.split()[0]
+    else:
+        speed_per_lane = int(port_speed) / 1000 / lane_count
+        speed_per_lane_int = int(speed_per_lane)
+        if speed_per_lane - speed_per_lane_int == 0:
+            speed_per_lane = speed_per_lane_int
+        lane_speed_key = LANE_SPEED_KEY_PREFIX + str(speed_per_lane) + 'G:' + str(lane_count)
 
     return lane_speed_key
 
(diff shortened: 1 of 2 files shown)


keboliu (dismissed):

HaiHuang008:
Hi @Junchao-Mellanox,  I'm also keeping an eye on this issue. I have a few minor suggestions regarding this.                               
For the get_lane_speed_key method, the application advertisement in some optical modules does not fully display the supported speeds, so I think it is worth considering having a default lane speed key (via port speed / lane_count ).  Return lane_speed_key as a list allows more flexibility and compatibility. 
You can refer to my patch：[media_settings_parser.patch](https://github.com/user-attachments/files/16173160/media_settings_parser.patch)

Junchao-Mellanox:
Hi @prgeor , @mihirpat1 , could you please review @HaiHuang008 's suggestion? It looks good to me.

liat-grozovik:
> Hi @prgeor , @mihirpat1 , could you please review @HaiHuang008 's suggestion? It looks good to me.

kindly reminder as this is needed also for 202405

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
```diff
@@ -58,6 +58,12 @@ def get_lane_speed_key(physical_port, port_speed, lane_count):
             host_electrical_interface_id = appl_adv_dict[app_id].get('host_electrical_interface_id')
             if host_electrical_interface_id:
                 lane_speed_key = LANE_SPEED_KEY_PREFIX + host_electrical_interface_id.split()[0]
+    else:
+        speed_per_lane = int(port_speed) / 1000 / lane_count
+        speed_per_lane_int = int(speed_per_lane)
+        if speed_per_lane - speed_per_lane_int == 0:
+            speed_per_lane = speed_per_lane_int
+        lane_speed_key = LANE_SPEED_KEY_PREFIX + str(speed_per_lane) + 'G:' + str(lane_count)
```
prgeor, line 66: @Junchao-Mellanox why we need lookup for number of lanes? do you have sample media_setting.json with this for non-CMIS cable? Is this key with no of lanes captured in any HLD?

prgeor:
> Hi @Junchao-Mellanox, I'm also keeping an eye on this issue. I have a few minor suggestions regarding this. For the get_lane_speed_key method, the application advertisement in some optical modules does not fully display the supported speeds, so I think it is worth considering having a default lane speed key (via port speed / lane_count ). Return lane_speed_key as a list allows more flexibility and compatibility. You can refer to my patch：[media_settings_parser.patch](https://github.com/user-attachments/files/16173160/media_settings_parser.patch)

@HaiHuang008 You mean some CMIS cable don't advertise application speed?

prgeor (commented):
@Junchao-Mellanox Doesn't this PR also achieve the same? https://github.com/sonic-net/sonic-platform-daemons/pull/533

prgeor:
@Junchao-Mellanox doesn't this PR address this issue? https://github.com/sonic-net/sonic-platform-daemons/pull/533

HaiHuang008:
@prgeor Yes, some cables do not advertise application speed.

prgeor:
> @prgeor Yes, some cables do not advertise application speed.

@HaiHuang008 How does the host then know which application to load?

Junchao-Mellanox:
close this one, since there is a similar PR https://github.com/sonic-net/sonic-platform-daemons/pull/533

PR #529, opened 2024-08-02
[PMON][psud] Fix the repeated NOTICE log message on Chassis platform

mlok-nokia:
#### Description
In psud script, first_run is always true in PsuChassisInfo class. This causes the method update_master_status() always set the psu status_master_led and log the status change messages although the status is not changed. The same NOTICE message is repeated and never stop.  Address this issue, added code to set the first_run to False after the first run.  This initializes set_led to False before the calculation.  So that it won't be True to trigger the set_status_maste_led() and log the same message while the status has NO change.

#### Motivation and Context
This PR fixes https://github.com/sonic-net/sonic-buildimage/issues/19780 and not set the LED to the same state or log the same message when there is NO change. 

Checking the syslog on the SUP.  The following message will not be repeated if the status is not changed
```
pmon#psud: PSU supplied power warning cleared: supplied power is back to norma
```

### DIFF
--- a/sonic-psud/scripts/psud
+++ b/sonic-psud/scripts/psud
@@ -268,6 +268,8 @@ class PsuChassisInfo(logger.Logger):
                                   'PSU supplied power warning: {}W supplied-power less than {}W consumed-power'.format(
                                       self.total_supplied_power, self.total_consumed_power)
                                   )
+        if self.first_run:
+            self.first_run = False
 
         return set_led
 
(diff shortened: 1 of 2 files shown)


assrinivasan (approved):
Change LGTM. Could you paste a syslog comparison before and after your change?

mlok-nokia:
> Change LGTM. Could you paste a syslog comparison before and after your change?

The context of log message is not changed.  This PR is to stop logging the same message periodically when the Pusd Master Status is not changed.

prgeor (approved):

mlok-nokia:
@prgeor @judyjoseph  Please cherry-pick this PR to 202405 branch

PR #541, opened 2024-09-26
Xcvrd crash and restart should not cause link flap on platforms needing custom NPU SI settings

mihirpat1:
#### Description

Ensure Xcvrd crash and restart should be handled gracefully on platforms that need custom SI settings.
The goal is to avoid link flap on platforms that need custom SI settings during xcvrd restart.

#### Motivation and Context

**Changeset 1**
The changeset is partial implementation of the changes required for https://github.com/sonic-net/SONiC/pull/1432/.
Specifically, the current change adds `NPU_SI_SETTINGS_SYNC_STATUS` key to the PORT_TABLE of STATE_DB. Following is the usage of `NPU_SI_SETTINGS_SYNC_STATUS` key

Value | Modifier thread and event | Consumer thread and purpose
-- | -- | --
NPU_SI_SETTINGS_DEFAULT | 1.   XCVRD main thread during cold start of XCVRD   2. SfpStateUpdateTask during transceiver removal | XCVRD   main thread during boot-up for deciding to notify NPU SI settings
NPU_SI_SETTINGS_NOTIFIED | 1. SfpStateUpdateTask while updating and notifying the NPU SI settings | Not being   used currently

Please note that the above table needs to be further modified while implementing complete OA crash HLD. The table assumes NPU SI settings are applied from SfpStateUpdateTask thread always (unlike the HLD wherein CMIS transceivers have the NPU SI settings being sent from CmisManagerTask).

**Changeset 2**
In addition to the above change, `TRANSCEIVER_INFO` table will not be deleted as part of xcvrd deinit.
On some platforms, `TRANSCEIVER_INFO` table is used to detect transceiver presence to handle `host_tx_ready` behavior and hence, deleting `TRANSCEIVR_INFO` table will cause `host_tx_ready` to change to `false` during xcvrd shutdown.

Impact due to skipping TRANSCEIVER_INFO table deletion
```
Scenario
In case if xcvrd crashes permanently after initializing 6 out of 32 ports (i.e. xcvrd never respawns), then "show int status" CLI will show 6 transceiver present and other ports would not show xcvr present unless we use sfputil. 
```

**Changeset 3**
`PHYSICAL_PORT_NOT_EXIST` has been defined in [media_settings_parser.py](https://github.com/sonic-net/sonic-platform-daemons/pull/541/files#diff-203c0123b1323e080246e562f662ac7c2a805bbae951ebd27f365758c6469f86) to prevent crash due to undefined value.
When the media_settings_parser.py file was created, the below line was moved from xcvrd.py. However, the definition of
`PHYSICAL_PORT_NOT_EXIST` was never ported. Hence, adding the definition now.
https://github.com/sonic-net/sonic-platform-daemons/blob/8c89f6ba2975a7699861f2c6d77083cebb62e97c/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py#L280

On platforms which do not restart xcvrd upon swss/syncd crash/restart, following behavior is observed
The first xcvrd crash after swss/syncd restarts will cause a link flap for ports which require NPU SI settings.
This is expected since as part of swss crash handling, the entire PORT_TABLE of STATE_DB is deleted. 
Since xcvrd is not restarted as part of swss/syncd crash handling, the NPU_SI_SETTINGS_SYNC_STATUS field is never populated. Hence, when xcvrd crashes/restarts first time after swss/syncd crash is triggered, xcvrd will update the NPU SI settings in the APPL_DB which in turn forces the port to go through a link flap while OA configures the NPU SI settings

Test summary
1. Platforms with xcvrd not restarting during swss/syncd crash

Event | Xcvrd restarted | NPU SI settings renotify | NPU_SI_SETTINGS_SYNC_STATUS   value after action   (for optics which require NPU SI settings) | Link flap
-- | -- | -- | -- | --
Config shut | N/A | No | NPU_SI_SETTINGS_NOTIFIED | N/A
Config no shut | N/A | No | NPU_SI_SETTINGS_NOTIFIED | N/A
Xcvrd restart | Yes | No | NPU_SI_SETTINGS_NOTIFIED | No
Xcvrd kill | Yes | No | NPU_SI_SETTINGS_NOTIFIED | No
Crash xcvrd just before updating   NPU_SI_SETTINGS_SYNC_STATUS | Yes | Yes | NPU_SI_SETTINGS_NOTIFIED | N/A
Pmon restart | Yes | No | NPU_SI_SETTINGS_NOTIFIED | No
Swss   restart | No | No | Key is   deleted | Yes
Syncd   restart | No | No | Key is   deleted | Yes
config reload | Yes | Yes | NPU_SI_SETTINGS_NOTIFIED | Yes
Cold reboot | N/A | No | Inline with expected value | N/A

2. Platforms with xcvrd restarting during swss/syncd crash

Event | Xcvrd restarted | NPU SI settings renotify | NPU_SI_SETTINGS_SYNC_STATUS   value after action   (for optics   which require NPU SI settings) | Link flap
-- | -- | -- | -- | --
Config shut | N/A | No | NPU_SI_SETTINGS_NOTIFIED | N/A
Config no shut | N/A | No | NPU_SI_SETTINGS_NOTIFIED | N/A
Xcvrd restart | Yes | No | NPU_SI_SETTINGS_NOTIFIED | No
Xcvrd kill | Yes | No | NPU_SI_SETTINGS_NOTIFIED | No
Crash xcvrd just before updating   NPU_SI_SETTINGS_SYNC_STATUS | Yes | Yes | NPU_SI_SETTINGS_NOTIFIED | N/A
Pmon restart | Yes | No | NPU_SI_SETTINGS_NOTIFIED | No
Swss restart | Yes | Yes | NPU_SI_SETTINGS_NOTIFIED | Yes
Syncd  restart | Yes | Yes | NPU_SI_SETTINGS_NOTIFIED | Yes
config reload | Yes | Yes | NPU_SI_SETTINGS_NOTIFIED | Yes
Cold reboot | N/A | No | NPU_SI_SETTINGS_NOTIFIED | N/A
Warm   reboot | N/A | No | NPU_SI_SETTINGS_NOTIFIED | No

3. Verified transceiver OIR
MSFT ADO - 29278409

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -28,7 +28,7 @@ try:
 
     from .xcvrd_utilities import sfp_status_helper
     from .sff_mgr import SffManagerTask
-    from .xcvrd_utilities.xcvr_table_helper import XcvrTableHelper
+    from .xcvrd_utilities.xcvr_table_helper import *
     from .xcvrd_utilities import port_event_helper
     from .xcvrd_utilities.port_event_helper import PortChangeObserver
     from .xcvrd_utilities import media_settings_parser
@@ -1934,7 +1934,7 @@ class SfpStateUpdateTask(threading.Thread):
 
                 # Do not notify media settings during warm reboot to avoid dataplane traffic impact
                 if is_warm_start == False:
-                    media_settings_parser.notify_media_setting(logical_port_name, transceiver_dict, xcvr_table_helper.get_app_port_tbl(asic_index), xcvr_table_helper.get_cfg_port_tbl(asic_index), port_mapping)
+                    media_settings_parser.notify_media_setting(logical_port_name, transceiver_dict, xcvr_table_helper, port_mapping)
                     transceiver_dict.clear()
             else:
                 retry_eeprom_set.add(logical_port_name)
@@ -2156,10 +2156,12 @@ class SfpStateUpdateTask(threading.Thread):
 
                                 if rc != SFP_EEPROM_NOT_READY:
                                     post_port_dom_threshold_info_to_db(logical_port, self.port_mapping, self.xcvr_table_helper.get_dom_threshold_tbl(asic_index))
-                                    media_settings_parser.notify_media_setting(logical_port, transceiver_dict, self.xcvr_table_helper.get_app_port_tbl(asic_index), self.xcvr_table_helper.get_cfg_port_tbl(asic_index), self.port_mapping)
+                                    media_settings_parser.notify_media_setting(logical_port, transceiver_dict, self.xcvr_table_helper, self.port_mapping)
                                     transceiver_dict.clear()
                             elif value == sfp_status_helper.SFP_STATUS_REMOVED:
                                 helper_logger.log_notice("{}: Got SFP removed event".format(logical_port))
+                                state_port_table = self.xcvr_table_helper.get_state_port_tbl(asic_index)
+                                state_port_table.set(logical_port, [(NPU_SI_SETTINGS_SYNC_STATUS_KEY, NPU_SI_SETTINGS_DEFAULT_VALUE)])
                                 update_port_transceiver_status_table_sw(
                                     logical_port, self.xcvr_table_helper.get_status_tbl(asic_index), sfp_status_helper.SFP_STATUS_REMOVED)
                                 helper_logger.log_notice("{}: received plug out and update port sfp status table.".format(logical_port))
@@ -2354,7 +2356,7 @@ class SfpStateUpdateTask(threading.Thread):
                 self.retry_eeprom_set.add(port_change_event.port_name)
             else:
                 post_port_dom_threshold_info_to_db(port_change_event.port_name, self.port_mapping, dom_threshold_tbl)
-                media_settings_parser.notify_media_setting(port_change_event.port_name, transceiver_dict, self.xcvr_table_helper.get_app_port_tbl(port_change_event.asic_id), self.xcvr_table_helper.get_cfg_port_tbl(port_change_event.asic_id), self.port_mapping)
+                media_settings_parser.notify_media_setting(port_change_event.port_name, transceiver_dict, self.xcvr_table_helper, self.port_mapping)
         else:
             status = sfp_status_helper.SFP_STATUS_REMOVED if not status else status
         update_port_transceiver_status_table_sw(port_change_event.port_name, status_tbl, status, error_description)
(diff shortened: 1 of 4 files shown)


sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2437,6 +2439,30 @@ def wait_for_port_config_done(self, namespace):
             if key in ["PortConfigDone", "PortInitDone"]:
                 break
 
+    """
+    Initialize NPU_SI_SETTINGS_SYNC_STATUS_KEY field in STATE_DB PORT_TABLE|<lport>
```
prgeor, line 2443: @mihirpat1 where do we delete the key when optics is removed?

mihirpat1: @prgeor We don't delete the key when optics is removed. We instead mark it as `NPU_SI_SETTINGS_DEFAULT`.
This change is at line 2163-2164 in xcvrd.py.

prgeor (approved):

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2156,10 +2156,12 @@ def task_worker(self, stopping_event, sfp_error_event):
 
                                 if rc != SFP_EEPROM_NOT_READY:
                                     post_port_dom_threshold_info_to_db(logical_port, self.port_mapping, self.xcvr_table_helper.get_dom_threshold_tbl(asic_index))
-                                    media_settings_parser.notify_media_setting(logical_port, transceiver_dict, self.xcvr_table_helper.get_app_port_tbl(asic_index), self.xcvr_table_helper.get_cfg_port_tbl(asic_index), self.port_mapping)
+                                    media_settings_parser.notify_media_setting(logical_port, transceiver_dict, self.xcvr_table_helper, self.port_mapping)
                                     transceiver_dict.clear()
                             elif value == sfp_status_helper.SFP_STATUS_REMOVED:
                                 helper_logger.log_notice("{}: Got SFP removed event".format(logical_port))
+                                state_port_table = self.xcvr_table_helper.get_state_port_tbl(asic_index)
```
prgeor, line 2163: @mihirpat1 is there any specific reason to not delete this key on optics removal?

mihirpat1: @prgeor I am not deleting the key on optics removal since as part of xcvrd boot-up, we are setting the key to `NPU_SI_SETTINGS_DEFAULT` for all ports (irrespective of transceiver presence). Hence, in order to keep the state consistent across device boot-up and optics OIR, I am setting the value as opposed to deleting the key.
Also, differentiating if a port requires NPU SI settings as part of device boot-up will put overhead on xcvrd main thread (via `initialize_port_init_control_fields_in_port_table`) to read the EEPROM and conclude if the optics requires NPU SI settings and set the value accordingly. Hence, I am setting the value for this key to `NPU_SI_SETTINGS_DEFAULT` for all ports and have SfpStateUpdateTask (and CmisManagerTask in future) to conclude if the optics requires NPU SI settings.

Junchao-Mellanox (approved):

mihirpat1:
@bingwang-ms Can you please help to cherry-pick this PR to 202405 branch?
MSFT ADO - 29278409

PR #554, opened 2024-10-25
[SmartSwitch] Add implementation for the DPU chassis daemon.

oleksandrivantsiv:
#### Description

Implement the Chassis daemon that runs on the Smart Switch DPU is responsible for maintaining the DPU state in the `CHASSIS_STATE_DB`. DB schema is defined in [Smart Switch PMON HLD](doc/smart-switch/pmon/smartswitch-pmon.md).

#### Motivation and Context

Provide up-to-date DPU data and control plane states in the CHASSIS_STATE_DB.

The new functionality is fully covered with the UT provided in this PR.

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -15,6 +15,7 @@ try:
     import sys
     import threading
     import time
+    from datetime import datetime
 
     from sonic_py_common import daemon_base, logger, device_info
     from sonic_py_common.task_base import ProcessTaskBase
@@ -81,8 +82,6 @@ CHASSIS_DB_CLEANUP_MODULE_DOWN_PERIOD = 30 # Minutes
 CHASSIS_LOAD_ERROR = 1
 CHASSIS_NOT_SUPPORTED = 2
 
-platform_chassis = None
-
 SELECT_TIMEOUT = 1000
 
 NOT_AVAILABLE = 'N/A'
@@ -122,6 +121,14 @@ def try_get(callback, *args, **kwargs):
         ret = default
 
     return ret
+
+def get_chassis():
+    try:
+        import sonic_platform.platform
+        return sonic_platform.platform.Platform().get_chassis()
+    except Exception as e:
+        self.log_error("Failed to load chassis due to {}".format(repr(e)))
+        sys.exit(CHASSIS_LOAD_ERROR)
 
 #
 # Module Config Updater ========================================================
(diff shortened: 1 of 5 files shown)


vivekrnv (approved):

prgeor (approved):

prgeor:
@rameshraghupathy  can you review

sonic-chassisd/scripts/chassisd
```diff
@@ -605,58 +612,153 @@ class ConfigManagerTask(ProcessTaskBase):
...
             self.config_updater.module_config_update(key, admin_state)
 
+
+#
+# State Manager task ========================================================
+#
+
+class DpuStateUpdater(logger.Logger):
+
+    DP_STATE = 'dpu_data_plane_state'
```
rameshraghupathy, line 622: How about state_reason?  Any indication of why the CP/DP is down etc?

rameshraghupathy (commented):
The limitation of pull model is that it is limited by the polling interval. Increasing the polling frequency may cause additional overhead. Potential chances of missing pulse events less than the polling interval.

sonic-chassisd/scripts/chassisd
oleksandrivantsiv: This information is available in the `show system-health detail` command output on the DPU. I see no reason to duplicate it.

oleksandrivantsiv:
> The limitation of pull model is that it is limited by the polling interval. Increasing the polling frequency may cause additional overhead. Potential chances of missing pulse events less than the polling interval.

I'll provide a subscription-based implementation for the platforms that don't require special handling of the data plane and control plane statuses in a separate PR.

This PR is to unblock the integration and allow the DPU to post its state to the CHASSIS_STATE_DB.

sonic-chassisd/scripts/chassisd
rameshraghupathy: Having a summary of any container/service/hw gradually degrading on the DPU is good to have on the NPU. Even when the DPU is not reachable having some data like this will be very helpful for debug.

oleksandrivantsiv: I don't agree. Based on my experience if the data plane, control plane, or both are in a permanent down state, the summary won’t help in understanding the root cause of the problem. The user will need to connect to the DPU, read logs, check dumps, and debug the issue.

If the state change was temporary, this also won’t help in understanding the problem because the reason will be overwritten when the state changes.

rameshraghupathy: No worries.  You can commit this.

rameshraghupathy: The PMON HLD was primarily based on https://github.com/sonic-net/SONiC/blob/6edaef09c23cdc37c15983ac4711a9e5ed304d6e/doc/pmon/pmon-chassis-design.md.  Please refer to 3.2.4

rameshraghupathy (approved):

PR #562, opened 2024-11-19
[xcvrd] Fix swsscommon set port table handle in media_settings parser

vdahiya12:
instead of ProducerStateTable for settings kvp, the xcvrd daemon should be just using swsscommon.Table which is what this PR fixes.


prgeor:
@vdahiya12 please update
```
Motivation and Context
How Has This Been Tested?
```

prgeor:
@vdahiya12 please reopen if required.

PR #565, opened 2024-11-21
Take non-CMIS xcvrs out of lpmode in SFF Manager

peterbailey-arista:
#### Description
Fix non-CMIS transceivers in down state by bringing them out of low power mode in the SFF Manager Task.
This is intended to work together with the change in https://github.com/sonic-net/sonic-buildimage/pull/20886.

#### Motivation and Context
Non-CMIS transceivers were not functioning correctly when put into Low Power mode. So XCVRD now brings them out of lpmode.

Loaded an image containing this change alongside the change from https://github.com/sonic-net/sonic-buildimage/pull/20886 on an Arista chassis containing a Clearwater2 linecard.
Verified that without this image some interfaces were in a down state but with the image all interfaces came up as expected.

### DIFF
--- a/sonic-xcvrd/xcvrd/sff_mgr.py
+++ b/sonic-xcvrd/xcvrd/sff_mgr.py
@@ -14,6 +14,7 @@ try:
 
     from .xcvrd_utilities.port_event_helper import PortChangeObserver
     from .xcvrd_utilities.xcvr_table_helper import XcvrTableHelper
+    from sonic_platform_base.sonic_xcvr.api.public.sff8472 import Sff8472Api
 except ImportError as e:
     raise ImportError(str(e) + " - required module not found")
 
@@ -434,6 +435,18 @@ class SffManagerTask(threading.Thread):
                 except (AttributeError, NotImplementedError):
                     # Skip if these essential routines are not available
                     continue
+                
+                if xcvr_inserted:
+                    set_lp_success = (
+                        sfp.set_lpmode(False) 
+                        if isinstance(api, Sff8472Api) 
+                        else api.set_lpmode(False)
+                    )
+                    if not set_lp_success:
+                        self.log_error(
+                            "{}: Failed to take module out of low power mode.".format(
+                                lport)
+                        )
 
                 if active_lanes is None:
                     active_lanes = self.get_active_lanes_for_lport(lport, subport_idx,


kenneth-arista:
@arlakshm @wenyiz2021 for awareness

wenyiz2021 (approved):
LGTM

wenyiz2021:
do we only need this PR for the LP mode causing links down issue?

kenneth-arista:
Yes, along with enabling the SFF manager in xcvrd per platform config (
https://github.com/sonic-net/sonic-buildimage/pull/20886).

prgeor:
@mihirpat1 can you please review

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -435,6 +435,8 @@ def task_worker(self):
                     # Skip if these essential routines are not available
                     continue
 
+                sfp.set_lpmode(False)
```
prgeor, line 438: @peterbailey-arista I would suggest to use sfp.set_lpmode() only for SFPs that follow SFF8472. All other transceivers like QSFP+, QSFP28 can support lpmode via EEPROM write. The above code expects each platform to implement `set_lpmode()` even thought that is NOT required for QSFP based modules.
```
if (SFP type module) {
    sfp.set_lpmode(False)
} else {
    api = set_lpmode(False)
}
```

peterbailey-arista: After discussing with @byu343 I ended up wrapping api.set_lpmode in a try except instead. Please let me know if this update works for you. Thanks!

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -435,6 +435,12 @@ def task_worker(self):
                     # Skip if these essential routines are not available
                     continue
 
+                try:
+                    api = sfp.get_xcvr_api()
+                    api.set_lpmode(False)
```
mihirpat1, line 440: @peterbailey-arista Per the below implementation, an exception will not be returned for SFF-8472 modules. Can you please handle this accordingly.
https://github.com/sonic-net/sonic-platform-common/blob/0f2e22faccd093a1e5d18235fe119a860be7855e/sonic_platform_base/sonic_xcvr/api/public/sff8472.py#L308

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -435,6 +435,12 @@ def task_worker(self):
                     # Skip if these essential routines are not available
                     continue
 
+                try:
+                    api = sfp.get_xcvr_api()
```
longhuan-cisco, line 439: `api` has been already obtained at the code before: 
https://github.com/sonic-net/sonic-platform-daemons/blob/78d427e2659089fb80bc60e1b2db63fe917489fc/sonic-xcvrd/xcvrd/sff_mgr.py#L410

sonic-xcvrd/xcvrd/sff_mgr.py
peterbailey-arista: I've now updated it to use sfp.set_lpmode() for only SFPs implementing SFF8472 as was originally suggested

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -435,6 +436,11 @@ def task_worker(self):
                     # Skip if these essential routines are not available
                     continue
 
+                if isinstance(api, Sff8472Api):
+                    sfp.set_lpmode(False)
```
mihirpat1, line 440: @peterbailey-arista Can we check for the return value in both the cases and log error if it returns False?
Also, can you please help in fixing the built failure?

kenneth-arista:
Above build failure is unrelated to the change in this PR. The failure is caused by:
```
dpkg: error: cannot access archive 'libnl-3-200_*.deb': No such file or directory
```

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -435,6 +436,17 @@ def task_worker(self):
                     # Skip if these essential routines are not available
                     continue
 
+                set_lp_success = (
```
longhuan-cisco, line 439: I assume lpmode handling is only needed at module insertion event, right?

Just some minor suggestion:
Maybe adding below condition check can avoid unnecessary lpmode handling in other cases (e.g.   the case of admin_status/host_tx_ready getting changed by `config interface shutdown/startup`)
```
if xcvr_inserted:
    <lpmode logic>
```

peterbailey-arista: The described scenario unfortunately still requires lpmode set False. If you shutdown then startup the interface without bringing it out of lpmode the interface remains down even if it was up before shutdown

peterbailey-arista: My mistake, I believe you are correct. I'll add that change, thanks

arlakshm:
@longhuan-cisco @mihirpat1, Can you please approve this change if all the comments are addressed.

mihirpat1 (commented):
@peterbailey-arista The changes look good to me. Can you please help in resolving the build failure?

longhuan-cisco (approved):

arlakshm (approved):

PR #573, opened 2024-12-12
[chassisd] Address the chassisd crash issue and add UT for it

mlok-nokia:
#### Description
On Nokia platform, slot name of Supervisor is string "A" instead of a number. Using "int" to convert it could cause issue backtrace.   We should use slot value to any checking without any conversion. This will fixes  https://github.com/sonic-net/sonic-buildimage/issues/21131
 

#### Motivation and Context
Modify the _get_module_info not to convert "slot" to a string value.  And also modify the code not to convert slot value to an to do any checking.  Just directly use the returned value of get_slot().  Also add UT  test_moduleupdater_check_slot_string() to valid it.

Tested on 202405 branch

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202405

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -283,7 +283,7 @@ class ModuleUpdater(logger.Logger):
 
         for module_index in range(0, self.num_modules):
             module_info_dict = self._get_module_info(module_index)
-            if self.my_slot == int(module_info_dict['slot']):
+            if self.my_slot == module_info_dict['slot']:
                 my_index = module_index
 
             if module_info_dict is not None:
@@ -300,7 +300,7 @@ class ModuleUpdater(logger.Logger):
 
                 fvs = swsscommon.FieldValuePairs([(CHASSIS_MODULE_INFO_DESC_FIELD, module_info_dict[CHASSIS_MODULE_INFO_DESC_FIELD]),
                                                   (CHASSIS_MODULE_INFO_SLOT_FIELD,
-                                                   module_info_dict[CHASSIS_MODULE_INFO_SLOT_FIELD]),
+                                                   str(module_info_dict[CHASSIS_MODULE_INFO_SLOT_FIELD])),
                                                   (CHASSIS_MODULE_INFO_OPERSTATUS_FIELD, module_info_dict[CHASSIS_MODULE_INFO_OPERSTATUS_FIELD]),
                                                   (CHASSIS_MODULE_INFO_NUM_ASICS_FIELD, str(len(module_info_dict[CHASSIS_MODULE_INFO_ASICS]))),
                                                   (CHASSIS_MODULE_INFO_SERIAL_FIELD, module_info_dict[CHASSIS_MODULE_INFO_SERIAL_FIELD])])
@@ -357,8 +357,8 @@ class ModuleUpdater(logger.Logger):
 
         # In line card push the hostname of the module and num_asics to the chassis state db.
         # The hostname is used as key to access chassis app db entries 
-        module_info_dict = self._get_module_info(my_index)
         if not self._is_supervisor():
+           module_info_dict = self._get_module_info(my_index)
            hostname_key = "{}{}".format(ModuleBase.MODULE_TYPE_LINE, int(self.my_slot) - 1)
            hostname = try_get(device_info.get_hostname, default="None")
            hostname_fvs = swsscommon.FieldValuePairs([(CHASSIS_MODULE_INFO_SLOT_FIELD, str(self.my_slot)), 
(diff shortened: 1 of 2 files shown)


mlok-nokia:
@judyjoseph and @arlakshm This PR address the chassisd crash issue.   I also add UT check module_db_update support slot number or slot string. Please review

judyjoseph (commented):
LGTM. 

@mlok-nokia  can you also confirm it will not conflict with these two platform API's module.get_slot(), chassis.get_my_slot() which as per commens will return int.

judyjoseph:
@kenneth-arista @anamehra could you check in your platforms too if this change is ok ?

arlakshm:
@kenneth-arista, please review this change

sonic-chassisd/scripts/chassisd
```diff
@@ -357,8 +357,8 @@ class ModuleUpdater(logger.Logger):
 
         # In line card push the hostname of the module and num_asics to the chassis state db.
         # The hostname is used as key to access chassis app db entries 
-        module_info_dict = self._get_module_info(my_index)
         if not self._is_supervisor():
+           module_info_dict = self._get_module_info(my_index)
            hostname_key = "{}{}".format(ModuleBase.MODULE_TYPE_LINE, int(self.my_slot) - 1)
```
spilkey-cisco, line 362: What is the expectation here? This makes an assumption that line card slots must be numbers, but should that be a requirement? Should this be using `my_index` instead? Or should there be some logic to handle if the LC slot was non-numeric?

sonic-chassisd/scripts/chassisd
```diff
@@ -393,7 +393,7 @@ class ModuleUpdater(logger.Logger):
 
         module_info_dict[CHASSIS_MODULE_INFO_NAME_FIELD] = name
         module_info_dict[CHASSIS_MODULE_INFO_DESC_FIELD] = str(desc)
-        module_info_dict[CHASSIS_MODULE_INFO_SLOT_FIELD] = str(slot)
+        module_info_dict[CHASSIS_MODULE_INFO_SLOT_FIELD] = slot
```
spilkey-cisco, line 396: This may be an unexpected behavior change. `slot` can still be an integer, so `module_info_dict[CHASSIS_MODULE_INFO_SLOT_FIELD]` can now have varying types. Previously when `slot was always `int`, all logic was expecting this field to be `str` since it was converted here (though I'm not sure any logic relies on that type). It may be best to revert this change to maintain a consistent (and existing) type requirement.

I do see there is a conversion for one of the usages of this dict entry to `str` to account for this, but it may just be best to maintain the expectation that this dict entry is always `str` from this point, as it was previously.

sonic-chassisd/scripts/chassisd
mlok-nokia: Move the self._get_module_info() call inside if condition since my_slot is only use inside.  Not need to call self._get_module_info(0 if it not required.

sonic-chassisd/scripts/chassisd
mlok-nokia: The fundamental issue is a slot for chassis my not be number,  it could be a string -- such as Nokia IXRE7250 
It is not necessary to convert to string in the buffer here.  We can directly use it as what function get_slot() returned.  The only place is needed to be converted to a string when it is stored into a STATE_DB.  This change simply and no need to  convert back and forth (StringToNumber or NumberToString) when use the slot info in module_info_dict[].

sonic-chassisd/scripts/chassisd
spilkey-cisco: But this is still requiring that line cards must be named as numbers, only supervisors are allowed to be alphanumeric. Is that expected and required? Why can't alphanumeric line card names also be supported since this was already found to be a limitation of supervisor naming?

sonic-chassisd/scripts/chassisd
spilkey-cisco: Sounds good. My concern was mostly that it may be unclear to developers that the type stored in this dict field can vary. Functionally it seems fine. The documentation of APIs such as `get_slot` and `get_my_slot` should be updated as @judyjoseph previously mentioned; they not need not return only an int, they may also return str.

sonic-chassisd/scripts/chassisd
mlok-nokia: Yes. Linecard slot must be number string or number.  It is required.  Otherwise, we will not able construct the linecard name.  BTW, this change just moved the _get_module_info() call into "if" section since we only use it inside the "if".

sonic-chassisd/scripts/chassisd
mlok-nokia: > Sounds good. My concern was mostly that it may be unclear to developers that the type stored in this dict field can vary. Functionally it seems fine. The documentation of APIs such as `get_slot` and `get_my_slot` should be updated as @judyjoseph previously mentioned; they not need not return only an int, they may also return str.

Agree.  For the long run, this needs to be addressed.

arlakshm:
@judyjoseph, @spilkey-cisco if the comments are resolved please help approve this PR. @kenneth-arista please signoff if this change works on Arista platform.

sonic-chassisd/scripts/chassisd
mlok-nokia: Agree. For the long term solution, we should introduce label attribute to support alphanumeric slot.

judyjoseph (approved):

kenneth-arista (approved):
I tested this quickly and the change works for me.

arlakshm (approved):

PR #574, opened 2024-12-13
Add high power class support for non-CMIS modules in xcvrd

longhuan-cisco:
HLD: https://github.com/sonic-net/SONiC/pull/1867

#### Description

- Add xcvrd/sff_mgr support for enabling high power class if module's power class is greater or equal to 5 (Refer to SFF-8636 spec 6.2.6 Control Functions (Page 00h, Bytes 86-99) for byte 93 definition)

#### Motivation and Context

Without enabling high power class for those modules, the module won't operate properly with power output and link won't come up.

Verified the module operate with proper power output.

### DIFF
--- a/sonic-xcvrd/xcvrd/sff_mgr.py
+++ b/sonic-xcvrd/xcvrd/sff_mgr.py
@@ -441,8 +441,8 @@ class SffManagerTask(threading.Thread):
 
                 if xcvr_inserted:
                     try:
-                        if api.handle_high_power_class():
-                            self.log_notice("{}: done handling high power class".format(lport))
+                        if api.set_high_power_class(True):
+                            self.log_notice("{}: done enabling high power class".format(lport))
                         else:
                             self.log_error("{}: failed to enable high power class".format(lport))
                     except (AttributeError, NotImplementedError):
--- a/sonic-xcvrd/xcvrd/sff_mgr.py
+++ b/sonic-xcvrd/xcvrd/sff_mgr.py
@@ -299,8 +299,15 @@ class SffManagerTask(threading.Thread):
                 self.log_error("{}: failed to get power class".format(lport))
                 return
 
+            # According to SFF-8636, section 6.2.6:
+            # In order to protect legacy host systems designed to support only
+            # the original 4 power classes, the High Power Class Enable control
+            # was defined at byte 93, bit 2. Modules in power classes 5, 6, 7 or
+            # 8 are required to limit power consumption to no more than a power
+            # class 4 module if the High Power Class Enable, byte 93 bit 2
+            # control is not set. For power class < 5, there's no such power
+            # consumption limiting, so no need to enable high power class.
             if power_class < 5:
-                # No action needed for power class < 5
                 return
 
             if xcvr_api.set_high_power_class(power_class, True):
--- a/sonic-xcvrd/xcvrd/sff_mgr.py
+++ b/sonic-xcvrd/xcvrd/sff_mgr.py
@@ -79,10 +79,9 @@ class SffManagerTask(threading.Thread):
 
     SFF_LOGGER_PREFIX = "SFF-MAIN: "
 
-    def __init__(self, enable_sff_mgr_controlled_tx, namespaces, main_thread_stop_event, platform_chassis, helper_logger):
+    def __init__(self, namespaces, main_thread_stop_event, platform_chassis, helper_logger):
         threading.Thread.__init__(self)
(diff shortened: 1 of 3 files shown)


sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -284,6 +288,32 @@ def convert_bool_array_to_bit_mask(self, bool_array):
...
+        try:
+            power_class = xcvr_api.get_power_class()
+            if power_class is None:
+                self.log_error("{}: failed to get power class".format(lport))
+                return
+
+            if power_class < 5:
+                # No action needed for power class < 5
+                return
+
```
longhuan-cisco, line 308: @prgeor 
Continue the discussion of https://github.com/sonic-net/sonic-platform-common/pull/521#discussion_r1986444901 here at sff_mgr PR, regarding whether setting power_override bit should be included as part of `handle_high_power_class`:

longhuan-cisco: high power class enabling (added by this PR) and [lpmode turning-off](https://github.com/sonic-net/sonic-platform-daemons/blob/ee4fb9e434e4c735488c6d4f93eca4277bd0facf/sonic-xcvrd/xcvrd/sff_mgr.py#L439-L449) are two different independent steps. Not all modules need high power class enabling(only for power_class>=5), while most of the optical modules need lpmode turning-off.

Today,  power_override bit setting has already been done by sff_mgr's [lpmode turning-off](https://github.com/sonic-net/sonic-platform-daemons/blob/ee4fb9e434e4c735488c6d4f93eca4277bd0facf/sonic-xcvrd/xcvrd/sff_mgr.py#L439-L449) (as part of which, xcvr API [`set_lpmode()`](https://github.com/sonic-net/sonic-platform-common/blob/45b10beaabd9bd62c06b4cccdc7a497fc7c30fa2/sonic_platform_base/sonic_xcvr/api/public/sff8636.py#L388) gets called setting power_override bit). There's no need to do extra setting for that bit as part of high power class enabling.

image
Also from functionality POV, would it make more sense to decouple power_override bit from handling of high_power_class? Because, according to the truth table 6-11, once high power class (byte 93 bit 2 or/and bit 3) gets set properly, the module will function in the desired power class after the module is brought out of lpmode by either HW control or SW control.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -2322,10 +2318,10 @@ def run(self):
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('--skip_cmis_mgr', action='store_true')
-    parser.add_argument('--enable_sff_mgr', action='store_true')
+    parser.add_argument('--enable_sff_mgr_controlled_tx', action='store_true')
```
prgeor, line 2321: @longhuan-cisco why we need this new flag? Can you point to the HLD in the PR description?

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -285,6 +289,32 @@ def convert_bool_array_to_bit_mask(self, bool_array):
             mask += (1 << i if flag else 0)
         return mask
 
+    def handle_high_power_class(self, xcvr_api, lport):
```
prgeor, line 292: @longhuan-cisco nit: Can you rename to `enable_high_power_class()`

prgeor:
@longhuan-cisco Why the SFF manager is calling  lpmode for SFF8472 when these SFPs do'nt suppot? See
https://github.com/sonic-net/sonic-platform-common/pull/512

sonic-xcvrd/xcvrd/sff_mgr.py
prgeor: @longhuan-cisco LGTM that we do power override in set_lpmode(False) case and enable high power class > 5 later on.

longhuan-cisco:
> @longhuan-cisco Why the SFF manager is calling lpmode for SFF8472 when these SFPs do'nt suppot? See [sonic-net/sonic-platform-common#512](https://github.com/sonic-net/sonic-platform-common/pull/512)

The below code was added via https://github.com/sonic-net/sonic-platform-daemons/pull/565:
https://github.com/sonic-net/sonic-platform-daemons/blob/fa41b9ce90ed77783e60d1fef1b419fd3283a8bb/sonic-xcvrd/xcvrd/sff_mgr.py#L444-L445

prgeor (approved):

PR #577, opened 2024-12-16
Fix removal of the package during tests

saiarcot895:
#### Description

When installing the package's dependencies for testing purposes, the package itself needs to be removed afterwards, so that there isn't any stale content there. This needs to use the package name, which is different for each package here.

Fix the pipeline to remove the package properly.


prgeor:
@saiarcot895 are you planning to make it ready for review?

PR #584, opened 2025-01-16
Fix for dpu midplane state update affecting dpu control_plane/data_plane state :issue-21371

rameshraghupathy:
Fix for dpu midplane state update affecting dpu control_plane/data_plane state

#### Description
The dpu midplane update and the dpu cp/dp state update happen in parallel. The midplane state update function reads the data from redis-DB, udpates the midplane state and then writes back the entire table. It appears like the DPU cp/dp state update sometimes goes in between the midplane read and and write causing a wrong update of dpu cp/dp states. The fix is to avoid reading the table and just write only the dpu midplane state.  This will solve the problem.

#### Motivation and Context
The dpu midplane update and the dpu cp/dp state update happen in parallel. The midplane state update function reads the data from redis-DB, udpates the midplane state and then writes back the entire table. It appears like the DPU cp/dp state update sometimes goes in between the midplane read and and write causing a wrong update of dpu cp/dp states. The fix is to avoid reading the table and just write only the dpu midplane state.  This will solve the problem.

Toggle the DPU on/off state multiple times and see if the state is reflected properly.  Since, this code path is nvidia specific relying on their testing as well.

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -765,28 +765,22 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
 
     def update_dpu_state(self, key, state):
         """
-        Update DPU state in chassisStateDB using the given key.
+        Update specific DPU state fields in chassisStateDB using the given key.
         """
         try:
             # Connect to the CHASSIS_STATE_DB using daemon_base
             if not self.chassis_state_db:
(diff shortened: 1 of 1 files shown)


gpunathilell (approved):

vvolam (approved):
LGTM

prgeor (approved):

PR #585, opened 2025-01-29
thermalctld: Ignore exception when deleting chassisdb entry fails

patrickmacarthur:
#### Description

After sonic-mgmt tests, it does a config consistency check and if the config consistency check fails, it runs recover_chassis(). That function simultaneously runs config_reload on the supervisor and all linecards (in parallel). This means that the supervisor networking can be in the process of restarting when thermalctld exits and goes through the cleanup process of deleting entries from chassisdb.

#### Motivation and Context

This fixes a thermalctld crash when the supervisor and linecards simultaneously do a `config reload` (fixes sonic-net/sonic-buildimage#21058). This causes a failure during sonic-mgmt testing.

This change has been applied in Arista internal test environment and the thermalctl crash was not seen with this change applied.

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -559,8 +559,14 @@ class TemperatureUpdater(logger.Logger):
             table_keys = self.table.getKeys()
             for tk in table_keys:
                 self.table._del(tk)
-                if self.is_chassis_upd_required and self.chassis_table is not None:
-                    self.chassis_table._del(tk)
+                try:
+                    if self.is_chassis_upd_required and self.chassis_table is not None:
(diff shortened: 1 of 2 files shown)


judyjoseph (dismissed):
LGTM

judyjoseph:
@patrickmacarthur could you help improve test coverage before we merge ? 

```
Pull Request Coverage
Total: 5 lines
Missing: 2 lines
Coverage: 60%
Threshold: 80%
[Diff coverage](https://dev.azure.com/mssonic/be1b070f-be15-4154-aade-b1d3bfb17054/_build/results?buildId=758998&view=codecoverage-tab)
```

prgeor (approved):

judyjoseph (approved):

PR #586, opened 2025-02-04
Add NOTICE log for flat memory to skip CMIS FSM

AnoopKamath:
#### Description
no suitable app for the port appl ERROR message seen for flat memory modules

https://migsonic.atlassian.net/browse/MIGSMSFT-868

#### Motivation and Context
Although the DAC and AOC modules link up after " no suitable app for the port appl" error but it is flagged in SONIC management runs. This should be advertised in Page 0 Byte 2 bit 7 but some modules don't and cause this failure message. Add a NOTICE message for this case.

Verify with DAC and AOC cables where we see this message.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1270,8 +1270,13 @@ class CmisManagerTask(threading.Thread):
                     if state == CMIS_STATE_INSERTED:
                         self.port_dict[lport]['appl'] = get_cmis_application_desired(api, host_lane_count, host_speed)
                         if self.port_dict[lport]['appl'] is None:
-                            self.log_error("{}: no suitable app for the port appl {} host_lane_count {} "
-                                            "host_speed {}".format(lport, appl, host_lane_count, host_speed))
+			    if api.get_module_media_type() == 'passive_copper_media_interface' or \
+                               api.get_module_media_type() == 'active_cable_media_interface':
+                                self.log_notice("{}: no suitable app for the port appl {} host_lane_count {} host_speed {} skipping CMIS FSM "
+                                                "for media_type {}".format(lport, appl, host_lane_count, host_speed, api.get_module_media_type()))
+                            else:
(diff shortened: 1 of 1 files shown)


mihirpat1:
@AnoopKamath Agreed - flat memory modules should not go through the CMIS initialization sequence itself. Wondering why such modules are not being handled in the below block which is executed before CMIS initialization sequence begins?

https://github.com/sonic-net/sonic-platform-daemons/blob/8c89f6ba2975a7699861f2c6d77083cebb62e97c/sonic-xcvrd/xcvrd/xcvrd.py#L1373-L1376

prgeor:
@mihirpat1 can you review?

mihirpat1:
> @mihirpat1 can you review?

Yes - already provided the below comment
https://github.com/sonic-net/sonic-platform-daemons/pull/586#issuecomment-2634905697

AnoopKamath:
> @AnoopKamath Agreed - flat memory modules should not go through the CMIS initialization sequence itself. Wondering why such modules are not being handled in the below block which is executed before CMIS initialization sequence begins?
> 
> https://github.com/sonic-net/sonic-platform-daemons/blob/8c89f6ba2975a7699861f2c6d77083cebb62e97c/sonic-xcvrd/xcvrd/xcvrd.py#L1373-L1376

@mihirpat1 : Thanks, I have seen certain cases where some of DAC cables don't advertise as flat memory module in Page 0 offset 2 bit 7. But this can be genuine error in case of AOC modules. I am trying to find some modules that have these problems. Also, **test_iface_namingmode.py** script throws these errors when it tries to set the module speed based on interface supported speed. This is wrong since not all modules support all speeds. I am looking into it as well

prgeor (changes requested):
@AnoopKamath can you get the module fixed by the vendor to follow CMIS spec

AnoopKamath:
> @AnoopKamath can you get the module fixed by the vendor to follow CMIS spec

@prgeor : that was my first suggestion for this issue. Also, will request SONIC management runs to use module which supports all modes. I will close this PR. Thanks

PR #587, opened 2025-02-07
Credo AEC 400G firmware v4.1 needs to set low power mode when change to DPDeinit state

CharlieChenEC:
Credo AEC 400G firmware v4.1 needs to set low power mode when change to DPDeinit state

Description
Credo AEC 400G needs to set low power mode when changing to DPDeinit state on firmware v4.1. After https://github.com/sonic-net/sonic-platform-daemons/pull/254 is merged, low power mode will not be set when changing to DPDeinit state.

Motivation and Context
This patch adds the code to set low power mode only when changing to DPDeinit state when the transceiver's vendor is 'Credo' and 'host_assign' is '1'(Which means the application mode only has 1 logical port on the transceiver). With this patch, the Credo AEC 400G can work correctly.

How Has This Been Tested?
Insert AEC 400G on AS9716 and verify the cable is link up.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1144,6 +1144,31 @@ class CmisManagerTask(threading.Thread):
             if key in ["PortConfigDone", "PortInitDone"]:
                 break
 
+    def need_lp_mode_for_dpdeinit(self, api, appl):
+        """
+        Determine whether the cmis transceiver needs to enter
+        low power mode when the state is changed to DPDeinit.
+        """
+        host_assign = api.get_host_lane_assignment_option(appl)
+        vendor = api.get_manufacturer().strip()
+
+        if vendor == "Credo" and host_assign == 0x1:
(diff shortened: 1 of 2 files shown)


mihirpat1:
@CharlieChenEC Please fix the coverage check

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1144,6 +1144,31 @@ def wait_for_port_config_done(self, namespace):
             if key in ["PortConfigDone", "PortInitDone"]:
                 break
 
+    def need_lp_mode_for_dpdeinit(self, api, appl):
```
prgeor, line 1147: @CharlieChenEC I don't think we can keep add vendor specific checks in a code that is generic and expects all modules to follow CMIS spec

prgeor (changes requested):
@CharlieChenEC please get the module updated to the firmware that is fully CMIS compliant. A module should honor low power mode at any time.

PR #598, opened 2025-03-20
Add exception handling and calculation of the rate for each individual lane

micas-net:
#### Why I did it
The CMIS protocol type optical module fails to obtain the single-lane rate accurately

##### Work item tracking
- Microsoft ADO **(number only)**:

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202505

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ]

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
+++ b/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
@@ -78,8 +78,8 @@ def get_lane_speed_key(physical_port, port_speed, lane_count):
             host_electrical_interface_id = appl_adv_dict[app_id].get('host_electrical_interface_id')
             if host_electrical_interface_id:
                 lane_speed_key = LANE_SPEED_KEY_PREFIX + host_electrical_interface_id.split()[0]
-            if lane_speed_key is None or str(port_speed//1000) in lane_speed_key:
-                lane_speed_key = '{}{}G'.format(LANE_SPEED_KEY_PREFIX, port_speed // lane_count // 1000)
+        if lane_speed_key is None or str(port_speed//1000) in lane_speed_key:
+            lane_speed_key = '{}{}G'.format(LANE_SPEED_KEY_PREFIX, port_speed // lane_count // 1000)
     else:
         # Directly calculate lane speed and use it as key, this is especially useful for
         # non-CMIS transceivers which typically have no host_electrical_interface_id
--- a/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
+++ b/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
@@ -78,6 +78,8 @@ def get_lane_speed_key(physical_port, port_speed, lane_count):
             host_electrical_interface_id = appl_adv_dict[app_id].get('host_electrical_interface_id')
             if host_electrical_interface_id:
                 lane_speed_key = LANE_SPEED_KEY_PREFIX + host_electrical_interface_id.split()[0]
+            if lane_speed_key is None or str(port_speed//1000) in lane_speed_key:
+                lane_speed_key = '{}{}G'.format(LANE_SPEED_KEY_PREFIX, port_speed // lane_count // 1000)
     else:
         # Directly calculate lane speed and use it as key, this is especially useful for
         # non-CMIS transceivers which typically have no host_electrical_interface_id
--- a/sonic-xcvrd/tests/test_xcvrd.py
+++ b/sonic-xcvrd/tests/test_xcvrd.py
@@ -1051,7 +1051,7 @@ class TestXcvrdScript(object):
 
         mock_api.get_application_advertisement = MagicMock(return_value=mock_app_adv_value)
         result = media_settings_parser.get_media_settings_key(0, xcvr_info_dict, 100000, 2)
-        assert result == { 'vendor_key': 'MOLEX-1064141421', 'media_key': 'QSFP-DD-sm_media_interface', 'lane_speed_key': 'speed:100GBASE-CR2' }
+        assert result == { 'vendor_key': 'MOLEX-1064141421', 'media_key': 'QSFP-DD-sm_media_interface', 'lane_speed_key': 'speed:50G' }
 
     @pytest.mark.parametrize("data_found, data, expected", [
         (True, [('speed', '400000'), ('lanes', '1,2,3,4,5,6,7,8'), ('mtu', '9100')], (400000, 8, 0)),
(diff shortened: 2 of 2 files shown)


sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
```diff
@@ -78,6 +78,8 @@ def get_lane_speed_key(physical_port, port_speed, lane_count):
             host_electrical_interface_id = appl_adv_dict[app_id].get('host_electrical_interface_id')
             if host_electrical_interface_id:
                 lane_speed_key = LANE_SPEED_KEY_PREFIX + host_electrical_interface_id.split()[0]
+            if lane_speed_key is None or str(port_speed//1000) in lane_speed_key:
```
prgeor, line 81: @philo-micas Why you want to assume the lane speed from the configured speed and ignore the module capability to support the configured speed?

philo-micas: Because the function of this function interface is to obtain the rate of a single lane. The capacity obtained by the module is generally for the optical module, not a single lane, so secondary processing is required. In addition, this code is only added for abnormal data processing and does not affect normal data.

prgeor: @philo-micas can you give me an example where this secondary processing is required?

philo-micas: Here are examples of different optical module information retrieved before the modification. 
The variable "lane_speed_key" is the return value of this function interface. 
Sometimes it is None; sometimes it represents the port rate instead of the rate of a single lane. Therefore, it is necessary to handle the abnormal values.
eg：
{'vendor_key': 'LUXSHARE-ICT    -LPZDD002-SD-R   ', 'media_key': 'QSFP-DD-passive_copper_media_interface-1.0M', 'lane_speed_key': None}
{'vendor_key': 'LUXSHARE-ICT    -LPZDD002-SD-R   ', 'media_key': 'QSFP-DD-passive_copper_media_interface-1.0M', 'lane_speed_key': 'speed:400G'}
{'vendor_key': 'BIZLINK         -C45593-A557-D10 ', 'media_key': 'QSFP-DD-passive_copper_media_interface-1.0M', 'lane_speed_key': 'speed:400G'}
{'vendor_key': 'OEM             -QSFPDD-400G-CU1M', 'media_key': 'QSFP-DD-passive_copper_media_interface-1.0M', 'lane_speed_key': None}

prgeor: @micas-net looks like you are using a CMIS compliant QSFP-DD passive module. Since [get_host_electrical_interface()](https://github.com/sonic-net/sonic-platform-common/blob/b536f2cb229add2f82dde85bb331d29266ad0bae/sonic_platform_base/sonic_xcvr/api/public/cmis.py#L965) returns N/A for flat memory probably that is the reason `lane_speed_key` is not derived from host electrical interface id. Can you remove the `is_flat_memory()` check in `get_host_electrical_interface` and try?

micas-net: Thank you for your advice. Removing the is_flat_memory() check in get_host_electrical_interface() does solve the problem of get_host_electrical_interface() returning N/A.
However, considering the port breakout scenario, we still cannot use the solution you provided.
When configure ports breakout using a normal QSFP-DD optical module, xcvrd.get_cmis_application_desired() returns None.
Therefore, in this scenario the lane_speed_key of the breakout port cannot be obtained from the eeprom of the optical module as the old solution.
So, if the correct lane_speed_key cannot be obtained from the eeprom of the optical module, 
the lane_speed_key must be calculated by port_speed/lane_count. Avoid lane_speed_key return None.

prgeor:
@micas-net merge conflict.

micas-net:
@prgeor

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
prgeor: @micas-net I am not able to understand the problem you are facing. I am not sure why xcvrd.get_cmis_application_desired() is returning None for your module. Also please add more details to the PR description 


Why I did it        ==> ADD more details
The CMIS protocol type optical module fails to obtain the single-lane rate accurately


How I did it
How to verify it

micas-net: The purpose of the "get_lane_speed_key" function is to obtain the rate of a single lane rather than the capability of the optical module. 
Parameters such as pre-emphasis are adjusted based on the return value of the "get_lane_speed_key" function. 
However, the following two scenarios are inconsistent with the intended design:

1、First Scenario:
In the following example, the 'lane_speed_key' is set to speed:400G. However, 400G represents the capability of the optical module, not the rate of a single lane.
Since QSFP-DD modules have 8 lanes, the 'lane_speed_key' should correspond to the per-lane speed, which is 50G instead of 400G.
Example:
{'vendor_key': 'BIZLINK -C45593-A557-D10', 'media_key': 'QSFP-DD-passive_copper_media_interface-1.0M', 'lane_speed_key': 'speed:400G'}

2、Second Scenario:
When configuring a breakout port, for instance, using the command:
"config interface breakout Ethernet1 '4x100G' -yf"
In the following example, the 'lane_speed_key' is set to None because the optical module does not support port breakout.
While it is not entirely reasonable to configure a breakout port when the optical module lacks support for it, but users may choose to configure the breakout port first and replace the optical module later.
This sequence of actions can lead to incorrectly configured pre-emphasis parameters, potentially causing system instability or even coredumps.
To enhance system robustness under such scenarios, the function should still return the single lane rate of 50G instead of None.
Example:
{'vendor_key': 'LUXSHARE-ICT -LPZDD002-SD-R', 'media_key': 'QSFP-DD-passive_copper_media_interface-1.0M', 'lane_speed_key': None}

The above two scenarios highlight the reasons behind the modifications made to the "get_lane_speed_key" function.
The key point is that the "get_lane_speed_key" function is designed to return the rate of a single lane, not the overall capability of the optical module.
Therefore, it is essential to ensure that the 'lane_speed_key' function always returns the single lane rate, regardless of the capability of the optical module.

AndrewQu-Micas (approved):
the changes look good to me

micas-net:
@prgeor If there are new comments,please help push it to merge state, thanks!

micas-net:
will close this as https://github.com/sonic-net/sonic-buildimage/pull/22637 can fix it.

PR #600, opened 2025-03-21
Sensormond changes to support configurable time for logging threshold and better handling of fatal signals

gregoryboudreau:
#### Description

Two quality of life improvements for sensormond:

1. Allow for the user to configure timeout for printing time log warning (keep default at 30), this is set in the the platform_env.conf through the `SENSORMOND_WARNING_TIME` value
2. Cascade stop events (from signal handler) into voltage and current updaters to allow for more graceful checking _during_ the update process, current setup has 10 second timeout and given the updaters can take some time to run can result in SIGKILLs getting sent during operations.

#### Motivation and Context

With a lot of sensors, the logs can be filled with warnings about exceeding the time to read even if it is expected by the vendor. Additionally, with this longer time, the SIGTERM can not be checked in enough time before supervisorctl falls back to using a SIGKILL

Tested on Cisco Smartswitch, time configuration now no longer results in logs being triggered every loop and tested killing w/ `supervisorctl stop sensormond` and no longer see any error logs from system that were being hit when SIGKILL interrupted reads from driver.

### DIFF
--- a/sonic-sensormond/scripts/sensormond
+++ b/sonic-sensormond/scripts/sensormond
@@ -10,7 +10,7 @@ import sys
 import threading
 import time
 import yaml
-import argparse
+import os
 
 import sonic_platform
 from sonic_py_common import daemon_base, logger, device_info
@@ -22,6 +22,8 @@ SYSLOG_IDENTIFIER = 'sensormond'
 NOT_AVAILABLE     = 'N/A'
 CHASSIS_INFO_KEY  = 'chassis 1'
 INVALID_SLOT      = -1
+
+PLATFORM_ENV_CONF_FILE = "/usr/share/sonic/platform/platform_env.conf"
 
 PHYSICAL_ENTITY_INFO_TABLE = 'PHYSICAL_ENTITY_INFO'
 
@@ -201,7 +203,6 @@ class VoltageUpdater(SensorUpdater):
         self.log_debug("Start voltage update")
 
         sensor_list = self.fs_sensors + self.chassis.get_all_voltage_sensors()
-
         for index, voltage_sensor in enumerate(sensor_list):
             if(stop_event_signal.is_set()):
                 self.log_info("Stop signal received while running voltage update")
@@ -461,8 +462,9 @@ class SensorMonitorDaemon(daemon_base.DaemonBase):
     INITIAL_INTERVAL = 5
     # Periodic Update interval 
     UPDATE_INTERVAL = 60
+    UPDATE_ELAPSED_THRESHOLD = 30
 
-    def __init__(self, update_elapsed_threshold):
+    def __init__(self):
         '''
         Initializer of SensorMonitorDaemon
         '''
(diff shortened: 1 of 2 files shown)


arlakshm:
update platform specific file with the new timeout value iso of updating the supervisord.conf

anamehra:
As discussed, please checl platform_env.conf in platform dir to provide any platform config.

rlhui:
@kenneth-arista will check as well.

judyjoseph:
@bmridul @anamehra  please can you discuss on this internally and get back

gregoryboudreau:
ack, am following up internally, will update this PR by next week

mlok-nokia (approved):

abdosi:
@bmridul : code coverage checker failing.

abdosi:
@judyjoseph : to help review on this.

sonic-sensormond/scripts/sensormond
```diff
@@ -22,6 +23,8 @@ NOT_AVAILABLE     = 'N/A'
 CHASSIS_INFO_KEY  = 'chassis 1'
 INVALID_SLOT      = -1
 
+PLATFORM_ENV_CONF_FILE = "/usr/share/sonic/platform/platform_env.conf"
```
Staphylo, line 26: This configuration file name is already taken by broadcom so we should probably rename it to something different.
Unless this was thought through and we feel it's fine for a config file to be shared by both asic drivers and generic daemons.
I believ we should CC someone from Broadcom to make sure they're ok with that.
https://github.com/sonic-net/sonic-buildimage/blob/e74ef02efdb3a76803124e961f02ef3f7c756867/platform/broadcom/saibcm-modules/debian/opennsl-modules.init#L98

prabhataravind:
@kperumalbfn could you please approve this for 202411? @rameshraghupathy for viz

rameshraghupathy:
> > > > @prabhataravind @rameshraghupathy This is common for all SKUs, do we need this in 202411? Could you please check smartswitch SKUs with this change.
> > > 
> > > 
> > > @kperumalbfn @prabhataravind This is not must as the sonic-mgmt already skipping this error message. As long as we have that PR included we don't need this. More over there is a platform counterpart to it and that is not committed to 202411 branch. So, we can skip this.
> > 
> > 
> > @rameshraghupathy what is the other PR in sonic-mgmt to skip this error? Could you please add it here?
>  @prabhataravind @yejianquan  Here is the PR: https://github.com/sonic-net/sonic-mgmt/pull/17263  in sonic-mgmt to skip this error
> Hi @prabhataravind @rameshraghupathy , please also confirm whether this change is a must to go to 202505
@yejianquan @prabhataravind We don't need this PR in 202411.

yejianquan:
remove the request label, it's not needed in 202411, suppose 202505 don't need it, please add label again if needed

gregoryboudreau:
it addresses the underlying failure that the sonic-mgmt ignore modification covers, if not a major headache to bringing back I'd say it's preferable to have it in 202411 and 202505 but not a _must_ have.

sonic-sensormond/scripts/sensormond
```diff
@@ -522,8 +552,9 @@ class SensorMonitorDaemon(daemon_base.DaemonBase):
 
         begin = time.time()
 
-        self.voltage_updater.update()
-        self.current_updater.update()
+        if not (self.voltage_updater.update(self.stop_event) or self.current_updater.update(self.stop_event)):
+            # Fatal signal received in the process of reading voltage or current sensors
+            return False
```
gordon-nexthop, line 557: When testing on my several of our devices if `self.voltage_updater.upate` evalutates as true then `self.current_updater.update` will not execute. Recommend separating the two update functions as beforehand.

gordon-nexthop: Thank you!

PR #601, opened 2025-03-25
chassisd: add fabric module information to STATE_DB

ymd-arista:
#### Description

1. Add Fabric Card Module information to STATE_DB. This information
is necessary as in the tests we expect for fans under fabric cards,
their parent information (i.e fabric cards) is also present in the
`PHYSICAL_ENTITY_INFO` of `STATE_DB`.

2. Rather than storing a module by its index, store it by its name.
This makes it easier to refer to modules as sometimes index/position
of module may change, but it's name remains consistent.

#### Motivation and Context

This was done as part of changes to fix https://github.com/sonic-net/sonic-mgmt/issues/16189,
where the test was failing as we expect parent of fan to be present, but
for fans under fabric modules, fabric modules were not present in `STATE_DB`.

Added a unit-test which tests whether `PHYSICAL_ENTITY_INFO` is updated
when a fabric module is inserted.

Also, tested manually by checking `STATE_DB` on internal machines.

Other PR's of interest for this change: https://github.com/sonic-net/sonic-snmpagent/pull/349, https://github.com/sonic-net/sonic-mgmt/pull/17686

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -56,6 +56,9 @@ CHASSIS_MODULE_INFO_OPERSTATUS_FIELD = 'oper_status'
 CHASSIS_MODULE_INFO_NUM_ASICS_FIELD = 'num_asics'
 CHASSIS_MODULE_INFO_ASICS = 'asics'
 CHASSIS_MODULE_INFO_SERIAL_FIELD = 'serial'
+CHASSIS_MODULE_INFO_PRESENCE_FIELD = 'presence'
+CHASSIS_MODULE_INFO_MODEL_FIELD = 'model'
+CHASSIS_MODULE_INFO_REPLACEABLE_FIELD = 'is_replaceable'
 
 CHASSIS_ASIC_INFO_TABLE = 'CHASSIS_ASIC_TABLE'
 CHASSIS_FABRIC_ASIC_INFO_TABLE = 'CHASSIS_FABRIC_ASIC_TABLE'
@@ -79,6 +82,8 @@ DEFAULT_LINECARD_REBOOT_TIMEOUT = 180
 DEFAULT_DPU_REBOOT_TIMEOUT = 360
 PLATFORM_ENV_CONF_FILE = "/usr/share/sonic/platform/platform_env.conf"
 PLATFORM_JSON_FILE = "/usr/share/sonic/platform/platform.json"
+
+PHYSICAL_ENTITY_INFO_TABLE = 'PHYSICAL_ENTITY_INFO'
 
 CHASSIS_INFO_UPDATE_PERIOD_SECS = 10
 CHASSIS_DB_CLEANUP_MODULE_DOWN_PERIOD = 30 # Minutes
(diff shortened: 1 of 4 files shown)


mlok-nokia (approved):
look ok to me

sonic-chassisd/scripts/chassisd
```diff
@@ -146,6 +151,15 @@ def get_formatted_time(datetimeobj=None, op_format=None):
     date_obj = datetimeobj if datetimeobj else datetime.now(timezone.utc)
     return date_obj.strftime(op_format if op_format else "%a %b %d %I:%M:%S %p UTC %Y")
 
+def update_entity_info(table, parent_name, key, index, serial, model, is_replaceable):
+    fvs = swsscommon.FieldValuePairs(
+        [('position_in_parent', str(index)),
```
judyjoseph, line 156: is this module_index 0 based or 1 based

judyjoseph (approved):
LGTM , just cross check the module index .. as I have seen places where we do a +1

sonic-chassisd/scripts/chassisd
ymd-arista: It's `0` based.

In `chassisd`, even when we try to get module information via `_get_module_info`, we use `module_index` rather than `module_index + 1`. So, `module_index` should indicate the right position of module in the Chassis.

anamehra:
Hi @ymd-arista , could you please add more details why we can't use the data in Chassis module table on redis_chassis.server? Please dd UT logs showing the dump of dbs etc.

ymd-arista:
@anamehra : Our understanding is that PHYSICAL_ENTITY_TABLE should basically be a mirror/staging area for everything that we export via SNMP as opposed to having SNMP having to look in multiple places for all of the info it needs.

DB dumps:

1. After adding some more fields about module to 'CHASSIS_MODULE_TABLE`

`CHASSIS_MODULE_TABLE|FABRIC-CARD0: {'desc': 'DCS-7804-FM', 'slot': '51', 'oper_status': 'Online', 'num_asics': '2', 'serial': 'SSN19430153', 'presence': 'True', 'is_replaceable': 'True', 'model': 'DCS-7804-FM'}`

`CHASSIS_MODULE_TABLE|LINE-CARD1: {'desc': 'Unknown', 'slot': '4', 'oper_status': 'Empty', 'num_asics': '0', 'serial': 'N/A', 'presence': 'False', 'is_replaceable': 'True', 'model': 'N/A'}`

2. After adding information about module to `PHYSICAL_ENTITY_INFO`

`root@nfc405:/home/admin# sonic-db-cli STATE_DB hgetall "PHYSICAL_ENTITY_INFO|fan5/1"
{'position_in_parent': '1', 'parent_name': 'FABRIC-CARD5'}`

`root@nfc405:/home/admin# sonic-db-cli STATE_DB hgetall "PHYSICAL_ENTITY_INFO|FABRIC-CARD5"
{'position_in_parent': '6', 'parent_name': 'chassis 1', 'serial': 'SSN20420826', 'model': '7808R3-FM', 'is_replaceable': 'True'}`

PR #613, opened 2025-05-09
[xcvrd] do not wait state change while calling cmis.set_lpmode

Junchao-Mellanox:
Depends on https://github.com/sonic-net/sonic-platform-common/pull/564

#### Description
Ignore time wait in `cmis.set_lpmode`

#### Motivation and Context
In `CmisManagerTask`, it calls `cmis.set_lpmode` for each logical port. The time wait causes a `wait_time * num_of_logical_port` delay in link up time. The PR is to optimize it.

Manual test
unit test

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1268,7 +1268,7 @@ class CmisManagerTask(threading.Thread):
                             continue
 
                         #Sets module to high power mode and doesn't impact datapath if module is already in high power mode
-                        api.set_lpmode(False)
+                        api.set_lpmode(False, wait_state_change = False)
                         self.update_port_transceiver_status_table_sw_cmis_state(lport, CMIS_STATE_AP_CONF)
                         dpDeinitDuration = self.get_cmis_dp_deinit_duration_secs(api)
                         modulePwrUpDuration = self.get_cmis_module_power_up_duration_secs(api)


keboliu (approved):

mihirpat1:
@Junchao-Mellanox Can you please help in answering the below question in reference to the PR description?
```
In CmisManagerTask, it calls cmis.set_lpmode for each logical port. The time wait causes a wait_time * num_of_logical_port delay in link up time. The PR is to optimize it.
```

Question
1. From my understanding, the wait time should only apply to the first subport of a physical port during CMIS initialization. For the remaining subports, the port should already be in high power mode, making the wait effectively a NOOP.
Is the goal of this PR to reduce CMIS initialization time by allowing subsequent logical ports to proceed without waiting for the high power mode transition?

2. Could you also share any performance metrics or stats that demonstrate the improvement from this change?

liat-grozovik:
@prgeor kindly reminder to provide your feedback on this PR

prgeor:
> @prgeor kindly reminder to provide your feedback on this PR

@liat-grozovik waiting for answers to Mihirs question

prgeor:
> @Junchao-Mellanox Can you please help in answering the below question in reference to the PR description?
> 
> ```
> In CmisManagerTask, it calls cmis.set_lpmode for each logical port. The time wait causes a wait_time * num_of_logical_port delay in link up time. The PR is to optimize it.
> ```
> 
> Question
> 
> 1. From my understanding, the wait time should only apply to the first subport (specifically, the first subport to go through CMIS init for a breakout port group) of a physical port during CMIS initialization. For the remaining subports, the port should already be in high power mode, making the wait effectively a NOOP.
>    Is the goal of this PR to reduce CMIS initialization time by allowing subsequent logical ports to proceed without waiting for the high power mode transition?
> 2. Could you also share any performance metrics or stats that demonstrate the improvement from this change?

@mihirpat1 [1]. I think at present set_lpmode() is a busy wait function, which means when you have 8 subports, Cmis Mgr will wait for module to transition from `LpMode` -> `ModuelReady`. This PR is removing this busy wait and making use of `CMIS_STATE_AP_CONF` state where Xcvrd checks for `ModuleReady` in a non-busy way.

@Junchao-Mellanox We need to make sure

1. If module is transitioning from `LpMode` -> `ModuleReady` , and there is a set_lpmod(False) called again(breakout scenario), does that impact the ongoing transition?
2. If there is a timeout for `ModuleReady` , CMIS mgr does `force_cmis_reinit()`, can you confirm if this will eventually help to move the module to `ModuleReady` ?

Junchao-Mellanox:
Hi @mihirpat1 ,

Due my test, set_lpmode takes ~0.5s when there is wait. And it takes ~0.01s when there is no wait. After applying this change, the link up time for 512 ports saves 20~30 seconds. You are right about the subport: only the first subport executes the wait. But it still took 64 * 0.5 = ~32 seconds. Also, the idea is that we don't need this wait in set_lpmode function since CmisManagerTask has already done the wait in state machine.

Junchao-Mellanox:
> > @Junchao-Mellanox Can you please help in answering the below question in reference to the PR description?
> > ```
> > In CmisManagerTask, it calls cmis.set_lpmode for each logical port. The time wait causes a wait_time * num_of_logical_port delay in link up time. The PR is to optimize it.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Question
> > 
> > 1. From my understanding, the wait time should only apply to the first subport (specifically, the first subport to go through CMIS init for a breakout port group) of a physical port during CMIS initialization. For the remaining subports, the port should already be in high power mode, making the wait effectively a NOOP.
> >    Is the goal of this PR to reduce CMIS initialization time by allowing subsequent logical ports to proceed without waiting for the high power mode transition?
> > 2. Could you also share any performance metrics or stats that demonstrate the improvement from this change?
> 
> @mihirpat1 [1]. I think at present set_lpmode() is a busy wait function, which means when you have 8 subports, Cmis Mgr will wait for module to transition from `LpMode` -> `ModuelReady`. This PR is removing this busy wait and making use of `CMIS_STATE_AP_CONF` state where Xcvrd checks for `ModuleReady` in a non-busy way.
> 
> @Junchao-Mellanox We need to make sure
> 
> 1. If module is transitioning from `LpMode` -> `ModuleReady` , and there is a set_lpmod(False) called again(breakout scenario), does that impact the ongoing transition?
> 2. If there is a timeout for `ModuleReady` , CMIS mgr does `force_cmis_reinit()`, can you confirm if this will eventually help to move the module to `ModuleReady` ?

Hi @prgeor ,

For the first question, I suppose it does not matter. If user set module to low power mode during the transitioning from `LpMode` -> `ModuleReady`, the busy wait does not help either. Because set_lpmod(False) could be called right after busy wait succesfully return.

For the second question, my test never hit the timeout for `ModuleReady`. It usually takes no more than 0.5s to reach `ModuleReady`. I think if we reach such timetout, the issue is on module side.

Junchao-Mellanox:
I will test and get back to you.

mihirpat1 (approved):

prgeor (approved):

PR #616, opened 2025-05-20
xcvrd: Remove SFP API object when SFP is removed

jianyuewu:
Depends on: https://github.com/sonic-net/sonic-platform-common/pull/562

When an SFP is physically removed from a port, the SFP API object should be removed from the sfp_obj_dict to prevent stale object references and ensure proper cleanup. This change ensures that when the SFP status is detected as removed, the corresponding SFP API object is properly deleted from the dictionary.

#### Description

Add check for SFP_STATUS_REMOVED in SfpStateUpdateTask.
Remove SFP API object from sfp_obj_dict when SFP is removed.
Add unit test to verify SFP object removal functionality.

#### Motivation and Context

Make sure SFP API object is removed when SFP removed, so cmis cache can also be re-created.

Tested with config reload.

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd.py
+++ b/sonic-xcvrd/xcvrd/xcvrd.py
@@ -1735,6 +1735,12 @@ class SfpStateUpdateTask(threading.Thread):
                                     media_settings_parser.notify_media_setting(logical_port, transceiver_dict, self.xcvr_table_helper, self.port_mapping)
                                     transceiver_dict.clear()
                             elif value == sfp_status_helper.SFP_STATUS_REMOVED:
+                                # Remove the SFP API object for this physical port
+                                try:
+                                    sfp = platform_chassis.get_sfp(int(key))
+                                    sfp.remove_xcvr_api()
+                                except (NotImplementedError, AttributeError) as e:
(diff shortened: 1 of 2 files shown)


keboliu (approved):

Junchao-Mellanox (approved):

keboliu:
@prgeor would you please review?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1735,6 +1735,12 @@ def task_worker(self, stopping_event, sfp_error_event):
                                     media_settings_parser.notify_media_setting(logical_port, transceiver_dict, self.xcvr_table_helper, self.port_mapping)
                                     transceiver_dict.clear()
                             elif value == sfp_status_helper.SFP_STATUS_REMOVED:
+                                # Remove the SFP API object for this physical port
+                                try:
```
prgeor, line 1739: @jianyuewu need to make sure the API object held by other thread can become invalid. Did you test the scenario?

jianyuewu: image

mihirpat1 (approved):

PR #628, opened 2025-06-21
Remove space from the vendor key used for obtaining media si settings

prgeor:
#### Description
Remove space from the vendor key used for obtaining media si settings

#### Motivation and Context
To remove the vendor name + vendor PN where each of these have space character its better to remove them before using as the key because the media_settings.json will need the to speciy the key like

...
`'Eoptolink      -EO199HGPKT11PLQ  '`

instead of 

`'Eoptolink-EO199HGPKT11PLQ'`

Used custom SI settings with above module and ensured the Xcvrd is sending the correct SI settings to the orchagent.


prgeor:
@mihirpat1 can you review?

tshalvi (approved):

mihirpat1:
We no longer need this PR since the issue is fixed at the bottom layer now via the below PR
https://github.com/sonic-net/sonic-platform-common/pull/578

prgeor:
Thanks @mihirpat1 . closing this now.

PR #646, opened 2025-07-17
[chassis] modify chassisd to add fabric card information to PHYSICAL_ENTITY_INFO table only if module is present

vishal-nokia:
#### Description

Update chassisd to add fabric card module information to `PHYSICAL_ENTITY_INFO` table in `STATE_DB`  only if the module is present.

#### Motivation and Context

Fixes [#23372](https://github.com/sonic-net/sonic-buildimage/issues/23372)

Run `sonic-db-cli STATE_DB keys "PHYSICAL_ENTITY_INFO|FABRIC*`"`
It should only list the fabric card modules that are present.

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -407,14 +407,18 @@ class ModuleUpdater(logger.Logger):
                 prev_status = self.get_module_current_status(key)
                 self.module_table.set(key, fvs)
 
-                update_entity_info(self.phy_entity_table,
-                                   "chassis {}".format(1),
-                                   key,
-                                   module_index,
-                                   module_info_dict[CHASSIS_MODULE_INFO_SERIAL_FIELD],
-                                   module_info_dict[CHASSIS_MODULE_INFO_MODEL_FIELD],
-                                   module_info_dict[CHASSIS_MODULE_INFO_REPLACEABLE_FIELD])
-
+                if module_info_dict[CHASSIS_MODULE_INFO_PRESENCE_FIELD].lower() == "true":
+                    update_entity_info(self.phy_entity_table,
+                                       "chassis {}".format(1),
+                                       key,
+                                       module_index,
+                                       module_info_dict[CHASSIS_MODULE_INFO_SERIAL_FIELD],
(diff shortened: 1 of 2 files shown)


abdosi:
@judyjoseph @ymd-arista : please check.

sonic-chassisd/tests/test_chassisd.py
```diff
@@ -136,6 +136,12 @@ def test_moduleupdater_check_phyentity_fields():
     assert model == fvs[CHASSIS_MODULE_INFO_MODEL_FIELD]
     assert str(replaceable) == fvs[CHASSIS_MODULE_INFO_REPLACEABLE_FIELD]
 
+    presence = False
```
ymd-arista, line 139: Can we separate this into another test case?

Essentially, we would be testing two things in one and the name of the test doesn't signify the new changes we are testing here. The name just indicates that we are testing for the fields of an entry of fabric module.

Probably another new test case like `test_moduleupdater_check_phyentity_entry_after_fabric_removal`? or something like that?

sonic-chassisd/scripts/chassisd
```diff
@@ -157,14 +157,18 @@ def get_formatted_time(datetimeobj=None, op_format=None):
...
-def update_entity_info(table, parent_name, key, index, serial, model, is_replaceable):
-    fvs = swsscommon.FieldValuePairs(
-        [('position_in_parent', str(index)),
-         ('parent_name', parent_name),
-         ('serial', serial),
-         ('model', model),
-         ('is_replaceable', is_replaceable)])
-    table.set(key, fvs)
+def update_entity_info(table, parent_name, key, index, serial, model, is_replaceable, presence):
+    if presence.lower() == "true":
```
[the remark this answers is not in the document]
ymd-arista: If we go with this, we wouldn't need `presence` field in the PHYSICAL ENTITY TABLE as well.

sonic-chassisd/tests/test_chassisd.py
vishal-nokia: Pushed changes as per your comments. Please review.

abdosi (approved):

PR #653, opened 2025-07-31
Check low power mode before vdm handling

az-pz:

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd_utilities/utils.py
+++ b/sonic-xcvrd/xcvrd/xcvrd_utilities/utils.py
@@ -23,3 +23,14 @@ class XCVRDUtils:
         except (KeyError, NotImplementedError):
             self.logger.log_error(f"Failed to check flat memory for port {physical_port}")
             return True
+
+    def is_transceiver_lpmode_on(self, physical_port):
+        try:
+            api = self.sfp_obj_dict[physical_port].get_xcvr_api()
+            if not api:
+                return False
+            return api.get_lpmode()
+
+        except (KeyError, NotImplementedError):
+            self.logger.log_error(f"Failed to get low power mode for port {physical_port}")
+            return False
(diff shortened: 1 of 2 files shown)


sonic-xcvrd/xcvrd/xcvrd_utilities/utils.py
```diff
@@ -23,3 +23,14 @@ def is_transceiver_flat_memory(self, physical_port):
         except (KeyError, NotImplementedError):
             self.logger.log_error(f"Failed to check flat memory for port {physical_port}")
             return True
+
+    def is_transceiver_lpmode_on(self, physical_port):
```
prgeor, line 27: @mihirpat1 should we check instead the datapath is DatapathActivated state to do the VDM  collection instead?

mihirpat1: @prgeor I don't think we should check for DatapathActivated state since for a breakout port, if all subports except 1 subport is in shut state, there can be a mix of lanes being in DatapathActivated and DataPathDeactivated states.

mihirpat1 (approved):

sonic-xcvrd/xcvrd/xcvrd_utilities/utils.py
az-pz: @prgeor , This is a helper function which calls `get_lpmode()` of sfp api object on [line 32](https://github.com/sonic-net/sonic-platform-daemons/pull/653/files#diff-81b45260e57e4fb1b7ba51cdd8dee93da0dd35a977cff0731e31ee37663949fdR32)  . It also does some additional exception checking and logging around the api call. It is in line with other api calls which are wrapped in a helper method.

prgeor: @az-pz There are platforms that override the api.get_lpmode() function while  and use their own platform implementation API so in that case you are forcing the use of the api.get_lpmode. Eg here for mellanox:-

https://github.com/sonic-net/sonic-buildimage/blob/4c1f2569bb6ba75cb5962ba34acf0fa16370f4b5/platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py#L570

az-pz: I thought `api = self.sfp_obj_dict[physical_port].get_xcvr_api()` would get the api of the relevant platform and `get_lpmode` method would be called for the relevant platform and optic. 
What do you suggest we should do here instead?

mihirpat1:
@r12f @arlakshm @yejianquan - Can you please help to cherry pick this to 202412, 202503 and 202505?

PR #654, opened 2025-08-01
Refactor common xcvrd functions

bobby-nexthop:

### DIFF
--- a/sonic-xcvrd/xcvrd/dom/dom_mgr.py
+++ b/sonic-xcvrd/xcvrd/dom/dom_mgr.py
@@ -22,6 +22,7 @@ try:
     from xcvrd.xcvrd_utilities import sfp_status_helper
     from xcvrd.xcvrd_utilities.xcvr_table_helper import *
     from xcvrd.xcvrd_utilities import port_event_helper
+    from xcvrd.xcvrd_utilities import xcvrd_common
     from xcvrd.dom.utilities.dom_sensor.db_utils import DOMDBUtils
     from xcvrd.dom.utilities.vdm.utils import VDMUtils
     from xcvrd.dom.utilities.vdm.db_utils import VDMDBUtils
@@ -129,7 +130,7 @@ class DomInfoUpdateTask(threading.Thread):
             self.log_warning("Got invalid asic index for {} while checking cmis init status".format(logical_port_name))
             return False
 
-        cmis_state = xcvrd.get_cmis_state_from_state_db(logical_port_name, self.xcvr_table_helper.get_status_sw_tbl(asic_index))
+        cmis_state = xcvrd_common.get_cmis_state_from_state_db(logical_port_name, self.xcvr_table_helper.get_status_sw_tbl(asic_index))
         if cmis_state not in xcvrd.CMIS_TERMINAL_STATES:
             return True
         else:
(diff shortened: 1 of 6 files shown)


sonic-xcvrd/xcvrd/xcvrd_utilities/xcvrd_common.py
prgeor, line 1: @bobby-nexthop nit. I was wondering if we could rename
`xcvrd_utilities` ->`util`, since the dir is already under `xcvrd/`. Also, `xcvrd_common` -> `common`

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvrd_common.py
```diff
@@ -0,0 +1,246 @@
...
+    for tb_line in msg:
+        for tb_line_split in tb_line.splitlines():
+            helper_logger.log_error(tb_line_split)
+
+def update_port_transceiver_status_table_sw(logical_port_name, status_sw_tbl, status, error_descriptions='N/A'):
+    """Update port SFP status table for SW fields on receiving SFP change event"""
+    fvs = swsscommon.FieldValuePairs([('status', status), ('error', error_descriptions)])
+    status_sw_tbl.set(logical_port_name, fvs)
+
+def _wrapper_get_presence(physical_port):
```
prgeor, line 47: @bobby-nexthop there are many _wrapper in xcvrd.py. would it possible to clean all of them and just use actual function name. like

`_wrapper_get_presence()` -> `get_presence()`. Around the year 2021 when different platforms were transitioning to platform API 2.0, we provided this interim solution to use a wrapper function to use either platform API or the sfputil API, but by now all platforms should be on platform APIs.

Use platform chassis API and no need to fallback on sfputil.

prgeor (commented):
@bobby-nexthop can you also move the `CmisManager` task to new file, similar to how dom manager is now moved to new file.

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvrd_common.py
bobby-nexthop: I've changed this file to common.py - to keep the diffs small I'll rename xcvrd_utilities -> util in another PR

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1133,7 +935,7 @@ def task_worker(self):
                 if lport not in self.port_dict:
                     continue
 
-                state = get_cmis_state_from_state_db(lport, self.xcvr_table_helper.get_status_sw_tbl(self.get_asic_id(lport)))
+                state = common.get_cmis_state_from_state_db(lport, self.xcvr_table_helper.get_status_sw_tbl(self.get_asic_id(lport)))
```
longhuan-cisco, line 938: Just a minor suggestion (though might not be directly related to this PR's topic of `refactor common xcvrd functions`):
would it better to encapsulate this long function call into a new method of CmisManagerTask, e.g. 
```
class CmisManagerTask(threading.Thread):
...
    def get_cmis_state_from_transceiver_status_table(self, lport):
        return get_cmis_state_from_state_db(lport, self.xcvr_table_helper.get_status_tbl(self.get_asic_id(lport)))
```
so that it would be simplified as: `self.get_cmis_state_from_transceiver_status_table(lport)`

Apart from xcvrd.py, I saw there are quite a few such long function calling `get_cmis_state_from_state_db(lport, self.xcvr_table_helper.get_status_sw_tbl(self.get_asic_id(lport)))` in test_xcvrd.py.

sonic-xcvrd/xcvrd/dom/dom_mgr.py
```diff
@@ -142,19 +143,19 @@ def is_port_dom_monitoring_disabled(self, logical_port_name):
...
+            if not common._wrapper_get_presence(physical_port):
                 continue
 
             try:
                 if firmware_info_cache is not None and physical_port in firmware_info_cache:
                     # If cache is enabled and firmware information is in cache, just read from cache, no need read from EEPROM
                     transceiver_firmware_info_dict = firmware_info_cache[physical_port]
                 else:
-                    transceiver_firmware_info_dict = xcvrd._wrapper_get_transceiver_firmware_info(physical_port)
+                    transceiver_firmware_info_dict = common._wrapper_get_transceiver_firmware_info(physical_port)
```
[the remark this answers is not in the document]
bobby-nexthop: Created https://github.com/sonic-net/sonic-platform-daemons/issues/682 to track this in another review

mihirpat1 (approved):

PR #661, opened 2025-08-12
Added regex parsing to optics si settings parser

prgeor:
#### Description
Added regex parsing to optics si settings parser
Refactored the code to make more readable 

#### Motivation and Context
Some platform may need per port or range of ports that needs different SI settings on the module. Adding a regex parsing in vendor part number helps to minimize the change in optics_si_settings.json

Validated this change on platform already using GLOBAL_MEDIA_SETTINGS
Ensured the parser successfully parsed the SI settings for the required platform and optics Vendor name and PN.


prgeor:
@AnoopKamath, @mihirpat1, @bobby-nexthop can you review?

mihirpat1 (approved):

PR #672, opened 2025-09-05
[xcvrd] Improve logging in case of unable to find appropriate match for optic SI settings

mihirpat1:
#### Description

This PR improves the logging in the optics SI parser to provide better diagnostic information when unable to find an appropriate match for optic SI settings for a physical port.

#### Motivation and Context

The previous log message was an error which is misleading since there can be genuine cases wherein optics SI settings are not present for a module.
Hence, changing the log level to INFO and adding more debug data.

#### How Has This Been Tested?

Testing is in progress currently

#### Additional Information (Optional)
MSFT ADO - 34731836

PR #673, opened 2025-09-05
[202505] [xcvrd] Improve logging in case of unable to find appropriate match for optic SI settings (#672)

mihirpat1:
202505 cherry-pick for https://github.com/sonic-net/sonic-platform-daemons/pull/672

#### Description

This PR improves the logging in the optics SI parser to provide better diagnostic information when unable to find an appropriate match for optic SI settings for a physical port.

#### Motivation and Context

The previous log message was an error which is misleading since there can be genuine cases wherein optics SI settings are not present for a module.
Hence, changing the log level to INFO and adding more debug data.

Verified

MSFT ADO - 34731836


yejianquan (approved):
LGTM

PR #640, opened 2025-07-09
Fix physical entity tables in statedb to account for modified psu key naming

gregoryboudreau:
#### Description
Modify every instance of PSU parent naming to attempt to call the get_name API from the PSU/parent object, falling back to the prior implemented default behavior in the case that the API is not yet implemented by the vendor.

#### Motivation and Context
Issue raised: https://github.com/sonic-net/sonic-platform-daemons/issues/638, the key changes that took place previously on PSUD missing updating their associated phy entity tables in thermalctld. This PR makes the changes to account for this change.

#### How Has This Been Tested?
Ensure that phy entity tables in the statedb correctly reflect the name of their associated parent (in this case the PSU's get_name API response),  and run the SNMP docker to ensure that the reported failure no longer triggers.

Failure behavior:
```
2025 Jul  9 07:13:00.815494 MtFuji INFO snmp#snmp-subagent [ax_interface] INFO: OID registration complete. Waiting to receive PDUs...
2025 Jul  9 07:13:01.293223 MtFuji INFO snmp#supervisord 2025-07-09 07:13:01,292 INFO exited: dependent-startup (exit status 0; expected)
2025 Jul  9 07:13:05.800822 MtFuji ERR snmp#snmp-subagent [sonic_ax_impl] ERROR: fan_name = PSU1.fan0 get fan parent failed
2025 Jul  9 07:13:05.801176 MtFuji ERR snmp#snmp-subagent [sonic_ax_impl] ERROR: fan_name = PSU0.fan0 get fan parent failed
2025 Jul  9 07:13:11.831223 MtFuji ERR snmp#snmp-subagent [sonic_ax_impl] ERROR: fan_name = PSU1.fan0 get fan parent failed
2025 Jul  9 07:13:11.831596 MtFuji ERR snmp#snmp-subagent [sonic_ax_impl] ERROR: fan_name = PSU0.fan0 get fan parent failed
2025 Jul  9 07:13:17.860779 MtFuji ERR snmp#snmp-subagent [sonic_ax_impl] ERROR: fan_name = PSU1.fan0 get fan parent failed
2025 Jul  9 07:13:17.861094 MtFuji ERR snmp#snmp-subagent [sonic_ax_impl] ERROR: fan_name = PSU0.fan0 get fan parent failed
```
```
root@MtFuji:/home/cisco# redis-dump -d 6 -y -k "PHYSICAL_ENTITY_INFO*PSU*"
{
  "PHYSICAL_ENTITY_INFO|PSU0": {
[...]

PR #677, opened 2025-09-08
Fix physical entity tables in statedb to account for modified psu key naming - 202505 DC

gregoryboudreau:
#### Description

Creating DC of https://github.com/sonic-net/sonic-platform-daemons/pull/640 for 202505
#### Motivation and Context

Bringing fix for https://github.com/sonic-net/sonic-platform-daemons/pull/640 into 202505


ZhaohuiS (approved):
@gregoryboudreau  Thanks for the quick response!

PR #687, opened 2025-09-24
Add support for LPO in media_settings_parser

longhuan-cisco:
dependency: https://github.com/sonic-net/sonic-platform-common/pull/599

#### Description

LPO (Linear Pluggable Optics) is a new type of optical module without DSP, in which case the role of DSP (retiming/etc) is offloaded to host/NPU side, which generally might require different NPU SI settings to achieve good signal quality compared to today's DSP-based optics.

1. Today's `medium_lane_speed_key` in media_settings.json only supports two media_types OPTICAL and COPPER, additionally LPO as (extended) media_type is needed.
2. To achieve backward compatibility on platforms which don't want to distinguish between DSP-based optics and LPO, have optional flag  ENABLE_EXTENDED_MEDIA_TYPE_MATCHING in media_settings.json: only distinguish between DSP-based optics and LPO in the case of this flag being set.
3. This infra is extensible: in the future, new media_types (in addition to OPTICAL, COPPER, OPTICAL_LPO) can be easily added and enabled optionally in ENABLE_EXTENDED_MEDIA_TYPE_MATCHING.

Example media_settings.json:
```
{
    "ENABLE_EXTENDED_MEDIA_TYPE_MATCHING": [
        “OPTICAL_LPO”
    ],
    "GLOBAL_MEDIA_SETTINGS": {
        "0-15,48-63": {
            “OPTICAL100": {
...
            “OPTICAL_LPO100": {
...
            "COPPER100": {
...
```

### DIFF
--- a/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
+++ b/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
@@ -29,7 +29,7 @@ LANE_SPEED_DEFAULT_KEY = LANE_SPEED_KEY_PREFIX + DEFAULT_KEY
 MEDIA_COPPER = 'COPPER'
 MEDIA_OPTICAL = 'OPTICAL'
 # Extended media types:
-EXTENDED_MEDIA_OPTICAL_LPO = MEDIA_OPTICAL + '_LPO'
+MEDIA_LPO = 'LPO'
 
 # This is a top-level key in media_settings.json that can be added to enable
 # matching extended media types in addition to COPPER and OPTICAL. Its value is
@@ -120,8 +120,8 @@ def get_media_type(physical_port):
         return MEDIA_COPPER
 
     # For optical transceivers:
-    if is_extended_media_type_enabled(EXTENDED_MEDIA_OPTICAL_LPO) and get_is_lpo(physical_port):
-        return EXTENDED_MEDIA_OPTICAL_LPO
+    if is_extended_media_type_enabled(MEDIA_LPO) and get_is_lpo(physical_port):
+        return MEDIA_LPO
     return MEDIA_OPTICAL
 
 
--- a/sonic-xcvrd/tests/test_xcvrd.py
+++ b/sonic-xcvrd/tests/test_xcvrd.py
@@ -4686,14 +4686,14 @@ class TestXcvrdScript(object):
 
         # For the case of extended media type matching enabled for LPO:
         media_settings_g_dict = {
-            ENABLE_EXTENDED_MEDIA_TYPE_MATCHING: [EXTENDED_MEDIA_OPTICAL_LPO]
+            ENABLE_EXTENDED_MEDIA_TYPE_MATCHING: [MEDIA_LPO]
         }
 
         # Test LPO module
         with patch('xcvrd.xcvrd_utilities.media_settings_parser.g_dict', media_settings_g_dict):
             with patch('xcvrd.xcvrd_utilities.media_settings_parser.get_is_copper', return_value=False):
                 with patch('xcvrd.xcvrd_utilities.media_settings_parser.get_is_lpo', return_value=True):
-                    assert media_settings_parser.get_media_type(physical_port) == EXTENDED_MEDIA_OPTICAL_LPO
+                    assert media_settings_parser.get_media_type(physical_port) == MEDIA_LPO
 
         # Test non-LPO optical module
         with patch('xcvrd.xcvrd_utilities.media_settings_parser.g_dict', media_settings_g_dict):


prgeor (commented):
@bobby-nexthop can you review?

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
```diff
@@ -24,6 +24,18 @@
 DEFAULT_KEY = 'Default'
 # This is useful if default value is desired when no match is found for lane speed key
 LANE_SPEED_DEFAULT_KEY = LANE_SPEED_KEY_PREFIX + DEFAULT_KEY
+
+# Standard media types:
+MEDIA_COPPER = 'COPPER'
+MEDIA_OPTICAL = 'OPTICAL'
+# Extended media types:
+EXTENDED_MEDIA_OPTICAL_LPO = MEDIA_OPTICAL + '_LPO'
```
prgeor, line 32: @longhuan-cisco Why can't it be just `MEDIA_LPO = 'LPO' `?

longhuan-cisco:
> @longhuan-cisco I don' t think `"ENABLE_EXTENDED_MEDIA_TYPE_MATCHING"` is necessary

The reason of adding this `ENABLE_EXTENDED_MEDIA_TYPE_MATCHING ` is to provide backward compatibility on platforms which don't (or hasn't) distinguish between DSP-based optics and LPO. Without this, the behavior will change when matching today's media_settings.json which contains `medium_lane_speed_key` but doesn't have LPO specific settings.

e.g, Arista today already has a few media_settings.json files using `medium_lane_speed_key`
https://github.com/sonic-net/sonic-buildimage/blob/master/device/arista/x86_64-arista_7280dr3a_36/media_settings.json
Let's take this file as example, and assuming we have LPO DR optics inserted on port 1 with lane_speed 50G:
- The behavior of today's code: this port's `medium_lane_speed_key` will be `OPTICAL50`, which ends up matching [these SI settings](https://github.com/sonic-net/sonic-buildimage/blob/9a350d4e53e8b04211e1a10ed030d4407e2c8f8e/device/arista/x86_64-arista_7280dr3a_36/media_settings.json#L128-L189) in the [media_settings.json](https://github.com/sonic-net/sonic-buildimage/blob/master/device/arista/x86_64-arista_7280dr3a_36/media_settings.json)
- The behavior after adding LPO media_type but without the protection of `is_extended_media_type_enabled(LPO)`: this port's `medium_lane_speed_key ` will be `LPO50`, which ends up not matching any SI settings in [media_settings.json](https://github.com/sonic-net/sonic-buildimage/blob/master/device/arista/x86_64-arista_7280dr3a_36/media_settings.json)


https://github.com/longhuan-cisco/sonic-platform-daemons/blob/6f19de4f8434c2180756f34d2e40268649ccb486/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py#L108-L125
```
def get_media_type(physical_port):
    """
    Get the media type of the transceiver


    Args:
        physical_port: physical port number


    Returns:
        'COPPER', 'OPTICAL' or extended media types included in
        ENABLE_EXTENDED_MEDIA_TYPE_MATCHING of media_settings.json
    """
    if get_is_copper(physical_port):
        return MEDIA_COPPER


    # For optical transceivers:
    if is_extended_media_type_enabled(EXTENDED_MEDIA_OPTICAL_LPO) and get_is_lpo(physical_port):
        return EXTENDED_MEDIA_OPTICAL_LPO
    return MEDIA_OPTICAL
```

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
longhuan-cisco: Sure will update

longhuan-cisco: updated to `MEDIA_LPO = 'LPO'`

bobby-nexthop:
One way I can think of to not remove the need ENABLE_EXTENDED_MEDIA_TYPE_MATCHING and keep backwards compatibility is to use a fallback instead:
```python
if get_is_lpo(physical_port):
    return MEDIA_LPO
return MEDIA_OPTICAL
```

Then add some form of fallback logic when looking up settings:

```python
def get_media_settings_value(port_config, medium_lane_speed_key):
    # Try exact match first
    if medium_lane_speed_key in port_config:
        return port_config[medium_lane_speed_key]
    
    # If LPO key not found, automatically fallback to OPTICAL and log
    
```

@prgeor @longhuan-cisco any thoughts ?

prgeor:
> One way I can think of to not remove the need ENABLE_EXTENDED_MEDIA_TYPE_MATCHING and keep backwards compatibility is to use a fallback instead:
> 
> ```python
> if get_is_lpo(physical_port):
>     return MEDIA_LPO
> return MEDIA_OPTICAL
> ```
> 
> Then add some form of fallback logic when looking up settings:
> 
> ```python
> def get_media_settings_value(port_config, medium_lane_speed_key):
>     # Try exact match first
>     if medium_lane_speed_key in port_config:
>         return port_config[medium_lane_speed_key]
>     
>     # If LPO key not found, automatically fallback to OPTICAL and log
>     
> ```
> 
> @prgeor @longhuan-cisco any thoughts ?

@longhuan-cisco can you make this change?

longhuan-cisco:
Thanks @bobby-nexthop for the suggestion!

Will make change to fallback to OPTICAL settings in the case of LPO medium_lane_speed_key not found in media_settings.json

Issue #688, opened 2025-09-25
xcvrd does not use the gearbox line lanes for CMIS application lookup


aditya-nexthop:
On systems that use gearbox PHYs where the number of lanes on the switch ASIC is different from the number of lanes on the gearbox PHY, xcvrd tries to find an application that uses the number of lanes from the switch ASIC PHY. However, due to the gearbox PHY, the number of host lanes for the transceiver should be the number of lanes from the gearbox PHY.

This causes either a matching CMIS application to not be found or a wrong application to be selected.

PR #689, opened 2025-09-25
Fix up xcvrd to use line lanes from gearbox table if present

aditya-nexthop:
#### Description

- Implement new method get_gearbox_line_lanes_dict in xcvr_table_helper.py to return a dict of number of gearbox line lanes indexed by logical port.
- Building this dict and caching it in CmisTaskManager before a new iteration over port_dict
- Implement new method get_host_lane_count to look up the gearbox line lanes in this cached dict for every logical port. If found, use this as the host_lane_count, else use the host_lane_count from port_config
- Use this new method when setting host_lane_count for every logical port iteration

#### Motivation and Context
On switches with gearbox PHYs, for a given interface, the number of lanes between the switch ASIC and the gearbox PHY may be different from the number of lanes between the gearbox PHY and the transceiver. `xcvrd` currently looks only at the switch ASIC lanes (port_config) as `host_lane_count` which can result in `xcvrd` either not finding a matching CMIS application or choosing a wrong application.

This change allows `xcvrd` to consider the gearbox line lanes as `host_lane_count` when present. This is accurate to the hardware.

With this change, `xcvrd` is successfully able to select the correct CMIS application on a platform with such a gearbox PHY chip. Additional unit tests have also been added in support of this change.

fixes https://github.com/sonic-net/sonic-platform-daemons/issues/688


sonic-xcvrd/xcvrd/xcvrd_utilities/xcvr_table_helper.py
```diff
@@ -238,3 +239,52 @@ def is_npu_si_settings_update_required(self, lport, port_mapping):
...
+        Note:
+            - Returns empty dict if no gearbox configuration is found
+            - Silently skips invalid or malformed entries
+            - Only processes keys that start with "interface:"
+        """
+        gearbox_line_lanes_dict = {}
+        try:
+            for asic_id in self.appl_db:
+                appl_db = self.appl_db[asic_id]
+                gearbox_table = swsscommon.Table(appl_db, "_GEARBOX_TABLE")
```
prgeor, line 269: @aditya-nexthop can you make this namespace aware for platforms that may have multiple ASICs?

prgeor (approved):

PR #694, opened 2025-10-17
thermalctld: fix test warnings

fraserg-arista:
#### Description
 * remove duplicated code for mocking swsscommon
 * add missing method to mocked swsscommon.Table
 * remove a deprecated import

#### Motivation and Context

Fixes exceptions which produce warnings (and harm test coverage) when running thermalctld tests.

Ran thermalctld tests, observed no more warnings.


sonic-thermalctld/tests/mock_swsscommon.py
```diff
@@ -2,33 +2,7 @@
     Mock implementation of swsscommon package for unit testing
 '''
 
-STATE_DB = ''
-CHASSIS_STATE_DB = ''
+from .mocked_libs.swsscommon import swsscommon
```
vvolam: Move these 3 lines of code directly to test_thermalctld.py as the file size is reduced now

PR #697, opened 2025-10-29
Skip setting lpmode of a xcvr if it's not supported

kewei-arista:
#### Description

This change updates the logic to set lpmode on xcvr insertion or port enabling to make sure the API is only called when lpmode is supported by a module.

#### Motivation and Context

We noted that the following error is emitted when a power class 1 module (e.g. 40GBASE-SR4) is used on the events like insertion or enabling the port:

```
 moby202 ERR pmon#xcvrd[34]: SFF-MAIN: Ethernet513: Failed to take module out of low power mode.
```

It turns out the issue is that power class 1 module doesn't support setting lpmode: https://github.com/sonic-net/sonic-platform-common/blob/master/sonic_platform_base/sonic_xcvr/api/public/sff8636.py#L416, but SFF manager is unconditionally setting it on xcvr insertion or port enabling: https://github.com/sonic-net/sonic-platform-daemons/blob/7c75b2840cc93809ee1accc9c159dbd160780963/sonic-xcvrd/xcvrd/sff_mgr.py#L478

This caused some noise in the logs as there's no actual issue with link.

Tested on our switch and confirm the error log is no longer existed with SR4.

- [x] 202505

### DIFF
--- a/sonic-xcvrd/tests/test_xcvrd.py
+++ b/sonic-xcvrd/tests/test_xcvrd.py
@@ -2241,9 +2241,10 @@ class TestXcvrdScript(object):
         assert mock_xcvr_api.get_power_class.call_count == 5
         assert mock_xcvr_api.set_high_power_class.call_count == 1
 
+    @patch('xcvrd.xcvrd.helper_logger')
     @patch('xcvrd.xcvrd.platform_chassis')
     @patch('xcvrd.sff_mgr.PortChangeObserver', MagicMock(handle_port_update_event=MagicMock()))
-    def test_SffManagerTask_task_worker(self, mock_chassis):
+    def test_SffManagerTask_task_worker(self, mock_chassis, mock_logger):
         mock_xcvr_api = MagicMock()
         mock_xcvr_api.tx_disable_channel = MagicMock(return_value=True)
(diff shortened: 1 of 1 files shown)


sdszhang:
@prgeor @mihirpat1 for viz.

sdszhang:
@kewei-arista is this a master branch only change? If it's needed for 202505, please share more detail on which case it fails.

kewei-arista:
> @kewei-arista is this a master branch only change? If it's needed for 202505, please share more detail on which case it fails.

@sdszhang we need to backport this PR to 202505 as we recently enabled SFF manager for Moby on 202505: https://github.com/sonic-net/sonic-buildimage/pull/24262. I updated the description to include that.

kewei-arista:
> @kewei-arista can you check the 202505 cherry-pick conflict?

@sdszhang https://github.com/sonic-net/sonic-platform-daemons/pull/698/files

kewei-arista:
> LGTM thanks @kewei-arista
> 
> Can we add a unit testcase to cover the `if not set_lp_success` case ?

Thanks for looking @bobby-nexthop. I just added the test case in the last commit.

Issue #699, opened 2025-10-30
Test failures in test_thermalctld.py: ThermalControlDaemon.__init__() missing 3 required positional arguments


louis-nexthop:
`test_get_chassis_exception` and `test_get_chassis_success` are failing with the following error:

```
TypeError: ThermalControlDaemon.__init__() missing 3 required positional arguments: 'thermal_monitor_initial_interval', 'thermal_monitor_update_interval', and 'thermal_monitor_update_elapsed_threshold' 
```

This is because #635 introduced those 3 arguments, but it got merged after #652 added the tests that just call `daemon_thermalctld = thermalctld.ThermalControlDaemon()` without passing any arguments

louis-nexthop:
I opened https://github.com/sonic-net/sonic-platform-daemons/pull/700 for fixing this issue

PR #700, opened 2025-10-30
Fix test failures in test_thermalctld.py by passing arguments to ThermalControlDaemon()

louis-nexthop:
#### Description

Fix test failures in test_thermalctld.py, that was introduced in #652. `ThermalControlDaemon.__init__()` now requires 3 arguments, so we have to pass them.

#### Motivation and Context

Fixes https://github.com/sonic-net/sonic-platform-daemons/issues/699, which will allow Azure Pipelines to pass.

`test_get_chassis_exception` and `test_get_chassis_success` were failing with the following error:
```
TypeError: ThermalControlDaemon.__init__() missing 3 required positional arguments: 'thermal_monitor_initial_interval', 'thermal_monitor_update_interval', and 'thermal_monitor_update_elapsed_threshold' 
```

because they didn't pass thermal-monitor-related args to `ThermalControlDaemon()`, which were introduced in https://github.com/sonic-net/sonic-platform-daemons/pull/635.

Ran unit tests locally and passed.

### DIFF
--- a/sonic-thermalctld/tests/test_thermalctld.py
+++ b/sonic-thermalctld/tests/test_thermalctld.py
@@ -813,7 +813,7 @@ class TestThermalControlDaemon(object):
             
             # ThermalControlDaemon should raise SystemExit with CHASSIS_GET_ERROR code when chassis initialization fails
             with pytest.raises(SystemExit) as exc_info:
-                daemon_thermalctld = thermalctld.ThermalControlDaemon()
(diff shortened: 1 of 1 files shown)


vvolam (approved):
why the https://github.com/sonic-net/sonic-platform-daemons/pull/635 got merged with these test failures, was the code not compiled before raising the PR?

PR #708, opened 2025-11-14
[202505] Re-arm timer for CMIS Datapath pre-init check (#707)

mihirpat1:
Cherry-pick for https://github.com/sonic-net/sonic-platform-daemons/pull/707/

**Description**
Re-arm the timer before entering CMIS Datapath pre-init check

**Motivation and Context**
There could be a module which may take more time to transition its Datapath to DPinitialized or Deactivated state when the TX is turned OFF. The retry code today will not set the timer again after the DP state is transitioning through DPInit and sometime can be stuck there (bad module).

By re-arming the timer before entering the CMIS_STATE_DP_PRE_INIT_CHECK ensures the state machine will timeout appropriately

**How Has This Been Tested?**
xcvrd remains stable post restart
Manually modified the DP state to be not DataPathDeactivated or DataPathInitialized and ensured that the CMIS SM eventually transitions to FAILED state post 4 retries post issuing interface shutdown followed by no shutdown.
Ensure link comes up post shut/no shut

Additional Information (Optional)
ADO - 35916818


StormLiangMS (approved):
LGTM

PR #727, opened 2025-12-08
Stop the `config_manager` child process on exception in `chassisd`

arista-nwolfe:
Context is in: https://github.com/sonic-net/sonic-buildimage/issues/24775

TLDR:
`chassisd` process doesn't die when a crash occurs because the `config_manager` child process is still running.
Adding `try/finally` to cleanup the `config_manager` properly when a crash occurs to let `chassisd` die and restart.

Confirmed that if I add an `assert False` after `config_manager` is created I see the `chassisd` process die and restart as expected (previously the process would live on).

Note: When reviewing the diff select `hide whitespaces` to make the diff easier to read.

Cast Request:
- 202405

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -1334,6 +1334,7 @@ class ChassisdDaemon(daemon_base.DaemonBase):
         super(ChassisdDaemon, self).__init__(log_identifier)
 
         self.stop = threading.Event()
+        self.loop_interval = CHASSIS_INFO_UPDATE_PERIOD_SECS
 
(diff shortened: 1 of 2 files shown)


YatishSVC (approved):
lgtm, Thanks

anamehra (approved):
LGTM

arlakshm (commented):
@nwolfe, is it possible to add UT for this?

mlok-nokia (approved):

arlakshm (approved):

vmittal-msft:
@arista-nwolfe please help raise pr for 202511 due to cherry-pick conflict

PR #757, opened 2026-02-20
Set DOM polling interval based on the time at the start of the loop instead of the end

### LINKED ISSUE #756: DomInfoUpdateTask polling period DOM_INFO_UPDATE_PERIOD_SECS improperly delays polling
`DOM_INFO_UPDATE_PERIOD_SECS` should be the time delay between polling information from a specified interface, say `Ethernet0`

So the expectation is:
```
Polled Ethernet0 @ t=0
Polled Ethernet4 @ t = ...
...
Polled Ethernet200 @ t = X

Polled Ethernet0 @ t=DOM_INFO_UPDATE_PERIOD_SECS
```

Instead what we see is:
```
Polled Ethernet0 @ t=0
Polled Ethernet4 @ t = ...
...
Polled Ethernet200 @ t = X

Polled Ethernet0 @ t=X+DOM_INFO_UPDATE_PERIOD_SECS
```

This means no information is polled from any transceiver for `DOM_INFO_UPDATE_PERIOD_SECS` after the polling loop completes, even though more time than `DOM_INFO_UPDATE_PERIOD_SECS` may have elapsed since Ethernet0 was polled.

This is due to the following code:
```
class DomInfoUpdateBase(threading.Thread):
...
    def task_worker(self):
...
        # Start loop to update dom info in DB periodically and handle port change events
        while not self.task_stopping_event.is_set():
            # Check if periodic db update is needed
            if next_periodic_db_update_time <= datetime.datetime.now():
                is_periodic_db_update_needed = True
...
            for physical_port, logical_ports in self.port_mapping.physical_to_logical.items():
...
            if is_periodic_db_update_needed:
                next_periodic_db_update_time = datetime.datetime.now() + \          <<===============
                                               datetime.timedelta(seconds=dom_info_update_periodic_secs)
                is_periodic_db_update_needed = False
```
In order to enforce proper timing delay, `now` should be set at the start of the loop and used at the highlighted place to set `next_periodic_db_update_time`

aditya-nexthop:
#### Description
During profiling we found that there was a significant period of time when transceiver information was not being updated.
Upon study, it was seen that for DOM_INFO_UPDATE_PERIOD_SECS after the last transceiver in the loop was polled, there were no updates.

#### Motivation and Context
fixes #756 
This improves DOM polling performance of transceivers in `xcvrd` and correctly ensures that DOM_INFO_UPDATE_PERIOD_SECS time elapses between updating a specific transceiver.

We did a CPU profiling without and with the change over a 10 min window just after `xcvrd` starts
Before:
image
After:
image

The CPU is more uniformly utilized after the change instead of staying idle when more than DOM_INFO_UPDATE_PERIOD_SECS has already passed since a specific transceiver was polled.

### DIFF
--- a/sonic-xcvrd/xcvrd/dom/dom_mgr.py
+++ b/sonic-xcvrd/xcvrd/dom/dom_mgr.py
@@ -272,7 +272,8 @@ class DomInfoUpdateTask(DomInfoUpdateBase):
         # Start loop to update dom info in DB periodically and handle port change events
         while not self.task_stopping_event.is_set():
             # Check if periodic db update is needed
-            if next_periodic_db_update_time <= datetime.datetime.now():
+            now = datetime.datetime.now()
+            if next_periodic_db_update_time <= now:
                 is_periodic_db_update_needed = True
 
             # Handle port change event from main thread
@@ -391,8 +392,7 @@ class DomInfoUpdateTask(DomInfoUpdateBase):
 
             # Set the periodic db update time after all the ports are processed
             if is_periodic_db_update_needed:
-                next_periodic_db_update_time = datetime.datetime.now() + \
-                                               datetime.timedelta(seconds=dom_info_update_periodic_secs)
+                next_periodic_db_update_time = now + datetime.timedelta(seconds=dom_info_update_periodic_secs)
                 is_periodic_db_update_needed = False
 
         self.log_notice("Stop DOM monitoring loop")


mihirpat1:
> > @aditya-nexthop @prgeor This change looks correct for fixing the interval anchoring. However, a broader question: if each port takes ~1s to update DOM data and there are 64 ports, the loop takes ~64s per iteration. With DOM_INFO_UPDATE_PERIOD_SECS potentially shorter than that, are we okay with the loop running back-to-back with no sleep/yield? Could there be a concern about starving other tasks or excessive CPU usage in that scenario?
> 
> @mihirpat1 in that case we may have to define `DOM_INFO_UPDATE_PERIOD_SECS ` per platform basis? As the polling period of default 60 seconds does NOT work universally?

@prgeor That's correct - defining it per platform can help in addressing this.

PR #763, opened 2026-02-26
[202405] Stop the config_manager child process on exception in chassisd

kenneth-arista:
This is a manual cherry-pick of https://github.com/sonic-net/sonic-platform-daemons/pull/727

-------

Context is in: sonic-net/sonic-buildimage#24775

TLDR:
`chassisd` process doesn't die when a crash occurs because the `config_manager` child process is still running. Adding `try/finally` to cleanup the `config_manager` properly when a crash occurs to let `chassisd` die and restart.

Confirmed that if I add an `assert False` after `config_manager` is created I see the `chassisd` process die and restart as expected (previously the process would live on).

Note: When reviewing the diff select `hide whitespaces` to make the diff easier to read.

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -622,6 +622,7 @@ class ChassisdDaemon(daemon_base.DaemonBase):
         super(ChassisdDaemon, self).__init__(log_identifier)
 
         self.stop = threading.Event()
+        self.loop_interval = CHASSIS_INFO_UPDATE_PERIOD_SECS
 
     # Override signal handler from DaemonBase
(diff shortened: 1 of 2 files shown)


abdosi (approved):

abdosi:
@bmridul @judyjoseph @arlakshm for viz.

judyjoseph:
@kenneth-arista  could you rebase your branch ? we have pushed a commit to fix this PR buid issue

kenneth-arista:
Merged equivalent cherry-pick to https://github.com/Azure/sonic-platform-daemons.msft/pull/62 instead

PR #770, opened 2026-03-10
Delete only 'PHYSICAL_ENTITY_INFO' entries that daemon has added

### LINKED ISSUE #25652: Bug: Race condition in PHYSICAL_ENTITY_INFO table cleanup during daemon shutdown
### Is it platform specific

generic

### Importance or Severity

Medium

### Description of the bug

Currently, `sensormond` and `thermalctld` delete **all entries** in the `PHYSICAL_ENTITY_INFO` table of `STATE_DB` when their class objects are garbage collected during shutdown. This creates race conditions where one daemon's data can be inadvertently deleted by another daemon.

Meanwhile, `psud` does **not** delete its entries from `PHYSICAL_ENTITY_INFO` at all during shutdown, which was masked with `thermalctld`/`sensormond` deleting all entries from the table.

### Steps to Reproduce

1. Modify `PSU_INFO_UPDATE_PERIOD_SECS` from 3 seconds to 100 seconds (this is to have larger window to illustrate that even with periodically updating `PSU_INFO`, race condition can still happen)
2. Restart `thermalctld` daemon
3. When `thermalctld` stops, it deletes all `PHYSICAL_ENTITY_INFO` entries (including PSU entries owned by `psud`)
4. During the 100-second window before `psud` re-populates its data, SNMP queries will fail to retrieve PSU entity information.

You would see an error with below signature:

```
ERROR: MIBUpdater.start() caught an unexpected exception during update_data()
  File "/usr/local/lib/python3.11/dist-packages/ax_interface/mib.py", line 51, in start
    self.reinit_data()
  File "/usr/local/lib/python3.11/dist-packages/sonic_ax_impl/mibs/ietf/rfc2737.py", line 347, in reinit_data
    raise Exception(exceptions)
Exception: [ValueError("invalid literal for int() with base 10: ''")]
ERROR: invalid literal for int() with base 10: ''
  File "/usr/local/lib/python3.11/dist-packages/sonic_ax_impl/mibs/ietf/rfc2737.py", line 332, in reinit_data
    updater.reinit_data()
  File "/usr/local/lib/python3.11/dist-packages/sonic_ax_impl/mibs/ietf/rfc2737.py", line 703, in reinit_data
    self._update_entity_cache(name)
  File "/usr/local/lib/python3.11/dist-packages/sonic_ax_impl/mibs/ietf/rfc2737.py", line 893, in _update_entity_cache
    psu_position = int(psu_position)
    ^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: ''
```

Expected Behavior:

1. We should not be seeing exceptions/errors in the syslog when we shutdown any daemons.
2. Shutting down one daemon shouldn't affect the data populated by other daemons.

### Relevant log output

```shell

```

```shell

```

### Attach files (if any)

_No response_

ymd-arista:
#### Description

In the current design, daemons like psud, sensormond, and thermalctld delete all entries in the `PHYSICAL_ENTITY_INFO` table of `STATE_DB` when their classes are being garbage collected.

This creates race conditions where one daemon might try to access its own data in the `PHYSICAL_ENTITY_INFO` table that another daemon has deleted.

For example, during system shutdown, if `thermalctld` daemon stops first, its class objects get garbage collected and all entries in the `PHYSICAL_ENTITY_INFO` table are deleted. If `psud` is still running and periodically updating its own data, SNMP queries between the deletion and re-population will fail to retrieve PSU entity information.

This patch fixes the issue by limiting what each daemon deletes to only the entries it
has added to the `PHYSICAL_ENTITY_INFO` table:
- `psud`: Deletes only PSU entries
- `thermalctld`: Deletes only fan, drawer, and temperature entries
- `sensormond`: Deletes only voltage and current entries

Additionally, defensive `.get()` checks are added before deletion to prevent KeyError exceptions if entries have already been removed by other processes.

This fixes https://github.com/sonic-net/sonic-buildimage/issues/25652

1. Ran the snmp sonic-mgmt test suite.
2. Ran a stress test which reboots the dut containing these changes over 100 times and didn't observe  https://github.com/sonic-net/sonic-buildimage/issues/25652.
3. Also added corresponding unit tests for this new design and ran those tests.

### DIFF
--- a/sonic-psud/scripts/psud
+++ b/sonic-psud/scripts/psud
@@ -503,6 +503,11 @@ class DaemonPsud(daemon_base.DaemonBase):
                 self.psu_tbl._del(get_psu_key(psu_index))
             except Exception as e:
                 self.log_error("Failed to delete PSU {} info from DB: {}".format(psu_index, e))
+            if self.phy_entity_tbl.get(get_psu_key(psu_index)) is not None:
+                try:
+                    self.phy_entity_tbl._del(get_psu_key(psu_index))
+                except:
+                    self.log_error("Failed to delete physical entity info from DB: {}".format(e))
 
         if self.chassis_tbl is not None:
             try:
--- a/sonic-psud/tests/test_DaemonPsud.py
+++ b/sonic-psud/tests/test_DaemonPsud.py
@@ -531,6 +531,40 @@ class TestDaemonPsud(object):
         daemon_psud._update_single_psu_entity_info(3, mock_psu1)
         daemon_psud.phy_entity_tbl.set.assert_called_with('PSU 3', expected_fvp)
 
+    def test_deinit(self):
+        psud.platform_chassis = MockChassis()
+        daemon_psud = psud.DaemonPsud(SYSLOG_IDENTIFIER)
+        daemon_psud.num_psus = 2
+        daemon_psud.psu_tbl = psud.swsscommon.Table("STATE_DB", "psu_table")
+        daemon_psud.psu_tbl._del = mock.MagicMock()
+        daemon_psud.phy_entity_tbl = psud.swsscommon.Table("STATE_DB", "phy_entity_table")
+        # Pre-populate physical entity table with PSU entries and extra entries
+        daemon_psud.phy_entity_tbl.mock_dict['PSU 1'] = {}
+        daemon_psud.phy_entity_tbl.mock_dict['PSU 2'] = {}
+        daemon_psud.phy_entity_tbl.mock_dict['FAN 1'] = {}  # Extra entry not owned by psud
+        daemon_psud.phy_entity_tbl._del = mock.MagicMock()
+        daemon_psud.chassis_tbl = psud.swsscommon.Table("STATE_DB", "chassis_table")
+        daemon_psud.chassis_tbl._del = mock.MagicMock()
+
+        daemon_psud.__del__()
+
+        # Verify PSU table entries are deleted for all PSUs
+        assert daemon_psud.psu_tbl._del.call_count == 2
+        expected_psu_calls = [mock.call('PSU 1'), mock.call('PSU 2')]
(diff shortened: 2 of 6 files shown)


mlok-nokia (approved):

rlhui:
@judyjoseph , can we review please? thx

bmridul:
Does chassisd need similar changes?

judyjoseph (approved):

judyjoseph:
> Does chassisd need similar changes?

@ymd-arista could you check.. I am merging this PR as it is open for a while. Will try to get in 202511

ymd-arista:
No, chassisd doesn't need similar changes. We only delete specific information populated by chassisd from the `PHYSICAL_ENTITY_INFO` table.

bmridul:
> No, chassisd doesn't need similar changes. We only delete specific information populated by chassisd from the `PHYSICAL_ENTITY_INFO` table.

Pls check, Looks like same issue might be here. Chassisd is deleting entries on module removal (OIR) not process kill (unless  I am missing something)

https://github.com/sonic-net/sonic-platform-daemons/blob/e5da370eab1a1d424b8d6e02aa7353ad3ff0e6da/sonic-chassisd/scripts/chassisd#L400

ymd-arista:
@bmridul

What this PR changes: When a daemon exits, it must delete only the entires it added to the `PHYSICAL_ENTITY_INFO` table in `STATE_DB`. Previously, the exiting daemons were deleting _all_ the entries in the `PHYSICAL_ENTITY_INFO` table, wiping out entries other daemons had written and breaking those daemons when they tried to read their own entries.

What this PR is not: It doesn't propose that daemons only delete entries at termination. That approach would leave `STATE_DB` full of stale entries during normal operation.

`Chassisd` writes to PHYSICAL_ENTITY_INFO as modules are added or removed. On termination, `Chassisd` already deletes only its own module entries, leaving the rest of the table intact.

PR #771, opened 2026-03-10
[smartswitch][chassisd] Fix missing DPU reboot cause when config-reload interrupts DPU reboot

### LINKED ISSUE #24275: Bug:[Smartswitch] DPU reboot cause is not updated if reboot is ongoing and config reload is executed
### Is it platform specific

generic

### Importance or Severity

Critical

### Description of the bug

If config reload is executed, the reboot cause of the DPU is not updated if the reboot of the DPU was ongoing parallely

### Steps to Reproduce

reboot -d DPU1 
config reload after 10 seconds
wait for containers to come up
Execute
show reboot cause all
show reboot history
 
```
root@sonic:/home/admin# show reboot-cause all

Device    Name                 Cause       Time                             User

--------  -------------------  ----------  -------------------------------  ------

NPU       2025_10_10_06_49_00  Power Loss  N/A                              N/A

DPU3      2025_10_10_07_43_05  Power Loss  Fri Oct 10 07:43:05 AM UTC 2025

DPU2      2025_10_10_07_43_05  Power Loss  Fri Oct 10 07:43:05 AM UTC 2025

DPU1      2025_10_10_07_36_12  Reboot      Fri Oct 10 07:36:12 AM UTC 2025

DPU0      2025_10_10_07_43_05  Power Loss  Fri Oct 10 07:43:05 AM UTC 2025

 

root@sonic:/home/admin# show reboot-cause history DPU1

Device    Name                 Cause         Time                             User    Comment

--------  -------------------  ------------  -------------------------------  ------  --------------------------------

DPU1      2025_10_10_07_36_12  Reboot        Fri Oct 10 07:36:12 AM UTC 2025          Reset from Main board

DPU1      2025_10_10_06_57_08  Reboot        Fri Oct 10 06:57:08 AM UTC 2025          Reset from Main board

DPU1      2025_10_10_06_50_08  Power Loss    Fri Oct 10 06:50:08 AM UTC 2025          power auxiliary outage or reload

DPU1      2025_10_10_06_36_29  Reboot        Fri Oct 10 06:36:29 AM UTC 2025          Reset from Main board

DPU1      2025_10_10_06_28_48  Non-Hardware  Fri Oct 10 06:28:48 AM UTC 2025          Reset due to Power off

DPU1      2025_10_10_06_28_28  Non-Hardware  Fri Oct 10 06:28:28 AM UTC 2025          Reset due to Power off
```

Actual behaviour:
Reboot cause of DPU is not updated
Expected behaviour:
Reboot cause has to be updated

### Relevant log output

```shell

```

```shell
202506 latest hash
```

### Attach files (if any)

_No response_

### LINKED ISSUE #26254: Bug:[Smartswitch]  DPU's reboot cause is being checked when the operational state changes to offline
### Is it platform specific

generic

### Importance or Severity

Critical

### Description of the bug

In chassisd implementation, when the operational state of the DPU has transitioned to offline, we are querying the reboot cause of the DPU, this is irrelevant at this point of time because the reboot cause of the DPU needs to be checked after the DPU has completed rebooting, we are trying to access the DPUs reboot cause capabilities while the DPU is powered off,

### Steps to Reproduce

power off DPU (`config chassis modules shutdown DPU0`) check logs:
```
2026 Mar  3 06:38:29.047071 sonic NOTICE pmon#chassisd: DPU2 operational status transitioning to offline
2026 Mar  3 06:38:29.076384 sonic NOTICE pmon#chassisd[39]: Reset reason for DPU2 is Non-Hardware
```

Actual behavior:
Checking for reboot cause when DPU is powered off
Expected behaviour:
Checking for reboot cause of DPU after it has powered on

### Relevant log output

```shell

```

```shell
202511 latest hash
```

### Attach files (if any)

_No response_

vvolam:
#### Description

Refactor DPU reboot-cause tracking in `SmartSwitchModuleUpdater` to correctly handle all operational-status transitions, including edge cases around `config reload` and chassisd restarts.

Key changes:
- Extract `_handle_module_status_change()` from `module_db_update()` for clearer transition handling
- Defer `get_reboot_cause` queries until the DPU is online — querying a powered-off DPU is unreliable (fixes #26254)
- On `Known→Offline`, persist the reboot execution time as fallback (if `module_base.py` has not already created it) and defer the cause query
- On `EMPTY→Offline` (chassisd restart while DPU is down), defer entirely until the DPU comes back online, then persist only if the cause actually changed
- Short-circuit `EMPTY→ONLINE` with no deferred state — just repopulate `CHASSIS_STATE_DB` from on-disk history (no reboot happened, chassisd just restarted). On first boot, persist the initial cause.
- Replace the arbitrary `MAX_DPU_REBOOT_DURATION` (800s) timeout with direct timestamp comparison: compare the stored cause timestamp against the reboot execution time from `prev_reboot_time.txt` to determine whether the cause was already persisted
- Always repopulate `CHASSIS_STATE_DB` from on-disk history after chassisd restart, even when no new cause is persisted
- Add `persist_dpu_reboot_time_if_needed()` fallback: on Online→Offline, if `module_base.py` has not already created `prev_reboot_time.txt` for this reboot cycle (detected by comparing file time vs stored cause time), chassisd writes it. This covers spontaneous DPU crashes/kernel panics where no command invocation triggers `module_base.py`.
- Change `log_info` to `log_notice` for reboot-cause persistence messages

#### Transition Matrix

| Transition | Condition | Action |
|---|---|---|
| ONLINE→OFFLINE | Known, non-offline → offline | Persist reboot time (fallback if module_base.py has not), defer cause query |
| EMPTY→OFFLINE | Unknown → offline | Defer cause query (no I/O) |
| EMPTY→ONLINE (no deferred) | First boot | Persist initial cause + repopulate DB |
| EMPTY→ONLINE (no deferred) | Not first boot | Repopulate DB only (no reboot happened) |
| OFFLINE→ONLINE (deferred=reboot) | stored_time >= reboot_time | Skip persist, repopulate DB |
| OFFLINE→ONLINE (deferred=reboot) | stored_time < reboot_time | Persist new cause + repopulate DB |
| OFFLINE→ONLINE (deferred=restart) | Cause changed | Persist new cause + repopulate DB |
| OFFLINE→ONLINE (deferred=restart) | Cause unchanged / first boot | Repopulate DB only |

#### `prev_reboot_time.txt` generation

Primary source: `module_base.py` writes `prev_reboot_time.txt` at command invocation time (via sonic-net/sonic-platform-common#680 and sonic-net/sonic-utilities#4571).

Fallback (this PR): On Online→Offline transition, `persist_dpu_reboot_time_if_needed()` checks whether `module_base.py` has already created the file for this reboot cycle by comparing the file timestamp against the stored cause timestamp. If the file is missing or stale, chassisd writes the current time. This ensures spontaneous DPU crashes (no CLI command) still get a reboot time marker.

#### Motivation and Context

Fixes sonic-net/sonic-buildimage#24275
Fixes https://github.com/sonic-net/sonic-buildimage/issues/26254

On SmartSwitch platforms, issuing `config reload` shortly after `reboot -d DPU0` causes chassisd to restart. After restart, STATE_DB is empty so chassisd sees `EMPTY→Offline` instead of `Online→Offline`, and the DPU reboot cause was never recorded. Additionally, querying `get_reboot_cause` on a powered-off DPU is unreliable and was causing spurious updates.

Preceding PRs: https://github.com/sonic-net/sonic-platform-common/pull/680
https://github.com/sonic-net/sonic-utilities/pull/4571

1. **Unit tests** — 63 tests in `sonic-chassisd/tests/test_chassisd.py`, all passing:
   - `test_smartswitch_moduleupdater_status_transitions` — full ONLINE→OFFLINE→ONLINE cycle
   - `test_admin_shutdown_does_not_query_reboot_cause` — verifies `get_reboot_cause` NOT called while DPU is off
   - `test_persist_dpu_reboot_time_if_needed_creates_file_when_missing` — fallback writes when file missing
   - `test_persist_dpu_reboot_time_if_needed_skips_when_file_newer` — skips when module_base.py already wrote
   - `test_persist_dpu_reboot_time_if_needed_writes_when_file_stale` — writes when file is stale from prior reboot
   - `test_online_transition_skips_reboot_update` — EMPTY→ONLINE repopulates DB without persisting
   - `test_empty_to_online_first_boot_persists` — first boot persists initial cause
   - `test_empty_to_online_no_spurious_persist` — regression test for config reload spurious update
   - `test_deferred_reboot_skips_already_persisted` — stored_time >= reboot_time → skip
   - `test_deferred_reboot_persists_new_reboot` — stored_time < reboot_time → persist
   - `test_empty_to_offline_persists_changed_reboot_cause` — deferred restart with changed cause
   - `test_empty_to_offline_skips_same_reboot_cause` — deferred restart with same cause, DB repopulated
   - `test_extract_cause_tuple` — helper extracts scalar from tuple/list/string

2. **SmartSwitch hardware validation** on `NVIDIA SS`:
   - Ran `reboot -d DPU0` followed by `config reload -y` after 10 seconds
   - With original code: reboot cause was stale (bug reproduced)
   - With patched code: new reboot cause entry correctly recorded

3. **Testbed verification script** — `sonic-chassisd/tests/testbed/test_dpu_reboot_cause_persistence.sh`: end-to-end script covering all 6 DPU reboot-cause transition scenarios on real hardware (normal reboot, config reload during reboot, admin shutdown, history preservation, same-cause deferred check, back-to-back rapid reboot).

The fix only applies to `SmartSwitchModuleUpdater` (SmartSwitch/DPU platforms). It does not affect line card or supervisor module updater logic.


sonic-chassisd/scripts/chassisd
```diff
@@ -810,6 +810,28 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
                     # publish reboot cause to db
                     self.update_dpu_reboot_cause_to_db(key)
 
+                # DPU discovered offline after chassisd restart (e.g. config reload
+                # during an ongoing DPU reboot).  prev_status is EMPTY because
+                # STATE_DB was flushed.  Check whether the reboot cause changed
+                # since the last persisted one and, if so, record it.
+                elif prev_status == ModuleBase.MODULE_STATUS_EMPTY and current_status == str(ModuleBase.MODULE_STATUS_OFFLINE):
+                    reboot_cause = try_get(self.chassis.get_module(module_index).get_reboot_cause)
```
gpunathilell: can we disable the check for the reboot cause when the current state is offline? the reboot cause needs to be checked only after dpu is completely online

gpunathilell:
@vvolam would this add a reboot cause for first boot of the system?

sonic-chassisd/scripts/chassisd
```diff
@@ -775,69 +776,113 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            return
 
-                if not key.startswith(ModuleBase.MODULE_TYPE_DPU):
-                    self.log_error("Incorrect module-name {}. Should start with {} ".format(key,
-                                                            ModuleBase.MODULE_TYPE_DPU))
-                    continue
+        # DPU discovered offline after chassisd restart (e.g. config reload
+        # during an ongoing DPU reboot).  prev_status is EMPTY because
+        # STATE_DB was flushed.  Defer reboot-cause check until DPU is online.
+        if not was_known and now_offline:
```
gpunathilell: wouldnt this also handle cases where DPU is currently offline and was offline previously as well?

vvolam:
> @vvolam would this add a reboot cause for first boot of the system?

@gpunathilell No, first boot is safe. In the `deferred == 'restart' `path, the code checks `stored_cause is not None` before persisting. On first boot there's no `previous-reboot-cause.json`, so `retrieve_dpu_reboot_info()` returns `(None, None)` — the guard fails and we skip persistence entirely.

This is covered by `test_empty_to_offline_skips_first_boot` in the test file.

sonic-chassisd/scripts/chassisd
vvolam: No — that case only triggers once. The `EMPTY → Offline `branch (`not was_known and now_offline`) fires only when `prev_status is EMPTY`, i.e. immediately after chassisd restart when STATE_DB is flushed. On the very next poll cycle, `prev_status `becomes `OFFLINE `(because we just wrote it to the module table), so `was_known is True and was_offline is True`. That means `now_offline `is also `True`, and none of the three if branches match — it's a no-op for steady-state Offline→Offline.

In short: the deferred flag is set exactly once on restart, consumed when the DPU comes online, and ignored on repeated `Offline→Offline `polls.

vvolam:
[test_ss_module_db_update.sh](https://github.com/user-attachments/files/26580138/test_ss_module_db_update.sh)
@gpunathilell I have executed this test script to verify all scenarios on the smartswitch

sonic-chassisd/scripts/chassisd
```diff
@@ -775,69 +776,113 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            # Deferred from Online→Offline: the DPU went down and is now back.
+            # Reboot time was already persisted.  Skip persisting if this is a
+            # back-to-back reboot with the same cause (is_reboot detection).
+            if deferred == 'reboot':
+                if current_cause and stored_cause and stored_time_str:
+                    try:
+                        stored_dt = datetime.strptime(stored_time_str, "%Y_%m_%d_%H_%M_%S").replace(tzinfo=timezone.utc)
+                        delta_sec = (datetime.now(timezone.utc) - stored_dt).total_seconds()
+                        if current_cause == stored_cause and delta_sec < MAX_DPU_REBOOT_DURATION:
+                            self.log_info(f"{key}: is_reboot=True — same reboot cause within {int(delta_sec)}s")
```
chartsai-nvidia: In the [test_ss_module_db_update.sh](https://github.com/user-attachments/files/26580138/test_ss_module_db_update.sh), the test 6 doesn't pass because the log level is `info` here. After I changed to `notice` the shell script test passed.

chartsai-nvidia:
> test_ss_module_db_update.sh

Manually patched `chassisd` and ran the tests, confirmed all tests passed.
[test_script_changes_diff.txt](https://github.com/user-attachments/files/27375387/test_script_changes_diff.txt)

chartsai-nvidia:
The fix introduces new bugs: the `reboot-cause` result is unstable.

After running `config reload`, DPU0 didn't reboot but the `reboot-cause` was updated.
Before:
```
Device    Name                 Cause         Time                             User
--------  -------------------  ------------  -------------------------------  ------
NPU       2026_05_01_20_37_23  reboot        Fri May  1 08:32:15 PM UTC 2026  admin
DPU3      2026_05_04_20_53_07  Non-Hardware  Mon May 04 08:53:07 PM UTC 2026
DPU2      2026_05_04_20_53_06  Non-Hardware  Mon May 04 08:53:06 PM UTC 2026
DPU1      2026_05_04_20_53_06  Non-Hardware  Mon May 04 08:53:06 PM UTC 2026
DPU0      2026_05_02_06_45_53  Non-Hardware  **Sat May 02 06:45:53 AM UTC 2026**
```
After:
```
Device    Name                 Cause         Time                             User
--------  -------------------  ------------  -------------------------------  ------
NPU       2026_05_01_20_37_23  reboot        Fri May  1 08:32:15 PM UTC 2026  admin
DPU3      2026_05_04_20_53_07  Non-Hardware  Mon May 04 08:53:07 PM UTC 2026
DPU2      2026_05_04_20_53_06  Non-Hardware  Mon May 04 08:53:06 PM UTC 2026
DPU1      2026_05_04_20_53_06  Non-Hardware  Mon May 04 08:53:06 PM UTC 2026
DPU0      2026_05_04_21_05_08  Non-Hardware  **Mon May 04 09:05:08 PM UTC 2026**
```

Then run`reboot -d DPU0` and `config reload` again, all other DPUs was updated.
```
Device    Name                 Cause         Time                             User
--------  -------------------  ------------  -------------------------------  ------
NPU       2026_05_01_20_37_23  reboot        Fri May  1 08:32:15 PM UTC 2026  admin
DPU3      2026_05_04_21_22_45  Non-Hardware  Mon May 04 09:22:45 PM UTC 2026
DPU2      2026_05_04_21_22_44  Non-Hardware  Mon May 04 09:22:44 PM UTC 2026
DPU1      2026_05_04_20_53_37  Non-Hardware  Mon May 04 08:53:37 PM UTC 2026
DPU0      2026_05_04_21_18_30  Non-Hardware  Mon May 04 09:18:30 PM UTC 2026
```

And then run `reboot -d DPU0`, the DPU0 didn't update.
```
admin@bobcat-1250:~$ show reboot-cause all
Device    Name                 Cause         Time                             User
--------  -------------------  ------------  -------------------------------  ------
NPU       2026_05_01_20_37_23  reboot        Fri May  1 08:32:15 PM UTC 2026  admin
DPU3      2026_05_04_21_22_45  Non-Hardware  Mon May 04 09:22:45 PM UTC 2026
DPU2      2026_05_04_21_22_44  Non-Hardware  Mon May 04 09:22:44 PM UTC 2026
DPU1      2026_05_04_20_53_37  Non-Hardware  Mon May 04 08:53:37 PM UTC 2026
DPU0      2026_05_04_21_18_30  Non-Hardware  Mon May 04 09:18:30 PM UTC 2026
```

liat-grozovik:
@ganglyu can you please confirm this PR is OK from Nvidia SmartSwitch POV or there is more feedback to share? if you are OK and it was tested as well please approve and i will merge it

gpunathilell:
@liat-grozovik this PR is not OK from nvidia smartswitch POV and needs update as per @chartsai-nvidia's comments, please do not merge it

dgsudharsan (changes requested):
Please fix the test failures as pointed out by Charles.

sonic-chassisd/scripts/chassisd
vvolam: Fixed in 2cff182 — changed both `log_info` calls to `log_notice` for the `is_reboot=True` messages.

vvolam:
@chartsai-nvidia Thanks for the detailed testing and the bug report!

Fixed in 2cff182. The root cause was the EMPTY→ONLINE path (no deferred flag) — when chassisd restarts and a DPU is already online, the code fell through to the generic `is_reboot` check. If the stored reboot timestamp was >800s old, `is_reboot=False` and the code spuriously persisted a new reboot cause for every DPU. This then poisoned the stored data (same cause + recent timestamp), causing subsequent real reboots to be falsely skipped.

The fix short-circuits the EMPTY→ONLINE (no deferred) path: just repopulate CHASSIS_STATE_DB from on-disk history files without persisting anything new. On first boot (`_is_first_boot=True`), the initial cause is still persisted.

Summary of changes:
1. **EMPTY→ONLINE short-circuit** — no spurious persist when chassisd restarts with DPUs already online
2. **`log_info` → `log_notice`** for both `is_reboot=True` messages (fixes test 6)
3. **New unit tests**: `test_empty_to_online_no_spurious_persist` and `test_empty_to_online_first_boot_persists`

gpunathilell:
@vvolam is it not better to confirm we log the time when reboot was executed, and make sure that there is a reboot cause logged AFTER that reboot is executed (assuming a config reload is performed)? instead of adding a new arbitrary timeout

vvolam:
@gpunathilell Great suggestion — implemented in commit a6fb958.

Replaced the `MAX_DPU_REBOOT_DURATION` (800s) arbitrary timeout with a direct comparison of the stored reboot-cause timestamp vs the reboot execution time from `prev_reboot_time.txt` (written by `persist_dpu_reboot_time()` on `Online→Offline`).

New logic in the `deferred=="reboot"` path:
- `stored_time >= reboot_time` → cause was already persisted for this reboot → skip
- `stored_time < reboot_time` → new reboot, cause not yet recorded → persist

This eliminates the arbitrary timeout and correctly handles back-to-back reboots, config reload after cause is persisted, and reboots of any duration.

Also fixed a bug where the `deferred=="restart"` path (EMPTY→Offline→Online) was not calling `update_dpu_reboot_cause_to_db()` when the cause was unchanged, leaving CHASSIS_STATE_DB empty after config reload. Now it always repopulates the DB from on-disk history.

dgsudharsan:
@vvolam Can you please make sure that [test_ss_module_db_update.sh](https://github.com/user-attachments/files/26580138/test_ss_module_db_update.sh ) is run and passes with your latest changes?

vvolam:
> @vvolam Can you please make sure that [test_ss_module_db_update.sh](https://github.com/user-attachments/files/26580138/test_ss_module_db_update.sh) is run and passes with your latest changes?

@dgsudharsan I have verified and all tests are passing. I have even fix some syslog tests in the script and committed as part of this PR itself. Thanks!

vvolam:
@rameshraghupathy any concerns with the PR?

rameshraghupathy:
@vvolam I understand the rationale for handling the `config reload` / chassisd restart case where STATE_DB is empty and the transition becomes EMPTY→Offline instead of Online→Offline.

However, I am concerned that this PR changes the existing reboot-cause model much more broadly than required for that issue.

Today, reboot-cause is intentionally collected and persisted during the Online→Offline transition. This is not accidental behavior. There is platform design and history behind it:

1. This matches the behavior expected/supported by other vendor platforms.
2. The requirement exists specifically to help debug dead or unreachable DPUs.
3. For this reason, `get_reboot_cause()` is intentionally allowed even when the DPU is offline.
4. The reboot-cause is mirrored/latched through the NPU path specifically so it can still be retrieved after the DPU goes down.

Deferring reboot-cause collection until the DPU comes back online violates that design assumption. If the DPU remains dead or unreachable, the reboot cause may never be persisted, which defeats the original purpose of the latched reboot-cause mechanism.

The current SmartSwitchModuleUpdater logic already has centralized transition handling using previous/current oper-status. If the missing case is specifically EMPTY→Offline after chassisd restart/config reload, that should be handled as a minimal extension to the existing transition logic, while preserving the existing Online→Offline reboot-cause persistence semantics.

IMO this should not be treated as a broad reboot-cause framework change. The fix should preserve the existing design contract and only add the missing EMPTY→Offline/config-reload handling.

vvolam:
> sting transition logic, while preserving the existing Online→Offline reboot-cause persistence semantics.
> 
> IMO this should not be treated as a broad reboot-cause framework ch

@rameshraghupathy Thanks for the detailed review. I'd like to address your concerns by referencing the SmartSwitch PMON HLD itself ([smartswitch-pmon.md](vscode-file://vscode-app/c:/Users/vvolam/AppData/Local/Programs/Microsoft%20VS%20Code/0958016b2a/resources/app/out/vs/code/electron-browser/workbench/workbench.html), section "Reboot workflow"):

```
"This function looks for DPU operational status change events and when the DPUs come out of "offline" state, issues get_reboot_cause API to the platform."

"The dpu operational state transition from 'offline' to 'online' guarantees the reboot of a DPU."
```

The HLD explicitly designs `get_reboot_cause `to be called on the offline→online transition, not online→offline. This PR aligns with that design:

ONLINE→OFFLINE: persist reboot execution time only (no get_reboot_cause call on a powered-off DPU)
OFFLINE→ONLINE: query get_reboot_cause and persist the cause
The HLD also states: "The platform code will extract the DPU reboot cause from the NPU hardware itself even when the DPU is not reachable." This describes the hardware capability (the latch is available whenever we query it), not the query timing. The designed query timing is offline→online.

The current code on master actually deviates from the HLD by calling get_reboot_cause during/after online→offline. This is what causes issue #26254 — querying a powered-off DPU produces unreliable results. This PR restores the behavior the HLD intended.

Regarding your concern about DPUs that remain dead: the hardware latch preserves the reboot cause indefinitely, so whenever the DPU eventually comes back online, the cause will still be available and will be persisted at that point. If the DPU never comes back, that's an admin-shutdown or hardware failure scenario — not a reboot-cause scenario.

chartsai-nvidia:
Hi @vvolam,
I took a look at `test_empty_to_offline_skips_same_reboot_cause` — if that test passes, then it means the new reboot is **not** persisted, and `previous-reboot-cause.json` keeps the **previous** reboot's timestamp. That is exactly the user-visible symptom reported in sonic-net/sonic-buildimage#24275:

> DPU1     2025_10_10_07_36_12   Reboot     Fri Oct 10 07:36:12 AM UTC 2025

shown after a `reboot -d DPU1` at 07:40:34. So I don't think the PR closes #24275 for the specific same-cause reproducer in that issue.
### Where the skip happens
In `_handle_module_status_change`, the deferred-`'restart'` branch:
```python
if deferred == 'restart':
    if stored_cause is not None and current_cause and current_cause != stored_cause:
        self.log_notice(f"{key}: Reboot cause changed while chassisd was down ...")
        self.persist_dpu_reboot_time(key)
        self.persist_dpu_reboot_cause(reboot_cause, key)
    self.update_dpu_reboot_cause_to_db(key)
    return
```
The guard is `current_cause != stored_cause`. Two consecutive `reboot -d DPU1` events both produce the same cause type (e.g. `Reboot / Reset from Main board`, or `Non-Hardware` on my HW), so this is False and we skip both `persist_dpu_reboot_time` and `persist_dpu_reboot_cause`. The new reboot is dropped.

Here's the script I test for this scenario:
[test.sh](https://github.com/user-attachments/files/28133984/test.sh)

Could you walk me through how the deferred='restart' branch is supposed to handle the same-cause case? Maybe I'm misreading the intended flow.
Thanks

vvolam:
@chartsai-nvidia Great catch — you're exactly right. The `deferred='restart'` guard `current_cause != stored_cause` silently drops back-to-back reboots that produce the same cause type.

Fixed in 67de161. The `deferred='restart'` path now uses the same timestamp-based logic as `deferred='reboot'`:

```python
reboot_time = self.retrieve_dpu_reboot_time(key)
if reboot_time and stored_time_str and stored_time_str < reboot_time:
    # New reboot detected — persist regardless of cause string
    should_persist = True
elif stored_cause is not None and current_cause and current_cause != stored_cause:
    # Cause changed — persist
    should_persist = True
```

If `prev_reboot_time.txt` (written during the original Online→Offline transition before `config reload`) records a time newer than the stored cause's timestamp in `previous-reboot-cause.json`, we know a new reboot happened and persist unconditionally — even if the cause string is identical.

For your test scenario (`reboot -d DPU0` → `config reload` → DPU comes back):
1. Online→Offline writes `prev_reboot_time.txt` = reboot execution time
2. `config reload` → chassisd restarts → EMPTY→Offline → `deferred='restart'`
3. DPU online → `stored_time_str` (from prior reboot) < `reboot_time` → **persist**

Could you re-run your [test.sh](https://github.com/user-attachments/files/28133984/test.sh) against this commit?

vvolam:
@chartsai-nvidia I have fixed test scripts accordingly and now all tests are passing including the test.sh script you shared.

gpunathilell:
@vvolam 
 can you modify the `reboot` and `restart` variable pairs to `dpu_reboot` / `chassisd_restart` to avoid ambiguity (we can also use an enum here since only two options are present)?
And please remove `was_known` variable, we are checking for not was_known in three places, instead please have `prev_status_empty` or something similar - this is more clear

gpunathilell:
I would also suggest to create functions for top-level transitions, so that it will match with the Transition matrix provided above, so that it becomes more clear:
```python
def _is_known_to_offline(self, prev_status, current_status):
    """Online (or any non-EMPTY non-OFFLINE) → OFFLINE — chassisd witnessed it."""
    return (prev_status != ModuleBase.MODULE_STATUS_EMPTY
            and prev_status != str(ModuleBase.MODULE_STATUS_OFFLINE)
            and current_status == str(ModuleBase.MODULE_STATUS_OFFLINE))

def _is_empty_to_offline(self, prev_status, current_status):
    """EMPTY → OFFLINE — chassisd just (re)started, found DPU already offline."""
    return (prev_status == ModuleBase.MODULE_STATUS_EMPTY
            and current_status == str(ModuleBase.MODULE_STATUS_OFFLINE))

def _is_back_to_online(self, prev_status, current_status):
    """EMPTY/OFFLINE → ONLINE — DPU usable again, safe to call get_reboot_cause()."""
    return (prev_status in (ModuleBase.MODULE_STATUS_EMPTY,
                            str(ModuleBase.MODULE_STATUS_OFFLINE))
            and current_status != str(ModuleBase.MODULE_STATUS_OFFLINE))
```
This will also help in extending code assuming there are some unclear scenarios at this point

vvolam:
@gpunathilell Thanks for the suggestions — all implemented in f23e6a4:

1. Replaced `'reboot'` / `'restart'` with named constants `DEFERRED_DPU_REBOOT` / `DEFERRED_CHASSISD_RESTART`
2. Replaced `was_known` with positive-sense `prev_status_empty` — eliminates the triple double-negation
3. Extracted transition predicates into `_is_known_to_offline()`, `_is_empty_to_offline()`, `_is_back_to_online()` matching the Transition Matrix

All 60 tests pass. Let me know if this looks good.

gpunathilell:
@vvolam  there is still a race condition in this implementation - the reboot -d DPUX is executed but the chassisd is restarted before the reboot_time file is  created right i.e. the chassisd has to detect that the dpu is being powered off and only then this would work right? can we have the implementation for this in the module_base implementation (i.e. for `admin_state` down and the `reboot` function calls?) this way chassisd is not waiting until dpu is powered off to create the file. This makes it more easier to control, and we can have the reboot cause time aligned with when the user executed the command, not sure if we want to cover as bug fix for this PR. We can discuss offline if requried

vvolam:
@gpunathilell Good catch — yes, there's a narrow race if `config reload` happens within the ~10s poll interval before chassisd detects Online→Offline and writes `prev_reboot_time.txt`. In that case the timestamp-based comparison can't fire and same-cause reboots would be missed.

Agree that writing the timestamp at the platform layer (`module_base.reboot()` / `set_admin_state`) is the right long-term fix — it decouples from chassisd's poll timing and aligns the timestamp with the actual user action.

Since this is a platform-API-level change affecting vendor implementations, I'd prefer to track it as a follow-up rather than expanding this PR's scope. Will file a separate issue. The current PR already covers the common case (chassisd detects the transition before config reload), and the remaining gap is the narrow same-cause + config-reload-within-10s edge case.

gpunathilell:
@vvolam its not narrow, its the complete offline cycle, assuming reboot -d DPUX is executed, chassisd does not detect dpu offline until dpu is shutdown, so technically the window is time taken for DPU to actually go offline + ~10 seconds. But I am okay to have a seperate PR, we can use the same issue to track it: https://github.com/sonic-net/sonic-buildimage/issues/24275, we can close the issue https://github.com/sonic-net/sonic-buildimage/issues/26254 based on this pr.

vvolam:
> @vvolam its not narrow, its the complete offline cycle, assuming reboot -d DPUX is executed, chassisd does not detect dpu offline until dpu is shutdown, so technically the window is time taken for DPU to actually go offline + ~10 seconds. But I am okay to have a seperate PR, we can use the same issue to track it: [sonic-net/sonic-buildimage#24275](https://github.com/sonic-net/sonic-buildimage/issues/24275), we can close the issue [sonic-net/sonic-buildimage#26254](https://github.com/sonic-net/sonic-buildimage/issues/26254) based on this pr.

@gpunathilell I have also raised https://github.com/sonic-net/sonic-platform-common/pull/680 to complete the scenario

gpunathilell:
@vvolam please remove the `prev_reboot_time.txt` file generation from this file, so that we only have one source for the file creation (from module_base.py) other changes look good. @chartsai-nvidia  will test and once we confirm, will approve the PR

vvolam:
> @vvolam please remove the `prev_reboot_time.txt` file generation from this file, so that we only have one source for the file creation (from module_base.py) other changes look good. @chartsai-nvidia will test and once we confirm, will approve the PR

@gpunathilell Thank you, I have addressed the change.

rameshraghupathy:
@vvolam  I still think we are mixing two different issues here.

The config-reload issue is clear: chassisd restarts while DPU reboot is in progress, STATE_DB is empty, and the transition becomes EMPTY→Offline instead of Online→Offline. That missing transition can be handled directly in SmartSwitchModuleUpdater.module_db_update(), where we already read prev_status via get_module_current_status(key), update the module table, and compare previous/current operational state.

IMO the fix should be limited to adding EMPTY→Offline handling in the existing transition logic.

I do not agree with changing the broader reboot-cause model to defer get_reboot_cause until the DPU comes back online. The current Online→Offline behavior is intentional. Reboot cause is latched/mirrored through the NPU path specifically so it can be retrieved even when the DPU is offline or dead. That is required for debugging unreachable DPUs and is consistent with the existing design.

So the concern is not whether the new deferred flow can be validated. The concern is that it changes the design contract. If a DPU never comes back online, deferring reboot-cause collection defeats the purpose of the latched reboot-cause mechanism.

Can we keep the existing Online→Offline reboot-cause persistence behavior unchanged and only add the missing EMPTY→Offline/config-reload case as a minimal targeted change?

sonic-chassisd/scripts/chassisd
```diff
@@ -774,69 +780,143 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            return reboot_cause[0]
+        return reboot_cause
+
+    def _is_known_to_offline(self, prev_status, current_status):
+        """Online (or any non-EMPTY non-OFFLINE) → OFFLINE — chassisd witnessed it."""
+        return (prev_status != ModuleBase.MODULE_STATUS_EMPTY
+                and prev_status != str(ModuleBase.MODULE_STATUS_OFFLINE)
+                and current_status == str(ModuleBase.MODULE_STATUS_OFFLINE))
+
+    def _is_empty_to_offline(self, prev_status, current_status):
```
rameshraghupathy, line 795: @vvolum IMO this can be handled with a minimal addition in the existing module_db_update() logic itself.

Today, module_db_update() already:
- reads prev_status using get_module_current_status(),
- updates the module table,
- and compares previous/current oper-status for transition handling.

The new EMPTY→Offline case is simply another transition condition in that same flow.

The helper functions introduced in this PR ultimately still reduce to checking:

    prev_status == EMPTY and current_status == Offline

That does not require introducing a new transition framework, deferred reboot-cause processing, or restructuring the existing reboot-cause handling model.

A couple of targeted lines in the existing transition handling path should be sufficient while preserving the current Online→Offline reboot-cause persistence semantics and overall design behavior.

rameshraghupathy:
@vvolam I am not against checking/reading reboot-cause after the DPU powers back on.

The important requirement is that the reboot-cause event must be latched/recorded when the DPU transitions from a known non-Offline state to Offline. That transition is the point where we know the DPU went down, and the DPU may never come back online.

Reading or reconciling reboot-cause later after power-on is fine as an additional step, but it should not replace the existing down-transition latching behavior.

So the design should be:

1. On known non-Offline → Offline:
   latch/persist the reboot-cause event or at least preserve the latched reboot-cause context.

2. On Offline/Empty → Online:
   optionally read/reconcile/update reboot-cause if needed.

But the Online path must not be the only place where reboot-cause is captured, because that loses the dead-DPU debugging use case.

gpunathilell:
> @vvolam I am not against checking/reading reboot-cause after the DPU powers back on.
> 
> The important requirement is that the reboot-cause event must be latched/recorded when the DPU transitions from a known non-Offline state to Offline. That transition is the point where we know the DPU went down, and the DPU may never come back online.
> 
> Reading or reconciling reboot-cause later after power-on is fine as an additional step, but it should not replace the existing down-transition latching behavior.
> 
> So the design should be:
> 
> 1. On known non-Offline → Offline:
>    latch/persist the reboot-cause event or at least preserve the latched reboot-cause context.
> 2. On Offline/Empty → Online:
>    optionally read/reconcile/update reboot-cause if needed.
> 
> But the Online path must not be the only place where reboot-cause is captured, because that loses the dead-DPU debugging use case.

This is a platform specific handling and not the use-case for reboot cause, the get_reboot_cause is used for strictly obtaining the previous reboot cause only so it is implied that is checked when DPU is online, `Retrieves the cause of the previous reboot of the DPU module` as the docstring mentions, so it does not make sense to call it when the DPU is going from any state to Offline. If you want to save/record something while the DPU is powered off (so that it can be used later for actually determining the reboot cause) we can have a callback implemented for `operational_status_offline` which the platform can override and have its own implementation. you can use that to save information. and use it later when the `get_reboot_cause` is called. Would that work? @vvolam for viz

sonic-chassisd/scripts/chassisd
vvolam: @rameshraghupathy after our discussion in the meeting offline, do you have any other suggestions?

rameshraghupathy:
> > @vvolam I am not against checking/reading reboot-cause after the DPU powers back on.
> > The important requirement is that the reboot-cause event must be latched/recorded when the DPU transitions from a known non-Offline state to Offline. That transition is the point where we know the DPU went down, and the DPU may never come back online.
> > Reading or reconciling reboot-cause later after power-on is fine as an additional step, but it should not replace the existing down-transition latching behavior.
> > So the design should be:
> > 
> > 1. On known non-Offline → Offline:
> >    latch/persist the reboot-cause event or at least preserve the latched reboot-cause context.
> > 2. On Offline/Empty → Online:
> >    optionally read/reconcile/update reboot-cause if needed.
> > 
> > But the Online path must not be the only place where reboot-cause is captured, because that loses the dead-DPU debugging use case.
> 
> This is a platform specific handling and not the use-case for reboot cause, the get_reboot_cause is used for strictly obtaining the previous reboot cause only so it is implied that is checked when DPU is online, `Retrieves the cause of the previous reboot of the DPU module` as the docstring mentions, so it does not make sense to call it when the DPU is going from any state to Offline. If you want to save/record something while the DPU is powered off (so that it can be used later for actually determining the reboot cause) we can have a callback implemented for `operational_status_offline` which the platform can override and have its own implementation. you can use that to save information. and use it later when the `get_reboot_cause` is called. Would that work? @vvolam for viz

```
Thanks Gagan. I understand the generic interpretation of get_reboot_cause() as retrieving the previous reboot cause, and for many platforms that may naturally imply reading it when the module is online.

However, SmartSwitch has an explicit platform requirement that is different from a generic module model. The SmartSwitch PMON HLD says the NPU hardware should be capable of providing the DPU reboot-cause even when the DPUs are dead, and get_reboot_cause returns the current reboot-cause of the module.

So for SmartSwitch, calling get_reboot_cause while the DPU is offline is intentional platform-specific behavior, not an accidental misuse of the API.

I am okay with adding a platform callback like operational_status_offline if we want to make this cleaner architecturally. But the callback must preserve the existing behavior: when the DPU transitions from known non-Offline → Offline, the platform must be able to latch/persist the reboot-cause context immediately.

What I do not think is safe is changing the behavior so reboot-cause is only captured after Offline/Empty → Online. That loses the key SmartSwitch requirement where the DPU may be dead or may never come back online.

Also, the config-reload issue can still be fixed minimally by detecting EMPTY→Offline in the existing module_db_update() transition logic. A larger deferred reboot-cause framework should not be required for that narrow issue.
```

rameshraghupathy:
@vvolam I tested this on a 202511 SmartSwitch image where `reboot -d DPUx` works with the original 202511 chassisd.

When I replaced only chassisd with the master chassisd containing the #771 changes, `reboot -d DPUx` started failing. The reboot path timed out and invoked fallback powercycle:

    Pre-shutdown: timed out waiting for DPU 
    reboot: status already failed for DPU2

This reinforces my concern that #771 is not a safe/minimal fix.

The old chassisd persists reboot time and reboot cause immediately on known non-Offline→Offline transition. The #771 code changes that behavior and only records a deferred pending reboot check. That changes the established down-transition reboot-cause model.

For the config-reload issue, I still think the fix should be minimal:
- preserve existing known non-Offline→Offline behavior unchanged,
- add only the missing EMPTY→Offline handling in module_db_update(),
- avoid deferring or restructuring the normal down-transition reboot-cause path.

Given this test result, I do not think #771 should be merged in its current form.

The key counterpoint is already in the SmartSwitch PMON HLD:

“The NPU hardware should be capable of providing the DPU reboot-cause even when the DPUs are dead.”
“The get_reboot_cause will return the current reboot-cause of the module.”

So for SmartSwitch, this is not just “previous reboot cause when DPU is online.” The platform was explicitly designed to expose reboot cause even when the DPU is dead/offline.

gpunathilell:
@rameshraghupathy I have a few concerns with that solution. I tested this on a smartswitch and I think there are three issues we are leaving unresolved.

1. **800 seconds timeout doesnt work reliably** - this is the only check separating "same reboot we already persisted" from "different reboot we need to record" in the Offline→Online and EMPTY→Online paths. Two problems with it:
   - A real second reboot of the same cause within 800s gets dropped (the bug @chartsai-nvidia reported, partially addressed by 1627248c).
   - A spurious reboot cause entry gets created on config reload (verified on smartswitch). DPU was online, ran `sudo config reload -y`, a new entry appeared in `/host/reboot-cause/module/dpuX/history` with the current timestamp and the symlink moved to it, even though no reboot happened. The path: STATE_DB flush → new chassisd sees `prev_status=EMPTY, current_status=Online` → enters EMPTY→Online → previously persisted entry is older than 800s so `is_reboot=True` → persist runs and writes a fake entry.

2. **Reboot cause gets missed when config reload is executed** - this is the original bug this PR was expected to solve. If chassisd is restarted via config reload while a DPU reboot is in progress, STATE_DB is flushed and chassisd loses track of the in-progress reboot, so the actual reboot cause is not recorded.

3. **Two reboot causes get persisted for one reboot if the offline and online queries return different strings** - `get_reboot_cause` is called twice per reboot cycle (line 807 on Online→Offline, line 815 on Offline→Online). The dedup at lines 823-832 only fires when both cause strings match AND delta < 800s. If `get_reboot_cause` returns anything different across the two calls, we end up with two history entries with two different timestamps for the same reboot.

@rameshraghupathy the "only add EMPTY→Offline handling, leave everything else as is" idea doesnt resolve any of these. The Online→Offline persist you want to preserve is exactly what creates #3 - the cause we capture at offline transition is the cause of the PREVIOUS reboot (per the docstring `Retrieves the cause of the previous reboot of the DPU module`), not the reboot currently in progress. We then re-query on the online side, may get a different string back, and persist a second time.

If we want to capture state at the moment the DPU goes offline (for dead-DPU debugging), thats fine - but its not "reboot cause", its an "operational state change" event. That needs a separate API (an oper-state-change callback, or a different function), not overloading `get_reboot_cause` for both. The reboot cause of one reboot event should be one value, captured when its authoritative (DPU online, we can still have the latching mechanism you mentioned).

I can suggest a design that solves all three - we can discuss in the meeting

vvolam:
> @vvolam I tested this on a 202511 SmartSwitch image where `reboot -d DPUx` works with the original 202511 chassisd.
> 
> When I replaced only chassisd with the master chassisd containing the #771 changes, `reboot -d DPUx` started failing. The reboot path timed out and invoked fallback powercycle:
> 
> ```
> Pre-shutdown: timed out waiting for DPU 
> reboot: status already failed for DPU2
> ```
> 
> This reinforces my concern that #771 is not a safe/minimal fix.
> 
> The old chassisd persists reboot time and reboot cause immediately on known non-Offline→Offline transition. The #771 code changes that behavior and only records a deferred pending reboot check. That changes the established down-transition reboot-cause model.
> 
> For the config-reload issue, I still think the fix should be minimal:
> 
> * preserve existing known non-Offline→Offline behavior unchanged,
> * add only the missing EMPTY→Offline handling in module_db_update(),
> * avoid deferring or restructuring the normal down-transition reboot-cause path.
> 
> Given this test result, I do not think #771 should be merged in its current form.
> 
> The key counterpoint is already in the SmartSwitch PMON HLD:
> 
> “The NPU hardware should be capable of providing the DPU reboot-cause even when the DPUs are dead.” “The get_reboot_cause will return the current reboot-cause of the module.”
> 
> So for SmartSwitch, this is not just “previous reboot cause when DPU is online.” The platform was explicitly designed to expose reboot cause even when the DPU is dead/offline.

@rameshraghupathy there are preceding PRs: https://github.com/sonic-net/sonic-platform-common/pull/680
https://github.com/sonic-net/sonic-utilities/pull/4571. You have to test all 3 PRs together to have previous behavior

@vvolam Yes, with #680 and sonic-utilities#4571 patched, the specific command-initiated test may pass because reboot -d DPUx will write prev_reboot_time.txt before config reload.

But that only covers command-initiated reboot/shutdown. It does not address spontaneous DPU crash/kernel panic/dead-DPU cases where there is no module_base.py command invocation.

I am still concerned that #771 removes chassisd as a writer of prev_reboot_time.txt. The original #680 description even said chassisd write is preserved as fallback for spontaneous DPU crashes. But the later #771 commit changes that and makes module_base.py the exclusive writer.

So passing the config-reload test with all three PRs does not prove the design is safe. It only proves the command-path race is covered.

PR #783, opened 2026-03-31
SONiC BMC Support : New Module type to represent switch CPU card

chander-nexthop:
Add a new module type switch_host. This will be used to abstract the switch cpu as a module in SONiC BMC.
HLD Ref. https://github.com/sonic-net/SONiC/pull/2215/

#### Description
A new module type.

#### Motivation and Context
SONiC running on a BMC card would have a module to represent the Switch CPU card. So to support that abstraction, introduce a new module type

### DIFF
--- a/sonic-chassisd/scripts/chassis_db_init
+++ b/sonic-chassisd/scripts/chassis_db_init
@@ -32,6 +32,7 @@ CHASSIS_INFO_CARD_NUM_FIELD = 'module_num'
 CHASSIS_INFO_SERIAL_FIELD = 'serial'
 CHASSIS_INFO_MODEL_FIELD = 'model'
 CHASSIS_INFO_REV_FIELD = 'revision'
+CHASSIS_INFO_SWITCH_HOST_SERIAL_FIELD = 'switch_host_serial'
 
 CHASSIS_LOAD_ERROR = 1
 
@@ -65,7 +66,6 @@ def try_get(callback, *args, **kwargs):
 #
 # Functions
 # 
-
 def provision_db(platform_chassis, log):
     # Init state db connection
     state_db = daemon_base.db_connect("STATE_DB")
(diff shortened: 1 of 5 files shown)


sonic-chassisd/scripts/chassis_db_init
```diff
@@ -65,17 +66,24 @@ def try_get(callback, *args, **kwargs):
...
+    else:
+        switch_host_serial = NOT_AVAILABLE
+
+    # Populate DB with chassis hardware info
     fvs = swsscommon.FieldValuePairs([
                                         (CHASSIS_INFO_SERIAL_FIELD, try_get(platform_chassis.get_serial)),
                                         (CHASSIS_INFO_MODEL_FIELD, try_get(platform_chassis.get_model)),
-                                        (CHASSIS_INFO_REV_FIELD, try_get(platform_chassis.get_revision))
+                                        (CHASSIS_INFO_REV_FIELD, try_get(platform_chassis.get_revision)),
+                                        (CHASSIS_INFO_SWITCH_HOST_SERIAL_FIELD, switch_host_serial)
```
judyjoseph, line 86: Can we add the switch_host_serial only for bmc @chander-nexthop

roger-nexthop (approved):

sonic-chassisd/scripts/chassis_db_init
chander-nexthop: Then the FVP pairs itself will change. The NOT_AVAILABLE default will prevent it from being shown on any non-bmc platform. So I would prefer to keep this as is.

judyjoseph (approved):

vivekverma-arista:
```
    # Determine switch host serial number
    if platform_chassis.is_bmc():
        # BMC must also populate the switch host serial number
        switch_host_serial = try_get(platform_chassis.get_switch_host_serial)
```
```
root@dut8:~# docker exec -it pmon /usr/local/bin/chassis_db_init
File "/usr/local/bin/chassis_db_init", line 113, in
main()
File "/usr/local/bin/chassis_db_init", line 110, in main
provision_db(platform_chassis, log)
File "/usr/local/bin/chassis_db_init", line 77, in provision_db
switch_host_serial = try_get(platform_chassis.get_switch_host_serial)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Chassis' object has no attribute 'get_switch_host_serial'
```

We don't seem to be having this function in chassis_base.py @judyjoseph @chander-nexthop

PR #785, opened 2026-04-02
[thermalctld] Add per-component polling intervals from platform.json

vvolam:
#### Description

Add support for configurable per-component polling intervals in thermalctld via platform.json. Each component type can specify its own polling rate:

- `fan_drawers[0].polling_interval`: interval in seconds for fan updates
- `psus[0].polling_interval`: interval in seconds for PSU thermal updates
- `thermals[*].polling_interval`: per-thermal sensor interval in seconds

Config entries (without a `name` key) are inserted at the beginning of `fan_drawers` / `psus` arrays. Each thermal device entry with a `name` can have its own `polling_interval`.

When any custom thermal intervals are configured, thermals without an explicit `polling_interval` fall back to the original 60s default instead of running at the fast-loop rate. This prevents unconfigured sensors from being polled too frequently when the main loop interval is reduced to accommodate fast-polling sensors.

The main loop interval adjusts to the minimum of all configured intervals so fast-polling components are serviced on time, while slower components are throttled via per-component timestamp tracking.

#### Motivation and Context

On platforms with many thermal sensors, polling all sensors every cycle can add unnecessary overhead. Some sensors (e.g., ASIC temperature) need frequent monitoring while others (e.g., ambient temps, SODIMM) can be polled less often. This change allows platform vendors to fine-tune polling rates per component via platform.json without code changes.

1. **Unit tests**: Added tests for `_parse_platform_json_polling_intervals()`, `_should_update_thermal()` (including `default_interval` behavior), and PSU interval gating. All pass locally via `pytest`.

2. **Testbed verification on SN5640**: Ran tests on a physical testbed with the following platform.json intervals:
   - `fan_drawers`: `polling_interval: 40`
   - `psus`: `polling_interval: 30`
   - `thermals`: ASIC=5s, Ambient Fan Side=10s, Port Side=15s, CPU Pack=20s, SODIMM 2=25s

   Verified:
   - Each component updates at its configured rate (±tolerance for execution overhead)
   - Unconfigured thermals fall back to 60s default
   - Cross-check: fastest thermal (ASIC=5s) updates more often than fans (40s)
   - No errors in thermalctld syslog

3. **Backward compatibility**: Without any `polling_interval` in platform.json, behavior is identical to before (all components update every 60s cycle).

Example platform.json snippet:
```json
{
    "chassis": {
        "fan_drawers": [
            {"polling_interval": "40"},
            {"name": "drawer1", "fans": [{"name": "fan1"}]}
        ],
        "psus": [
            {"polling_interval": "30"},
            {"name": "PSU 1", "thermals": [{"name": "PSU-1 Temp"}]}
        ],
        "thermals": [
            {"name": "ASIC", "polling_interval": "5"},
            {"name": "CPU Pack Temp", "polling_interval": "20"}
        ]
    }
}
```

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -7,12 +7,13 @@
 
 from enum import Enum, auto
 import argparse
+import json
+import os
 import signal
 import sys
 import threading
 import time
 from datetime import datetime
-import argparse
 
 import sonic_platform
 from sonic_py_common import daemon_base, logger, device_info, multi_asic
@@ -38,6 +39,7 @@ TRANSCEIVER_DOM_TEMPERATURE_TABLE = 'TRANSCEIVER_DOM_TEMPERATURE'
 TRANSCEIVER_DOM_THRESHOLD_TABLE = 'TRANSCEIVER_DOM_THRESHOLD'
 TRANSCEIVER_DOM_SENSOR_TABLE = 'TRANSCEIVER_DOM_SENSOR'
 INVALID_SLOT_OR_DPU = -1
+PLATFORM_JSON_FILE = "/usr/share/sonic/platform/platform.json"
 
 ERR_UNKNOWN = 1
 CHASSIS_GET_ERROR = 2
@@ -71,6 +73,70 @@ def update_entity_info(table, parent_name, key, device, device_index):
         [('position_in_parent', str(try_get(device.get_position_in_parent, device_index))),
          ('parent_name', parent_name)])
     table.set(key, fvs)
+
+
+def _parse_platform_json_polling_intervals():
+    """Parse polling_interval values from platform.json for thermalctld components.
+
+    platform.json may contain per-component polling intervals:
+    - fan_drawers[0].polling_interval: interval in seconds for fan updates
+    - psus[0].polling_interval: interval in seconds for PSU updates
+    - thermals[*].polling_interval: per-thermal interval in seconds
+
+    The first entry in fan_drawers/psus without a 'name' key is treated as
+    configuration; subsequent entries with 'name' are actual devices.
+
+    Returns:
+        dict: {'fan_drawer': float|None, 'psu': float|None, 'thermals': {name: float}}
+    """
+    result = {'fan_drawer': None, 'psu': None, 'thermals': {}}
+    try:
+        if not os.path.isfile(PLATFORM_JSON_FILE):
+            return result
+        with open(PLATFORM_JSON_FILE, 'r') as f:
+            platform_data = json.load(f)
+    except Exception:
+        return result
+
+    if not platform_data:
+        return result
+
+    # platform.json nests components under 'chassis'
+    chassis = platform_data.get('chassis', platform_data)
+
+    # fan_drawers: config entry (no 'name') with polling_interval
+    for entry in chassis.get('fan_drawers', []):
+        if 'name' not in entry and 'polling_interval' in entry:
+            val = entry['polling_interval']
+            if val:
+                try:
+                    result['fan_drawer'] = float(val)
+                except (ValueError, TypeError):
+                    pass
+            break
+
+    # psus: config entry (no 'name') with polling_interval
+    for entry in chassis.get('psus', []):
+        if 'name' not in entry and 'polling_interval' in entry:
+            val = entry['polling_interval']
+            if val:
+                try:
+                    result['psu'] = float(val)
+                except (ValueError, TypeError):
+                    pass
+            break
+
+    # thermals: each device entry may have its own polling_interval
+    for entry in chassis.get('thermals', []):
+        name = entry.get('name')
+        val = entry.get('polling_interval', '')
+        if name and val:
+            try:
+                result['thermals'][name] = float(val)
+            except (ValueError, TypeError):
+                pass
+
+    return result
 
 
 class FanType(Enum):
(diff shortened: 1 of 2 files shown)


sonic-thermalctld/scripts/thermalctld
```diff
@@ -1026,12 +1126,51 @@ class ThermalMonitor(ThreadTaskBase):
...
 
     def main(self):
         begin = time.time()
-        self.fan_updater.update()
+
+        if self._fan_update_interval is None or (begin - self._last_fan_update) >= self._fan_update_interval:
+            self.fan_updater.update()
+            self._last_fan_update = begin
+
         self.temperature_updater.update()
```
Junchao-Mellanox: I did a test for this. I set ASIC polling interval to 3 seconds, but I see ASIC actual temperature update interval is randomly 3 seconds or 6 seconds. I believe this is caused by this code block. For exampe:
Iteration 1:
begin=0, elapsed=1, self.wait_time=(3-1)=2, self.fan_updater.update() took 0.5 second, asic temperature is updated at time 0.6.
Iteration 2: 
begin=3, self.fan_updater.update() took 0.3 seconds (0.2 seconds faster than iteration 1), asic tempareture update check starts at 3 + 0.3 + 0.1 = 3.4 seconds. In this case, `_should_update_thermal` returns False because 3.4 - 0.6 < 3. Because of this, asic temperature will be updated only in Iteration3 which is 3 seconds later.

I think this is an issue for time sensitive polling.

Junchao-Mellanox (changes requested):

sonic-thermalctld/scripts/thermalctld
vvolam: Good catch! The root cause is that `_should_update_thermal()` was calling `time.time()` at the moment each thermal is checked, but `fan_updater.update()` runs before thermals and its execution time varies between iterations. This causes the elapsed time seen by thermal checks to fluctuate, leading to the 3s→6s jitter you observed.

Fixed by threading the cycle `begin` timestamp from `ThermalMonitor.main()` through `update(now=begin)` → `_collect_thermals(..., now=now)` → `_should_update_thermal(name, now=now)`. All interval checks now use the same cycle-start reference point, so they are immune to mid-cycle execution time variance.

The `now` parameter defaults to `None` (falls back to `time.time()`) for backward compatibility.

sonic-thermalctld/scripts/thermalctld
```diff
@@ -1026,13 +1133,52 @@ class ThermalMonitor(ThreadTaskBase):
...
         self.logger.set_min_log_priority_info()
 
+        polling_intervals = polling_intervals or {}
+        fan_drawer_interval = polling_intervals.get('fan_drawer')
+        psu_interval = polling_intervals.get('psu')
+        thermal_intervals = polling_intervals.get('thermals', {})
+
+        # Fan update interval: minimum of fan_drawer and psu intervals
+        fan_intervals = [i for i in [fan_drawer_interval, psu_interval] if i is not None]
+        self._fan_update_interval = min(fan_intervals) if fan_intervals else None
```
Junchao-Mellanox: default value should be 60s instead of None

sonic-thermalctld/scripts/thermalctld
```diff
@@ -735,19 +814,42 @@ class TemperatureUpdater(logger.Logger):
...
-            self._refresh_temperature_status(parent_name, thermal, thermal_index)
+            name = try_get(thermal.get_name, '{} Thermal {}'.format(parent_name, thermal_index + 1))
+            # Always update entity info so PHYSICAL_ENTITY_INFO is populated
+            if 'SFP' not in parent_name:
+                update_entity_info(self.phy_entity_table, parent_name, name, thermal, thermal_index + 1)
+            if refresh and self._should_update_thermal(name, now=now):
+                self._refresh_temperature_status(parent_name, thermal, thermal_index)
         return True
 
     def _collect_sfp_thermals(self, available_thermals, parent_name, sfp_index, thermals):
```
Junchao-Mellanox: this function should also use 60s as default polling interval

sonic-thermalctld/scripts/thermalctld
vvolam: Done. Changed `_fan_update_interval` to default to `update_interval` (60s) instead of `None`, so fans are properly throttled even when no custom fan/psu interval is configured.

sonic-thermalctld/scripts/thermalctld
vvolam: Done. Added `_should_update_thermal()` gating to `_collect_sfp_thermals()` with a `now` parameter, so SFP thermals also respect the configured default polling interval (60s). Updated call sites in `update()` to pass `now=now`.

Junchao-Mellanox (approved):

judyjoseph:
changes LGTM ...though there are few scnario's we there could have some latency as we have a single thread handling all these timers ..  If we have asic timer at 5sec .. and for SFP it is 60sec, the same thread does a redis get for a large number of SFP modules (on a DUT with say 512 interfaces) will affect the next polling cycle of asic timer -- since this operation could take > 5 sec ? 

May be we should remove the SFP polling from redis .. now that we would like this optics temperature to be polled by xcvrd itself. We can add a separate thread thermalctld in future if we really need to get the xcvr temperature in thermalctld. 

@vvolam @Junchao-Mellanox  let me know thoughts

vvolam:
Thanks @judyjoseph for the review and the good callout!

You're right that on high-port-count platforms (e.g. 512 SFPs), the SFP Redis reads could add up and potentially delay the next fast-polling cycle. Today each SFP costs 3-5 Redis round-trips per cycle, so at scale that's non-trivial.

That said, this concern exists independent of this PR — the SFP polling already runs in the same thread at 60s today. The per-component interval feature doesn't make it worse; it just makes the impact more visible when you have a fast-polling sensor like ASIC at 5s.

I agree the right approach is to remove SFP temperature polling from thermalctld entirely and let xcvrd own it (there's already a TODO in the code for this). I'd prefer to handle that as a separate PR since it's a different scope of change.

Would it be okay to merge this as-is and I'll file a follow-up to remove the SFP polling path?

Junchao-Mellanox:
Hi @vvolam , could you please also update sonic-platform-daemon submodule pointer from master? Thanks!

vvolam:
@judyjoseph This automated submodule update PR https://github.com/sonic-net/sonic-buildimage/pull/27053 updates the sonic-platform-daemons submodule pointer in master branch.

PR #793, opened 2026-04-09
[202511 only] [cmis] Optimize data path state machine by reducing the time spent checking the fast reboot flag

Junchao-Mellanox:
#### Description
We observe an issue that toggling a logical port on SN5640 takes more than 40 seconds to be up. The log shows a significant delay between each CMIS state machine transition. The PR is to optimize the flow.

#### Motivation and Context
In `CmisManagerTask.process_single_lport`, it checks fast reboot flag for each logical port even if the port does not need run data path state machine. There are 2 issues:

1. We don't have to check fast reboot flag for each logical port. We can move this check out of the for loop. It reduces the call times from N to 1 where N is the logical port number
2. The existing implementation for checking fast reboot flag is via subproces which potentially create a new process. This is too heavy. We can replace it with existing function provided in `device_info.py` which use DB API to get the flag.

Manual test
Unit test

### DIFF
--- a/sonic-xcvrd/tests/test_xcvrd.py
+++ b/sonic-xcvrd/tests/test_xcvrd.py
@@ -5345,6 +5345,7 @@ class TestXcvrdScript(object):
     @patch('sonic_py_common.device_info.get_paths_to_platform_and_hwsku_dirs', MagicMock(return_value=('/tmp', '/tmp')))
     @patch('swsscommon.swsscommon.WarmStart', MagicMock())
     @patch('xcvrd.xcvrd.DaemonXcvrd.wait_for_port_config_done', MagicMock())
+    @patch('xcvrd.xcvrd_utilities.common.is_fast_reboot_enabled', MagicMock(return_value=True))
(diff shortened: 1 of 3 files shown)


Junchao-Mellanox:
Hi @judyjoseph , the PR is for 202511 only, master does not need this

keboliu (approved):

prgeor:
> Hi @judyjoseph , the PR is for 202511 only, master does not need this

@Junchao-Mellanox is the fix already in master? Which PR? Please specify that in PR description

prgeor (approved):

Junchao-Mellanox:
> > Hi @judyjoseph , the PR is for 202511 only, master does not need this
> 
> @Junchao-Mellanox is the fix already in master? Which PR? Please specify that in PR description

Hi @prgeor , I tend to think there is another bug in master branch. Please see line https://github.com/sonic-net/sonic-platform-daemons/blob/b2fda37052f80e5745ee72c8e74569859614c296/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py#L76 . I think it means that fast boot flag will always be true after it become true until there is a config reload or reboot. I think we need someone to fix it.

PR #794, opened 2026-04-10
[xcvrd] Support mixed application codes in CMIS decommission logic

mihirpat1:
#### Description

The original is_decommission_required() compared all 8 module lanes against a single application code (app_new) — the app code of whichever logical port triggered the check. On an OSFP-8X module configured with mixed breakout (e.g., 1x400G on lanes 0-3 + 4x100G on lanes 4-7), lanes belonging to sibling ports have a different valid app code. The old logic saw current_app != app_new on those sibling lanes and returned True, triggering an unnecessary decommission cycle.

#### Motivation and Context

**Root Cause**
is_decommission_required(api, app_new) had no awareness of sibling logical ports sharing the same physical port. It assumed all 8 lanes should run the same application code.

**Fix**
Two changes in cmis_manager_task.py:

1. New method get_desired_app_map(api, lport) — Reads CONFIG_DB PORT table to find all sibling logical ports on the same physical port (matched by index). For each sibling, resolves the desired CMIS application code and lane mask, building a per-lane desired app map ([app, app, ..., app] for 8 lanes). This correctly represents mixed configs like [3, 3, 3, 3, 1, 1, 1, 1].

    The get_desired_app_map() method reads sibling port configurations directly from the CONFIG_DB PORT table rather than self.port_dict. This is intentional since port_dict is populated incrementally via asynchronous port change events, so sibling entries may not yet exist or may contain stale data when the first logical port reaches the INSERTED state. CONFIG_DB is the authoritative, atomic source of truth for the desired port configuration and guarantees a consistent view of all siblings at the time of the decommission check.

2. Revised is_decommission_required(api, lport) - It now accepts lport instead of app_new. Calls get_desired_app_map() and compares each lane's current app against its per-lane desired app. Lanes that are currently unused (AppSel=0) are ignored since adding new DPs on unused lanes doesn't require decommission.

TC | Test | Result | Key Observation
-- | -- | -- | --
1 | Non-breakout baseline | PASS | All links are up and active application is in line with expecation
2 | sfputil reset + SNS | PASS | After reset, all lanes defaulted to application code as 1. Post SNS,   decommissioning logic was triggerred and the expected ports were in link up   state.
3 | xcvrd restart | PASS | No decommission and no link flap
4 | pmon restart | PASS | No decommission and no link flap
5 | swss restart | PASS | Decomissioning logic is not triggered and the relevant application modes   are unchanged and the link comes up
6 | config reload | PASS | Decomissioning logic is not triggered and the relevant application modes   are unchanged and the link comes up
7 | Device reboot | PASS | Decommission correctly triggered (module reset to   default [1,1,1,1,1,1,1,1] vs desired [1,1,1,1,3,3,3,3]), then   re-provisioned
8 | Loopback | PASS | Per-DP independence confirmed — loopback on one port doesn't affect   siblings
9 | Admin shut/no-shut | PASS | Individual port shutdown doesn't affect siblings, no spurious   decommission

MSFT ADO - 37489480

### DIFF
--- a/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
+++ b/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
@@ -381,7 +381,7 @@ class CmisManagerTask(threading.Thread):
             == CMIS_STATE_FAILED
         )
 
-    def get_desired_app_map(self, api, lport):
+    def get_desired_app_map(self, api, lport, gearbox_lanes_dict=None):
         """
         Build a per-lane desired application code map for all lanes of the
         physical port that lport belongs to, by reading sibling logical port
@@ -392,6 +392,9 @@ class CmisManagerTask(threading.Thread):
                 XcvrApi object
             lport:
                 String, logical port name triggering this check
+            gearbox_lanes_dict:
+                Optional dictionary of gearbox line lanes counts keyed by
+                logical port. When omitted, it is fetched from table helper.
 
         Returns:
             list of CMIS_MAX_HOST_LANES integers, desired app code per lane
@@ -403,7 +406,8 @@ class CmisManagerTask(threading.Thread):
         if cfg_port_tbl is None:
             self.log_error("{}: cfg_port_tbl is None while building desired app map".format(lport))
             return desired_map
-        gearbox_lanes_dict = self.xcvr_table_helper.get_gearbox_line_lanes_dict()
+        if gearbox_lanes_dict is None:
+            gearbox_lanes_dict = self.xcvr_table_helper.get_gearbox_line_lanes_dict()
 
         for sibling_lport in cfg_port_tbl.getKeys():
             # Single read per key: use full hash and derive index from fields.
@@ -445,7 +449,7 @@ class CmisManagerTask(threading.Thread):
 
         return desired_map
 
-    def is_decommission_required(self, api, lport):
+    def is_decommission_required(self, api, lport, gearbox_lanes_dict=None):
         """
         Check if CMIS decommission (reset AppSel to 0 for all lanes of the
         physical port) is required. Per CMIS spec, a DP's lane width can only
@@ -460,11 +464,14 @@ class CmisManagerTask(threading.Thread):
                 XcvrApi object
             lport:
                 String, logical port name triggering the check
+            gearbox_lanes_dict:
+                Optional dictionary of gearbox line lanes counts keyed by
+                logical port. When omitted, desired app map helper fetches it.
         Returns:
             True, if decommission is required
             False, if decommission is not required
         """
-        desired_map = self.get_desired_app_map(api, lport)
+        desired_map = self.get_desired_app_map(api, lport, gearbox_lanes_dict)
         current_map = [api.get_application(lane) for lane in range(self.CMIS_MAX_HOST_LANES)]
 
         self.log_notice("{}: current app map {}, desired app map {}".format(lport, current_map, desired_map))
@@ -852,7 +859,8 @@ class CmisManagerTask(threading.Thread):
         media_lanes_mask = self.port_dict[lport]['media_lanes_mask']
         self.log_notice("{}: Setting media_lanemask=0x{:x}".format(lport, media_lanes_mask))
 
-        if self.is_decommission_required(api, lport):
+        gearbox_lanes_dict = self.port_dict[lport].get('gearbox_lanes_dict')
+        if self.is_decommission_required(api, lport, gearbox_lanes_dict):
             self.set_decomm_pending(lport)
 
         if self.is_decomm_lead_lport(lport):
@@ -1253,6 +1261,7 @@ class CmisManagerTask(threading.Thread):
             self.port_dict[lport]['host_lane_count'] = host_lane_count
             self.port_dict[lport]['api'] = api
             self.port_dict[lport]['sfp'] = sfp
+            self.port_dict[lport]['gearbox_lanes_dict'] = gearbox_lanes_dict
         except AttributeError:
             # Skip if these essential routines are not available
             self.update_port_transceiver_status_table_sw_cmis_state(lport, CMIS_STATE_READY)
--- a/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
+++ b/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
@@ -61,6 +61,8 @@ class CmisManagerTask(threading.Thread):
         self.platform_chassis = platform_chassis
         self.xcvr_table_helper = XcvrTableHelper(self.namespaces)
         self._is_fast_reboot_enabled = None
+        # Cache of gearbox line lanes dict, refreshed once per task_worker iteration.
+        self._gearbox_lanes_dict = None
 
     def log_debug(self, message):
         helper_logger.log_debug(message)
@@ -197,20 +199,19 @@ class CmisManagerTask(threading.Thread):
     def get_cmis_module_power_down_duration_secs(self, api):
         return api.get_module_pwr_down_duration()/1000
 
-    def get_host_lane_count(self, lport, port_config_lanes, gearbox_lanes_dict):
+    def get_host_lane_count(self, lport, port_config_lanes):
         """
         Get host lane count from gearbox configuration if available, otherwise from port config
 
         Args:
             lport: logical port name (e.g., "Ethernet0")
             port_config_lanes: lanes string from port config (e.g., "25,26,27,28")
-            gearbox_lanes_dict: dictionary of gearbox line lanes counts
 
         Returns:
             Integer: number of host lanes
         """
-        # First try to get from gearbox configuration
-        gearbox_host_lane_count = gearbox_lanes_dict.get(lport, 0)
+        # First try to get from gearbox configuration cache (refreshed each task_worker iteration)
+        gearbox_host_lane_count = (self._gearbox_lanes_dict or {}).get(lport, 0)
         if gearbox_host_lane_count > 0:
             self.log_debug("{}: Using gearbox line lanes count: {}".format(lport, gearbox_host_lane_count))
             return gearbox_host_lane_count
@@ -381,33 +382,26 @@ class CmisManagerTask(threading.Thread):
             == CMIS_STATE_FAILED
         )
 
-    def get_desired_app_map(self, api, lport, gearbox_lanes_dict=None):
+    def get_sibling_port_configs(self, lport):
         """
-        Build a per-lane desired application code map for all lanes of the
-        physical port that lport belongs to, by reading sibling logical port
-        configurations from the CONFIG_DB PORT table.
+        Fetch sibling logical port configurations sharing the same physical port
+        as lport from the CONFIG_DB PORT table.
 
         Args:
-            api:
-                XcvrApi object
             lport:
                 String, logical port name triggering this check
-            gearbox_lanes_dict:
-                Optional dictionary of gearbox line lanes counts keyed by
-                logical port. When omitted, it is fetched from table helper.
 
         Returns:
-            list of CMIS_MAX_HOST_LANES integers, desired app code per lane
-            (0 = unused/unassigned)
+            list of dicts, one per sibling logical port on the same physical port,
+            each with keys: 'lport' (str), 'subport' (int), 'speed' (int),
+            'host_lane_count' (int).
         """
-        desired_map = [0] * self.CMIS_MAX_HOST_LANES
+        siblings = []
         pport = self.port_dict[lport].get('index')
         cfg_port_tbl = self.xcvr_table_helper.get_cfg_port_tbl(self.get_asic_id(lport))
         if cfg_port_tbl is None:
-            self.log_error("{}: cfg_port_tbl is None while building desired app map".format(lport))
-            return desired_map
-        if gearbox_lanes_dict is None:
-            gearbox_lanes_dict = self.xcvr_table_helper.get_gearbox_line_lanes_dict()
+            self.log_error("{}: cfg_port_tbl is None while fetching sibling port configs".format(lport))
+            return siblings
 
         for sibling_lport in cfg_port_tbl.getKeys():
             # Single read per key: use full hash and derive index from fields.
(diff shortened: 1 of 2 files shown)


sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -381,24 +381,84 @@ def is_decomm_failed(self, lport):
...
+            # Filter by physical port index to ensure sibling logical ports of the same physical port are considered
+            found, idx_str = cfg_port_tbl.hget(sibling_lport, 'index')
+            if not found or int(idx_str) != pport:
+                continue
+
+            found, port_info = cfg_port_tbl.get(sibling_lport)
+            if not found:
+                continue
+            port_info_dict = dict(port_info)
+
```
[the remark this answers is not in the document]
mihirpat1: Addressed it now

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -381,24 +381,84 @@ def is_decomm_failed(self, lport):
...
+            sibling_speed = int(port_info_dict.get('speed', 0))
+            sibling_lanes = port_info_dict.get('lanes', '')
+            sibling_subport = int(port_info_dict.get('subport', 0))
+
+            if not sibling_speed or not sibling_lanes:
+                continue
+
+            sibling_host_lane_count = len(sibling_lanes.split(','))
+            sibling_appl = common.get_cmis_application_desired(api, sibling_host_lane_count, sibling_speed)
```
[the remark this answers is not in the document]
mihirpat1: Addressed it now

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -381,24 +381,98 @@ def is_decomm_failed(self, lport):
...
+            if sibling_pport is None:
+                continue
+
+            try:
+                if int(sibling_pport) != pport:
+                    continue
+            except (TypeError, ValueError):
+                continue
+
+            try:
```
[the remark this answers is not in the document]
mihirpat1: We cannot use port_dict as source of truth as stated in the PR description.

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -381,24 +381,98 @@ def is_decomm_failed(self, lport):
...
+            (0 = unused/unassigned)
+        """
+        desired_map = [0] * self.CMIS_MAX_HOST_LANES
+        pport = self.port_dict[lport].get('index')
+        cfg_port_tbl = self.xcvr_table_helper.get_cfg_port_tbl(self.get_asic_id(lport))
+        if cfg_port_tbl is None:
+            self.log_error("{}: cfg_port_tbl is None while building desired app map".format(lport))
+            return desired_map
+        gearbox_lanes_dict = self.xcvr_table_helper.get_gearbox_line_lanes_dict()
+
```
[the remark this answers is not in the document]
mihirpat1: Resolved now

mihirpat1:
@AnoopKamath @longhuan-cisco Can you please help in reviewing this PR?

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -381,24 +381,112 @@ def is_decomm_failed(self, lport):
...
+                String, logical port name triggering the check
+            gearbox_lanes_dict:
+                Optional dictionary of gearbox line lanes counts keyed by
+                logical port. When omitted, desired app map helper fetches it.
         Returns:
             True, if decommission is required
             False, if decommission is not required
         """
+        desired_map = self.get_desired_app_map(api, lport, gearbox_lanes_dict)
+        current_map = [api.get_application(lane) for lane in range(self.CMIS_MAX_HOST_LANES)]
```
longhuan-cisco: Thanks for picking this up. I like the goal of keeping the fix smaller in scope.

I may be missing something, but I think `get_application()` might not be the best source for this decision.

My understanding from `sonic-platform-common` is that `get_application(lane)` reads the staged control AppSel field rather than the Active Control Set. If so, we can get a case like this:

```python
staged  = [3,3,3,3,1,1,1,1]
active  = [1,1,1,1,1,1,1,1]
desired = [3,3,3,3,1,1,1,1]
```

In that case this check would conclude that decommission is not required, even though the active datapath layout has not converged yet.

That was the main reason PR 663 used `get_active_apsel_hostlane()` / ActiveAppSel for conflict detection.

Would it make sense to base this comparison on ActiveAppSel instead of staged AppSel?

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -381,24 +381,112 @@ def is_decomm_failed(self, lport):
...
+        desired_map = [0] * self.CMIS_MAX_HOST_LANES
+        pport = self.port_dict[lport].get('index')
+        cfg_port_tbl = self.xcvr_table_helper.get_cfg_port_tbl(self.get_asic_id(lport))
+        if cfg_port_tbl is None:
+            self.log_error("{}: cfg_port_tbl is None while building desired app map".format(lport))
+            return desired_map
+        if gearbox_lanes_dict is None:
+            gearbox_lanes_dict = self.xcvr_table_helper.get_gearbox_line_lanes_dict()
+
+        for sibling_lport in cfg_port_tbl.getKeys():
```
longhuan-cisco, line 406: This feels better than using `port_dict`, but I still have some concern about using a live `CONFIG_DB PORT` scan here as if it were a full atomic sibling snapshot.

Have we tested this specifically in a DPB / config-replay sequence where sibling rows are still converging while `on_port_update_event()` is already forcing CMIS reinit on each `PORT_SET`?

The scenario I am worried about is something like:

```python
# transient partial view while siblings are still arriving
desired_map = [3,3,3,3,0,0,0,0]
current_map = [1,1,1,1,1,1,1,1]

# final intended breakout after all sibling rows are present
desired_map = [3,3,3,3,1,1,1,1]
```

In that transient window, missing siblings remain `0`, so we may trigger a decommission based on a partial view rather than the final intended breakout config.

That is the part that makes me a little cautious about this direction: if we need extra logic to wait for sibling config to become complete/stable, it may stop being much simpler than the PR 663 approach.

If DPB was covered in testing already, it would be helpful to call that out.

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
mihirpat1: @longhuan-cisco Thanks! I have addressed this now.

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
mihirpat1: @longhuan-cisco Thanks, this is a valid concern for DPB/config replay convergence windows.
This PR is scoped to static mode only for mixed application code support and does not change baseline steady-state behavior.
I agree this should be handled in a follow-up PR focused on convergence stabilization to support mixed application code.

longhuan-cisco: Thanks for the clarification. After looking at the flow a bit more, I think the earlier DPB/config-replay example is probably not the strongest concern here, since a true DPB operation is already whole-physical-port disruptive today and this path still seems likely to lead to at most the one expected decommission.

At least theoretically, true DPB still looks okay here because the first relevant mismatch should lead to the one expected whole-port decommission, and later sibling rows are mostly `0 -> desired_app` additions, which this path ignores.

That said, it would still be helpful to run a simple baseline non-mixed DPB regression matrix, e.g. `1x800G <-> 2x400G <-> 4x200G <-> 8x100G`, just to make sure this change does not introduce regressions there.

The more practical remaining gap seems to be cases like https://github.com/sonic-net/sonic-buildimage/issues/23006, where one subport changes app/speed but an unchanged sibling should stay up.

One smaller follow-up that might address that in practice would be to tighten the predicate in `is_decommission_required()` around the current mismatch check here:
https://github.com/mihirpat1/sonic-platform-daemons/blob/4d95149bef51dfef0657496912c2d2b51bded842/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py#L499

Something along these lines:

```python
# pseudo-code
if current_app == 0:
    continue
if desired_app == 0:
    return True
if app_width(current_app) != app_width(desired_app):
    return True
```

In other words, if the active app on a lane and the desired app on that lane have the same `host_lane_count`, let the normal per-subport update path handle it; only trigger the whole-port decommission path when the active datapath width changes (or the desired lane becomes unused).

That seems like a smaller practical step, even if it is not as general as a more explicit active-conflict/per-logical-port approach. Would something along those lines make sense for the follow-up?

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -778,7 +878,8 @@ def handle_cmis_inserted_state(self, lport):
         media_lanes_mask = self.port_dict[lport]['media_lanes_mask']
         self.log_notice("{}: Setting media_lanemask=0x{:x}".format(lport, media_lanes_mask))
 
-        if self.is_decommission_required(api, appl):
+        gearbox_lanes_dict = self.port_dict[lport].get('gearbox_lanes_dict')
```
prgeor: @mihirpat1 are we able to maintain same API argument as previous is_decommission_required() had by caching the required gearbox_lanes_dict in self object once for future use as well?

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -381,24 +381,124 @@ def is_decomm_failed(self, lport):
             == CMIS_STATE_FAILED
         )
 
-    def is_decommission_required(self, api, app_new):
+    def get_desired_app_map(self, api, lport, gearbox_lanes_dict=None):
```
prgeor: @mihirpat1 fetching the info from config DB should be happening once during xcvrd init and cached for future use?

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -381,24 +381,124 @@ def is_decomm_failed(self, lport):
...
+            # Single read per key: use full hash and derive index from fields.
+            found, port_info = cfg_port_tbl.get(sibling_lport)
+            if not found:
+                continue
+            port_info_dict = dict(port_info)
+
+            sibling_pport = port_info_dict.get('index')
+            if sibling_pport is None:
+                continue
+
```
prgeor, line 416: @mihirpat1 can we break this single function into two?
1. fetch the info from config DB
2. Use that to construct the map for desired application

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
mihirpat1: @longhuan-cisco The above makes sense for the follow-up.
For the DPB regression matrix which you mentioned, do we already have a HWSKU which supports these `1x800G <-> 2x400G <-> 4x200G <-> 8x100G` modes?

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
mihirpat1: @prgeor I have addressed this now. Also, we are now passing lport instead of appl as the argument.

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
mihirpat1: @prgeor No - caching at xcvrd init would go stale on DPB, since sibling port configuration can change at runtime. Reading from CONFIG_DB at decision time keeps the logic always consistent with the current config without needing extra invalidation handling.

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
longhuan-cisco: @mihirpat1 Yes, from the existing `sonic-buildimage` `platform.json` definitions, I found several candidate platforms that support those modes.

- Arista `x86_64-arista_7060x6_64pe`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/arista/x86_64-arista_7060x6_64pe/platform.json#L321-L337
- Arista `x86_64-arista_7060x6_64pe_b`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/arista/x86_64-arista_7060x6_64pe_b/platform.json#L329-L345
- Marvell `x86_64-marvell_dbmvtx9180-r0`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/marvell/x86_64-marvell_dbmvtx9180-r0/platform.json#L870-L873
- Nexthop `x86_64-nexthop_4010-r0`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/nexthop/x86_64-nexthop_4010-r0/platform.json#L860-L883
- Nexthop `x86_64-nexthop_4010-r1`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/nexthop/x86_64-nexthop_4010-r1/platform.json#L860-L883
- Nexthop `x86_64-nexthop_4020-r0`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/nexthop/x86_64-nexthop_4020-r0/platform.json#L422-L445
- Nexthop `x86_64-nexthop_4220-r0`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/nexthop/x86_64-nexthop_4220-r0/platform.json#L1271-L1304
- Nokia `x86_64-nokia_ixr7220_h5_32d-r0`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/nokia/x86_64-nokia_ixr7220_h5_32d-r0/platform.json#L474-L493
- Nokia `x86_64-nokia_ixr7220_h5_64d-r0`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/nokia/x86_64-nokia_ixr7220_h5_64d-r0/platform.json#L479-L498
- Nokia `x86_64-nokia_ixr7220_h5_64o-r0`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/nokia/x86_64-nokia_ixr7220_h5_64o-r0/platform.json#L479-L498
- Nokia `x86_64-nokia_ixr7220_h6_64-r0`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/nokia/x86_64-nokia_ixr7220_h6_64-r0/platform.json#L692-L708
- Supermicro `x86_64-supermicro_sse_t8164-r0`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/supermicro/x86_64-supermicro_sse_t8164-r0/platform.json#L277-L296
- Supermicro `x86_64-supermicro_sse_t8196-r0`: https://github.com/sonic-net/sonic-buildimage/blob/1975d2b66789bf5178e5bb6ee588f122348d633a/device/supermicro/x86_64-supermicro_sse_t8196-r0/platform.json#L306-L325

A few of them encode the top mode as `1x800G[400G]` or `1x800G[400G,200G]`, but they still advertise the same `2x400G`, `4x200G`, and `8x100G` breakout progression.

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
mihirpat1: @prgeor I have addressed this now.

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
mihirpat1: @longhuan-cisco It seems the platform I tried hasn't enabled end to end support for DPB. Would you be able to help to test this change?

longhuan-cisco: I have tested this change using DPB on one of cisco platforms with 800G optics. 
- 1x800G->2x400G->4x200G->8x100G, 50 iterations
- reverse order, 50 iterations

It all went through fine, link/lldp up, no crash.

longhuan-cisco (approved):

prgeor (approved):

ShauliTaragin:
Hi @mihirpat1 , @prgeor , 
Can you please take this PR to 202511?
BT1 uneven split SKU(Mellanox-SN5640-C508O1X2) is for 202511 and is dependent on this PR.
Thanks

mihirpat1:
> Hi @mihirpat1 , @prgeor , Can you please take this PR to 202511? BT1 uneven split SKU(Mellanox-SN5640-C508O1X2) is for 202511 and is dependent on this PR. Thanks

@ShauliTaragin This PR has already been cherry-picked to 202511 via https://github.com/sonic-net/sonic-platform-daemons/pull/860

PR #799, opened 2026-04-20
[xcvrd] SI settings sync protocol between xcvrd and orchagent

arpit-nexthop:
**What I did**

Implements https://github.com/sonic-net/SONiC/pull/2301

#### Description

  Implements the xcvrd side of a redesigned SI settings synchronization protocol between xcvrd and orchagent. The previous design used a single NPU_SI_SETTINGS_SYNC_STATUS field in STATE_DB with values NPU_SI_SETTINGS_DEFAULT / _NOTIFIED / _DONE, with xcvrd writing directly to STATE_DB. This is replaced by a two-field, counter-based protocol.

  Changes across files:

  xcvr_table_helper.py
  - Removes NPU_SI_SETTINGS_SYNC_STATUS_KEY, NPU_SI_SETTINGS_DEFAULT_VALUE, NPU_SI_SETTINGS_NOTIFIED_VALUE constants
  - Adds a cached swsscommon.Table (app_port_read_tbl) for APPL_DB reads, created once per ASIC at init time — avoids creating a new Table object on every call since ProducerStateTable is write-only
  - Adds get_next_si_notification_number(port_name, asic_id): reads current si_sync_status from APPL_DB and increments its counter, returning 1 if absent or malformed. Increments from any <type>:<N> format (not just NOTIFIED) to handle the removal→re-insertion counter sequence correctly
  - Adds get_appl_db_port_table_val_by_key(lport, port_mapping, key): generic APPL_DB PORT_TABLE key reader using the cached table
  - Rewrites is_npu_si_settings_update_required(): now reads si_sync_status from APPL_DB instead of STATE_DB; returns False only if value is SI_SETTINGS_NOTIFIED:<N>
  - Adds is_si_settings_already_applied(lport, port_mapping): verifies APPL_DB has SI_SETTINGS_NOTIFIED:<N> and STATE_DB has SI_SYNC_DONE:<N> with matching N — used by the restart-skip optimization
  - Removes dead get_state_db_port_table_val_by_key() (no production callers)

  xcvrd.py
  - On SFP removal: writes si_sync_status = SI_SETTINGS_DEFAULT:<N+1> to APPL_DB PORT_TABLE (incrementing the counter) instead of writing NPU_SI_SETTINGS_DEFAULT to STATE_DB
  - on_add_logical_port: removes initialization of NPU_SI_SETTINGS_SYNC_STATUS in STATE_DB (orchagent now owns this initialization via initializePortSiSettingsSyncStatusBulk)
  - Removes initialize_port_init_control_fields_in_port_table() and its call in init() — entirely superseded by orchagent-side initialization

  media_settings_parser.py
  - notify_media_setting(): appends si_sync_status = SI_SETTINGS_NOTIFIED:<N> to the existing APPL_DB PORT_TABLE write (allocates one extra FieldValuePairs slot); removes the separate STATE_DB write of NPU_SI_SETTINGS_NOTIFIED

  cmis_manager_task.py
  - Imports CMIS_STATE_SI_SETTINGS_WAIT (renamed from CMIS_STATE_NPU_SI_SETTINGS_WAIT)
  - Adds CMIS_SI_SETTINGS_WAIT_TIMEOUT = 10 seconds
  - Stores self.port_mapping in __init__ (needed by helper methods)
  - New methods:
    - check_si_settings_app_status(lport): reads si_settings_sync_status from STATE_DB via hget
    - parse_si_notification_number(lport, si_sync_status): extracts <N> from SI_SETTINGS_NOTIFIED:<N>
    - check_si_sync_done_match(lport, si_settings_status, expected_number): validates STATE_DB SI_SYNC_DONE:<N> matches the expected number; returns False (not True) when expected_number is None
    - can_skip_cmis_init_after_restart(lport, api, appl, host_lanes_mask): four-check restart optimization — application match on all active lanes, is_si_settings_already_applied, ConfigSuccess, DataPathActivated
  - handle_cmis_inserted_state(): calls can_skip_cmis_init_after_restart early; if all checks pass, transitions directly to CMIS_STATE_READY and posts active apsel, skipping CMIS reprogramming
  - process_cmis_state_machine() in CMIS_STATE_AP_CONF: after set_application, reads si_sync_status from APPL_DB; if SI_SETTINGS_NOTIFIED:<N> is present, stores the notification number in port_dict and transitions to CMIS_STATE_SI_SETTINGS_WAIT; otherwise goes directly to CMIS_STATE_DP_INIT
  - New CMIS_STATE_SI_SETTINGS_WAIT handler: polls STATE_DB for SI_SYNC_DONE:<N> matching the stored notification number; on match clears the number and transitions to CMIS_STATE_DP_INIT; on 10s timeout proceeds to CMIS_STATE_DP_INIT with a warning (no reinit)

  common.py
  - Adds CMIS_STATE_SI_SETTINGS_WAIT = 'SI_SETTINGS_WAIT' constant

#### Motivation and Context

  The old design had two problems:
  1. xcvrd wrote NPU_SI_SETTINGS_NOTIFIED directly to STATE_DB, which is not read by orchagent, which only subscribes the APPL_DB
  2. CMIS state machine is not waiting for the SI settings to be applied on the ASIC/gearbox before going on with the datapath programming.

Unit tests added/updated in test_xcvrd.py:
  - test_get_next_si_notification_number: increments from SI_SETTINGS_DEFAULT, SI_SETTINGS_NOTIFIED, absent, and malformed formats
  - test_get_appl_db_port_table_val_by_key: None asic_index, port not found, key absent, key present — using cached table directly
  - test_is_npu_si_settings_update_required: None, DEFAULT, NOTIFIED, and unknown format inputs
  - test_is_si_settings_already_applied: full matrix of None APPL_DB status, DEFAULT, malformed number, None asic_index, STATE_DB not found, wrong status, number mismatch, and matching numbers
  - test_can_skip_cmis_init_after_restart: application mismatch, SI not synced, config not success, datapath not activated, all pass
  - test_on_sfp_remove_event_resets_si_sync_status: verifies APPL_DB write of SI_SETTINGS_DEFAULT:<N> on SFP removal
  - Removed test_DaemonXcvrd_initialize_port_init_control_fields_in_port_table (function deleted)
  - CMIS SM task worker tests: set APPL_DB read mock to return not-found so state bypasses CMIS_STATE_SI_SETTINGS_WAIT and goes directly to CMIS_STATE_DP_INIT

Tests done on device:
 - shut / no shut tests: The notification count does not increase after 1.
 - xcvr removal and insertion tests: Checks for the programming of the SI settings and increase in the notification number
 - xcvrd restart test: The transceiver was not programmed again.


sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -589,6 +591,161 @@ def get_port_admin_status(self, lport):
...
+            else:
+                self.log_error("{}: Invalid SI_SYNC_DONE format: {}".format(
+                    lport, si_settings_status))
+                return False
+        except (ValueError, IndexError) as e:
+            self.log_error("{}: Failed to parse SI_SYNC_DONE number from {}: {}".format(
+                lport, si_settings_status, e))
+            return False
+
+    def can_skip_cmis_init_after_restart(self, lport, api, appl, host_lanes_mask):
```
arpit-nexthop: This should already be solved, check it

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -42,6 +42,7 @@ class CmisManagerTask(threading.Thread):
 
     CMIS_MAX_RETRIES     = 3
     CMIS_DEF_EXPIRED     = 60 # seconds, default expiration time
+    CMIS_SI_SETTINGS_WAIT_TIMEOUT = 10 # seconds, timeout for waiting SI settings to be applied on ASIC
```
mihirpat1, line 45: @arpit-nexthop Please move this common.py

mihirpat1 (commented):
**`get_next_si_notification_number` can return the same `N` twice due to producer/consumer lag**

Reads and writes go through different swsscommon objects on different tables:

- Writes use `app_port_tbl` (`ProducerStateTable`) - entries land in `_PORT_TABLE` and are only moved into `PORT_TABLE` after orchagent's `ConsumerStateTable` drains them.
- `get_next_si_notification_number` reads `app_port_read_tbl` (plain `Table` on `PORT_TABLE`) - i.e. the post-consumer view.

So two back-to-back calls (e.g. an initial notify followed shortly by a `force_cmis_reinit` on `SI_SETTINGS_WAIT` timeout) can both observe the same pre-write value and return the same `N`:

```
notify_media_setting() -> reads N=5 from PORT_TABLE -> writes :6 to _PORT_TABLE
SI_SETTINGS_WAIT timeout -> force_cmis_reinit
notify_media_setting() -> PORT_TABLE still shows :5 (OA hasn't drained) -> returns 6 again
```

This breaks the "monotonically increasing counter" invariant the HLD relies on, and `can_skip_cmis_init_after_restart` may then match a stale `SI_SYNC_DONE:N` from the first notification while the second is still pending. The window widens exactly when it matters most - high port-count startup or right after a `SI_SETTINGS_WAIT` timeout, since OA being slow is what caused the timeout in the first place.

mihirpat1:
@arpit-nexthop Please fix the build failure

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -1049,6 +1120,139 @@ def handle_cmis_dp_deinit_state(self, lport):
...
+                lport, si_settings_status, expected_notification_number))
+            self.port_dict[lport].pop('si_notification_number', None)
+            self.update_port_transceiver_status_table_sw_cmis_state(lport, CMIS_STATE_DP_INIT)
+            return True
+        elif self.is_timer_expired(expired):
+            self.log_notice("{}: timeout waiting for SI settings to be applied (status: {}, expected: {})".format(
+                lport, si_settings_status, expected_notification_number))
+            self.port_dict[lport].pop('si_notification_number', None)
+            self.port_dict[lport]['notify_si_settings'] = True
+            self.force_cmis_reinit(lport, retries + 1)
```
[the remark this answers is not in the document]
arpit-nexthop: Will fix the PR description

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
```diff
@@ -625,14 +622,24 @@ def notify_media_setting(logical_port_name, transceiver_dict,
             helper_logger.log_info("Error in obtaining media setting for {}".format(logical_port_name))
             return
```
ymd-arista, line 653: Can we move the logic of getting `notification_number` here? This way we can update `fvs_list` with `si_settings_notification` before we make field value pairs out of it and avoid doing `len(fvs_list) + 1`.

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
```diff
@@ -625,14 +622,24 @@ def notify_media_setting(logical_port_name, transceiver_dict,
...
+        si_settings_notification_value = "SI_SETTINGS_NOTIFIED:{}".format(notification_number)
+        fvs[len(fvs_list)] = ("si_settings_notification", si_settings_notification_value)
+        helper_logger.log_notice("{}:({},{}) ".format(len(fvs_list), "si_settings_notification", si_settings_notification_value))
+
         xcvr_table_helper.get_app_port_tbl(asic_index).set(port_name, fvs)
-        xcvr_table_helper.get_state_port_tbl(asic_index).set(logical_port_name, [(NPU_SI_SETTINGS_SYNC_STATUS_KEY, NPU_SI_SETTINGS_NOTIFIED_VALUE)])
         helper_logger.log_notice("Notify media setting: Published SI setting "
-                                 "for lport {} in APP_DB".format(logical_port_name))
+                                 "for lport {} in APP_DB with notification number {}".format(logical_port_name, notification_number))
+        last_notification_number = notification_number
```
ymd-arista, line 669: We are not consuming return value from `notify_media_settings` anywhere. Is it necessary for us to return `last_notification_number` here? If not, we can remove the use of this variable?

If it's necessary can't we just do `return notification_number` in the previous block itself?

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvr_table_helper.py
```diff
@@ -177,74 +175,41 @@ def get_firmware_info_tbl(self, asic_id):
...
-            lport:
-                logical port name
-            port_mapping:
-                A PortMapping object
-            key:
-                key for the corresponding value to be retrieved
-        Returns:
-            The value of the key if the key is found in STATE_DB PORT_TABLE|<lport>
-            None otherwise
+        Return the next SI notification number for port_name by reading the
```
ymd-arista: Can we also add information on the args in the function doc string? If we ever use some documentation generator, it would be helpful.

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvr_table_helper.py
```diff
@@ -177,74 +175,41 @@ def get_firmware_info_tbl(self, asic_id):
...
+        current_value = fvs_dict.get('si_settings_notification')
+        if not current_value:
+            return 1
+        try:
+            parts = current_value.split(':')
+            if len(parts) == 2:
+                return int(parts[1]) + 1
+        except (ValueError, IndexError):
+            pass
+        return 1
```
ymd-arista: Would it be better if we can create a constant like `DEFAULT_NOTIFICATION_NUMBER` at the top of the file with value as 1 and use it here rather than the magic number?

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvr_table_helper.py
```diff
@@ -177,74 +175,41 @@ def get_firmware_info_tbl(self, asic_id):
...
-            helper_logger.log_error("Get value by key from STATE_DB: Unable to find key {} "
-                                    "state_port_table_fvs_dict {} for lport {}".format(key, state_port_table_fvs_dict, lport))
-            return None
+            return 1
+        fvs_dict = dict(fvs)
+        current_value = fvs_dict.get('si_settings_notification')
+        if not current_value:
+            return 1
+        try:
+            parts = current_value.split(':')
```
ymd-arista: Do you think adding a comment here that we are expecting something of the form `SI_SETTINGS_NOTIFIED:<N>` or `SI_SETTINGS_DEFAULT:<N>` would be helpful to understand why we are doing `int(parts[1])` later on?

ymd-arista:
```
  try:
      _, count = current_value.split(':', 1)
      return int(count) + 1
  except ValueError:
      return 1
```

How about doing something like this?

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
```diff
@@ -551,6 +551,7 @@ def get_speed_lane_count_and_subport(port, cfg_port_tbl):
     return port_speed, lane_count, subport_num
 
 
+
```
ymd-arista: Nit: Remove this newline

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -899,6 +957,19 @@ def handle_cmis_inserted_state(self, lport):
         media_lanes_mask = self.port_dict[lport]['media_lanes_mask']
         self.log_notice("{}: Setting media_lanemask=0x{:x}".format(lport, media_lanes_mask))
 
+        if port_info.get('notify_si_settings'):
```
longhuan-cisco: This seems to weaken the “xcvrd restart should not re-notify SI settings” contract for both CMIS and non-CMIS paths.

On xcvrd restart, SffManagerTask/CmisManagerTask start before SfpStateUpdateTask reposts TRANSCEIVER_INFO. That TRANSCEIVER_INFO SET is treated as an insertion-style event and marks notify_si_settings=True. For CMIS, this block can publish a fresh SI_SETTINGS_NOTIFIED:<N> before the later DP_PRE_INIT_CHECK path decides no CMIS application update is required. For non-CMIS, SffManagerTask similarly publishes SI settings on the replayed TRANSCEIVER_INFO event.

So even if the module is already configured, xcvrd restart can still re-notify SI settings and cause OA to reprogram SerDes/admin-bounce the port. Should the “already synced / restart-skip” decision happen before either notify path?

sonic-xcvrd/tests/test_xcvrd.py
```diff
@@ -2278,6 +2309,8 @@ def _check_notify_media_setting_with_gearbox(self, index, gearbox_line_lanes, sy
 
         found, result = app_port_tbl.get(logical_port_name)
         result_dict = dict(result) if result else None
+        if result_dict is not None:
+            result_dict.pop('si_settings_notification', None)
```
ymd-arista, line 2467: This logic should also be used in `test_notify_media_setting_mixed_settings` and `test_notify_media_setting_custom_only`. The failures in `sonic_platform_daemons` are because these are missing.

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
arpit-nexthop: @longhuan-cisco ACK, will address this

sonic-xcvrd/tests/test_xcvrd.py
arpit-nexthop: ACK

abdosi:
also depends on https://github.com/sonic-net/sonic-swss/pull/4497

abdosi:
@mlok-nokia : can you please help sign-off from Nokia.
@longhuan-cisco : please help sign-off from Cisco
@ymd-arista : please help sign-off from Arista.
@prgeor / @mihirpat1 : please sign-off from MSFT

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -139,6 +141,18 @@ def on_port_update_event(self, port_change_event):
...
+            # TRANSCEIVER_INFO_TABLE (active_apsel / lane counts) during datapath (re)init, and
+            # those self-writes come back as TRANSCEIVER_INFO SET events. Re-arming notify on them
+            # (and re-syncing SI) drives an endless re-notify / SI re-program / MAC-bounce loop.
+            # So arm only when SI is not yet synced for the current module AND no notification is
+            # already outstanding. The synced flag and the cached number are cleared on genuine
+            # removal (below), so a real new module still re-syncs exactly once.
+            if port_change_event.table_name == 'TRANSCEIVER_INFO' \
+                    and not self.port_dict[lport].get('si_settings_synced', False) \
+                    and self.port_dict[lport].get('si_notification_number') is None:
+                self.port_dict[lport]['notify_si_settings'] = True
```
anamehra, line 159: Is it the right thing to make the restart/insertion decision only from these local flags?

```py
# So arm only when SI is not yet synced for the current module AND no notification is
# already outstanding.
if port_change_event.table_name == 'TRANSCEIVER_INFO' \
        and not self.port_dict[lport].get('si_settings_synced', False) \
        and self.port_dict[lport].get('si_notification_number') is None:
    self.port_dict[lport]['notify_si_settings'] = True
```

After xcvrd restart both `si_settings_synced` and `si_notification_number` are process-local and start empty again. A replayed `TRANSCEIVER_INFO` can look like a fresh notify even if orchagent already had `SI_SYNC_DONE:<N>` in STATE_DB. The design matrix says xcvrd/pmon restart and warm reboot should not re-notify or flap the link: https://github.com/sonic-net/SONiC/blob/7fce6219a6de03d03143e4e1e29ce05b1a9da83a/doc/sfp-cmis/Interface-Link-bring-up-sequence.md#L517-L526. Can we key this off the retained DB ack/notification also, not just in-memory state?

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -900,6 +974,29 @@ def handle_cmis_inserted_state(self, lport):
...
+        # for this insertion. A host_tx_ready-driven reinit (orchagent brings the MAC up after
+        # applying SI) bounces the SM back to INSERTED; without this check we would re-notify
+        # and force an unnecessary re-program.
+        last_si_number = port_info.get('si_notification_number')
+        if last_si_number is not None and \
+                self.check_si_sync_done_match(lport, self.check_si_settings_app_status(lport), last_si_number):
+            self.port_dict[lport]['si_settings_synced'] = True
+            self.port_dict[lport]['notify_si_settings'] = False
+
+        if port_info.get('notify_si_settings') and not self.port_dict[lport].get('si_settings_synced', False):
```
anamehra, line 1006: This still looks unsafe for xcvrd restart / warm-restart. The new notify path is:

```py
if port_info.get('notify_si_settings') and not self.port_dict[lport].get('si_settings_synced', False):
    self.port_dict[lport]['notify_si_settings'] = False
    ...
    notification_number = media_settings_parser.notify_media_setting(...)
```

But the existing no-update path in the same file says:

```py
if not need_update:
    # As part of xcvrd restart, the TRANSCEIVER_INFO table is deleted and
    # created with default value of 'N/A' for all the active apsel fields.
    ...
    self.update_port_transceiver_status_table_sw_cmis_state(lport, CMIS_STATE_READY)
    return False
```

So after xcvrd restart we can re-notify because local state is gone, then still go READY because the CMIS application itself does not need update. I think READY here needs to be tied to the retained `SI_SYNC_DONE:<N>` check as well, otherwise this does not meet the no-re-notify/no-link-flap restart requirement in the design.

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -1058,6 +1155,142 @@ def handle_cmis_dp_deinit_state(self, lport):
...
+            # re-notify / re-enter SI_SETTINGS_WAIT.
+            self.port_dict[lport]['si_settings_synced'] = True
+            self.port_dict[lport].pop('si_notification_number', None)
+            self.update_port_transceiver_status_table_sw_cmis_state(lport, CMIS_STATE_DP_INIT)
+            return True
+        elif self.is_timer_expired(expired):
+            self.log_notice("{}: timeout waiting for SI settings to be applied (status: {}, expected: {})".format(
+                lport, si_settings_status, expected_notification_number))
+            self.port_dict[lport].pop('si_notification_number', None)
+            self.port_dict[lport]['notify_si_settings'] = True
```
anamehra, line 1309: The timeout path does not do what the design says. Code today only clears local state and retries:

```py
self.port_dict[lport].pop('si_notification_number', None)
self.port_dict[lport]['notify_si_settings'] = True
```

The design text says timeout should reset APPL_DB `si_settings_notification` to `SI_SETTINGS_DEFAULT:<N+1>`, then set `notify_si_settings=True` and reinit: https://github.com/sonic-net/SONiC/blob/7fce6219a6de03d03143e4e1e29ce05b1a9da83a/doc/sfp-cmis/Interface-Link-bring-up-sequence.md#L259-L260. I do not see that APPL_DB DEFAULT write here, so orchagent can keep seeing the old notification value.

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -193,6 +195,7 @@ def on_port_update_event(self, port_change_event):
 
             if self.XCVR_TYPE in port_change_event.port_dict:
                 self.port_dict[lport][self.XCVR_TYPE] = port_change_event.port_dict[self.XCVR_TYPE]
+                self.port_dict[lport]['notify_si_settings'] = True
```
anamehra, line 198: Same restart issue exists in the SFF path. This arms SI notify whenever `XCVR_TYPE` is present:

```py
if self.XCVR_TYPE in port_change_event.port_dict:
    self.port_dict[lport][self.XCVR_TYPE] = port_change_event.port_dict[self.XCVR_TYPE]
    self.port_dict[lport]['notify_si_settings'] = True
```

There is no retained DB check here before re-notifying. After xcvrd restart, replayed DB contents can hit this branch even when orchagent already synced the SI settings. That conflicts with the design restart language for xcvrd/pmon restart and warm reboot.

sonic-xcvrd/xcvrd/sff_mgr.py
```diff
@@ -86,7 +87,7 @@ class SffManagerTask(threading.Thread):
 
     SFF_LOGGER_PREFIX = "SFF-MAIN: "
 
-    def __init__(self, namespaces, main_thread_stop_event, platform_chassis, helper_logger):
+    def __init__(self, namespaces, port_mapping, main_thread_stop_event, platform_chassis, helper_logger):
```
anamehra, line 90: Please update the unit tests for this constructor change:

```py
def __init__(self, namespaces, port_mapping, main_thread_stop_event, platform_chassis, helper_logger):
```

The current tests still call the old signature, for example:

```py
sff_mgr = SffManagerTask(DEFAULT_NAMESPACE, stop_event, MagicMock(), helper_logger)
task = SffManagerTask(DEFAULT_NAMESPACE, stop_event, MagicMock(), helper_logger)
```

I found those in `sonic-xcvrd/tests/test_xcvrd.py` around the existing SFF tests. Without updating those call sites, this should fail before the new behavior is really tested.

anamehra (dismissed):

snider-nokia:
I agree with some of the outstanding code review comments.

Having said that, these changes are quite complex and timing jitter of critical involved operations between OA and Xcvrd (depending on a variety of factors) is likely to also expose edge conditions in this new logic.

If @arpit-nexthop is comfortable (including with the extant testing that's been done to verify typical operations) then I would suggest that these changes be merged to master alongside the requisite associated OA PR https://github.com/sonic-net/sonic-swss/pull/4497.  Then the community can begin ironing out those edge conditions/logic that remain and/or prove problematic.

We tried to cherry-pick these changes along with #4497 to a private dev branch to do some testing/verification, unfortunately we are getting local cherry-pick conflict when attempting to do so.  @arpit-nexthop, If you don't think these changes are entirely ready for merge to master then can you once more merge/rebase latest master to your branch so we can get a clean cherry-pick alongside latest master (to test/verify these changes in our lab)?

rlhui:
please resolve conflicts at earliest, htanks.

snider-nokia:
@arpit-nexthop, We have now merged this PR and its sister PR SWSS #4497 to master branch in our environment, and we are attempting to run OC tests.
First problem is that the Inband ports are not coming oper up because admin up notification is never sent to SAI.  Relevant log output is pasted below (see specifically timestamp 01:23:58.870282).   

Have you guys run OC tests on these changes?

I have posted same comment at PR SWSS #4497.



```26 Aug 10 01:14:44.138044 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0', 'SET', (('inband_type', 'port'),))'
2026 Aug 10 01:14:44.138168 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0|3333::3:1/128', 'SET', (('NULL', 'NULL'),))'
2026 Aug 10 01:22:20.722348 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: nlmsg type:16 key:Ethernet-IB0 admin:0 oper:0 addr:18:c3:00:8d:f1:ee ifindex:90 master:0 flags:4098
2026 Aug 10 01:22:20.722678 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: nlmsg type:16 key:Ethernet-IB0 admin:0 oper:0 addr:18:c3:00:8d:f1:ee ifindex:90 master:0 flags:4098
2026 Aug 10 01:22:21.901451 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0|3.3.3.1/32', 'SET', (('NULL', 'NULL'),))'
2026 Aug 10 01:22:21.901566 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0', 'SET', (('inband_type', 'port'),))'
2026 Aug 10 01:22:21.901609 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0|3333::3:1/128', 'SET', (('NULL', 'NULL'),))'
2026 Aug 10 01:22:29.926826 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0', 'SET', (('vrf', ''),))'
2026 Aug 10 01:23:15.040076 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: nlmsg type:17 key:Ethernet-IB0 admin:0 oper:0 addr:18:c3:00:8d:f1:ee ifindex:90 master:0 flags:4098
2026 Aug 10 01:23:50.980935 ixre-egl-board201 NOTICE macsec0#macsecmgrd: :- disableMACsec: The MACsec was not enabled on the port 'Ethernet-IB0'
2026 Aug 10 01:23:51.870541 ixre-egl-board201 NOTICE swss0#orchagent: :- setPortSiSettingsSyncStatus: Set port Ethernet-IB0 SI settings sync status to SI_SETTINGS_DEFAULT:0
2026 Aug 10 01:23:51.870546 ixre-egl-board201 NOTICE swss0#orchagent: :- initializePortSiSettingsSyncStatusBulk: Initialize si_settings_ack as SI_SETTINGS_DEFAULT:0 for port Ethernet-IB0
2026 Aug 10 01:23:51.871263 ixre-egl-board201 NOTICE pmon#ledd[37]: Received PORT table event: key=Ethernet-IB0, state=down
2026 Aug 10 01:23:51.872404 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: nlmsg type:16 key:Ethernet-IB0 admin:0 oper:0 addr:18:c3:00:8d:f1:ee ifindex:119 master:0 flags:4098
2026 Aug 10 01:23:51.872665 ixre-egl-board201 NOTICE swss0#orchagent: :- addHostIntfs: Create host interface for port Ethernet-IB0 with oper status down
2026 Aug 10 01:23:51.872718 ixre-egl-board201 NOTICE swss0#orchagent: :- publish: EVENT_PUBLISHED: {"sonic-events-swss:if-state":{"ifname":"Ethernet-IB0","status":"down","timestamp":"2026-08-10T01:23:51.872676Z"}}
2026 Aug 10 01:23:51.872747 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: Publish Ethernet-IB0(ok:down) to state db
2026 Aug 10 01:23:51.874415 ixre-egl-board201 NOTICE swss0#orchagent: :- initPortsBulk: Initialized port Ethernet-IB0
2026 Aug 10 01:23:51.871504 ixre-egl-board201 NOTICE pmon#ledd[37]: Received PORT table event: key=Ethernet-IB0, state=down
2026 Aug 10 01:23:51.880763 ixre-egl-board201 NOTICE swss0#orchagent: :- setRouterIntfsMtu: Update MTU to 9100 for all router interfaces mapped to port: Ethernet-IB0
2026 Aug 10 01:23:51.880770 ixre-egl-board201 NOTICE swss0#orchagent: :- doPortTask: Set port Ethernet-IB0 MTU to 9100
2026 Aug 10 01:23:51.880776 ixre-egl-board201 NOTICE swss0#orchagent: :- doPortTask: Port Ethernet-IB0: deferring admin-up until SI settings are notified
2026 Aug 10 01:23:51.880892 ixre-egl-board201 NOTICE swss0#orchagent: :- setHostTxReady: Setting host_tx_ready status = false, alias = Ethernet-IB0, port_id = 0x1000000000201
2026 Aug 10 01:23:51.881023 ixre-egl-board201 NOTICE swss0#orchagent: :- initHostTxReadyState: initialize host_tx_ready as false for port Ethernet-IB0
2026 Aug 10 01:23:51.977441 ixre-egl-board201 NOTICE swss0#orchagent: :- addSystemPorts: Added system port 5d00000000006a for ixre-egl-board202|asic0|Ethernet-IB0
2026 Aug 10 01:23:51.987263 ixre-egl-board201 NOTICE swss0#orchagent: :- addSystemPorts: Added system port 5d000000000036 for ixre-egl-board201|asic0|Ethernet-IB0
2026 Aug 10 01:23:52.889775 ixre-egl-board201 NOTICE swss0#portmgrd: :- doTask: Configure Ethernet-IB0 MTU to 9100
2026 Aug 10 01:23:52.893911 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: nlmsg type:16 key:Ethernet-IB0 admin:1 oper:0 addr:18:c3:00:8d:f1:ee ifindex:119 master:0 flags:4099
2026 Aug 10 01:23:52.894533 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: Publish Ethernet-IB0(ok:down) to state db
2026 Aug 10 01:23:52.895114 ixre-egl-board201 NOTICE swss0#portmgrd: :- doTask: Configure Ethernet-IB0 admin status to up
2026 Aug 10 01:23:53.934628 ixre-egl-board201 NOTICE swss0#orchagent: :- addRouterIntfs: Create router interface Ethernet-IB0 MTU 9100
2026 Aug 10 01:23:55.019746 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0|3.3.3.1/32', 'SET', (('state', 'ok'),))'
2026 Aug 10 01:23:55.023811 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0|3333::3:1/128', 'SET', (('state', 'ok'),))'
2026 Aug 10 01:23:58.813371 ixre-egl-board201 NOTICE swss0#orchagent: :- addInbandNeighbor: Created inband neighbor 18:c3:00:8d:f1:ee on Ethernet-IB0
2026 Aug 10 01:23:58.817592 ixre-egl-board201 NOTICE swss0#orchagent: :- addInbandNeighbor: Created inband neighbor 18:c3:00:8d:f1:ee on Ethernet-IB0
2026 Aug 10 01:23:58.870282 ixre-egl-board201 NOTICE swss0#orchagent: :- doPortTask: Port Ethernet-IB0: deferring admin-up until SI settings are notified
2026 Aug 10 01:23:58.872848 ixre-egl-board201 NOTICE pmon#ledd[37]: Received PORT table event: key=Ethernet-IB0, state=down
2026 Aug 10 01:23:58.872908 ixre-egl-board201 NOTICE pmon#ledd[37]: Received PORT table event: key=Ethernet-IB0, state=down
2026 Aug 10 01:24:07.032595 ixre-egl-board201 NOTICE swss1#orchagent: :- addSystemPorts: Added system port 15d00000000006a for ixre-egl-board202|asic0|Ethernet-IB0
2026 Aug 10 01:24:07.109942 ixre-egl-board201 NOTICE swss1#orchagent: :- addSystemPorts: Added system port 15d000000000055 for ixre-egl-board201|asic0|Ethernet-IB0
2026 Aug 10 01:24:09.151645 ixre-egl-board201 NOTICE swss1#orchagent: :- addRouterIntfs: Create router interface ixre-egl-board201|asic0|Ethernet-IB0 MTU 9100
2026 Aug 10 01:24:22.053109 ixre-egl-board201 NOTICE swss0#orchagent: :- addRouterIntfs: Create router interface ixre-egl-board202|asic0|Ethernet-IB0 MTU 9100
2026 Aug 10 01:24:22.054334 ixre-egl-board201 NOTICE swss1#orchagent: :- addRouterIntfs: Create router interface ixre-egl-board202|asic0|Ethernet-IB0 MTU 9100```

**Comment by arpit-nexthop** [comment]
> @arpit-nexthop, We have now merged this PR and its sister PR SWSS #4497 to master branch in our environment, and we are attempting to run OC tests. First problem is that the Inband ports are not coming oper up because admin up notification is never sent to SAI. Relevant log output is pasted below (see specifically timestamp 01:23:58.870282).
> 
> Have you guys run OC tests on these changes?
> 
> I have posted same comment at PR SWSS #4497.
> 
> ```
> 2026 Aug 10 01:14:44.138168 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0|3333::3:1/128', 'SET', (('NULL', 'NULL'),))'
> 2026 Aug 10 01:22:20.722348 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: nlmsg type:16 key:Ethernet-IB0 admin:0 oper:0 addr:18:c3:00:8d:f1:ee ifindex:90 master:0 flags:4098
> 2026 Aug 10 01:22:20.722678 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: nlmsg type:16 key:Ethernet-IB0 admin:0 oper:0 addr:18:c3:00:8d:f1:ee ifindex:90 master:0 flags:4098
> 2026 Aug 10 01:22:21.901451 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0|3.3.3.1/32', 'SET', (('NULL', 'NULL'),))'
> 2026 Aug 10 01:22:21.901566 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0', 'SET', (('inband_type', 'port'),))'
> 2026 Aug 10 01:22:21.901609 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0|3333::3:1/128', 'SET', (('NULL', 'NULL'),))'
> 2026 Aug 10 01:22:29.926826 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0', 'SET', (('vrf', ''),))'
> 2026 Aug 10 01:23:15.040076 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: nlmsg type:17 key:Ethernet-IB0 admin:0 oper:0 addr:18:c3:00:8d:f1:ee ifindex:90 master:0 flags:4098
> 2026 Aug 10 01:23:50.980935 ixre-egl-board201 NOTICE macsec0#macsecmgrd: :- disableMACsec: The MACsec was not enabled on the port 'Ethernet-IB0'
> 2026 Aug 10 01:23:51.870541 ixre-egl-board201 NOTICE swss0#orchagent: :- setPortSiSettingsSyncStatus: Set port Ethernet-IB0 SI settings sync status to SI_SETTINGS_DEFAULT:0
> 2026 Aug 10 01:23:51.870546 ixre-egl-board201 NOTICE swss0#orchagent: :- initializePortSiSettingsSyncStatusBulk: Initialize si_settings_ack as SI_SETTINGS_DEFAULT:0 for port Ethernet-IB0
> 2026 Aug 10 01:23:51.871263 ixre-egl-board201 NOTICE pmon#ledd[37]: Received PORT table event: key=Ethernet-IB0, state=down
> 2026 Aug 10 01:23:51.872404 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: nlmsg type:16 key:Ethernet-IB0 admin:0 oper:0 addr:18:c3:00:8d:f1:ee ifindex:119 master:0 flags:4098
> 2026 Aug 10 01:23:51.872665 ixre-egl-board201 NOTICE swss0#orchagent: :- addHostIntfs: Create host interface for port Ethernet-IB0 with oper status down
> 2026 Aug 10 01:23:51.872718 ixre-egl-board201 NOTICE swss0#orchagent: :- publish: EVENT_PUBLISHED: {"sonic-events-swss:if-state":{"ifname":"Ethernet-IB0","status":"down","timestamp":"2026-08-10T01:23:51.872676Z"}}
> 2026 Aug 10 01:23:51.872747 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: Publish Ethernet-IB0(ok:down) to state db
> 2026 Aug 10 01:23:51.874415 ixre-egl-board201 NOTICE swss0#orchagent: :- initPortsBulk: Initialized port Ethernet-IB0
> 2026 Aug 10 01:23:51.871504 ixre-egl-board201 NOTICE pmon#ledd[37]: Received PORT table event: key=Ethernet-IB0, state=down
> 2026 Aug 10 01:23:51.880763 ixre-egl-board201 NOTICE swss0#orchagent: :- setRouterIntfsMtu: Update MTU to 9100 for all router interfaces mapped to port: Ethernet-IB0
> 2026 Aug 10 01:23:51.880770 ixre-egl-board201 NOTICE swss0#orchagent: :- doPortTask: Set port Ethernet-IB0 MTU to 9100
> 2026 Aug 10 01:23:51.880776 ixre-egl-board201 NOTICE swss0#orchagent: :- doPortTask: Port Ethernet-IB0: deferring admin-up until SI settings are notified
> 2026 Aug 10 01:23:51.880892 ixre-egl-board201 NOTICE swss0#orchagent: :- setHostTxReady: Setting host_tx_ready status = false, alias = Ethernet-IB0, port_id = 0x1000000000201
> 2026 Aug 10 01:23:51.881023 ixre-egl-board201 NOTICE swss0#orchagent: :- initHostTxReadyState: initialize host_tx_ready as false for port Ethernet-IB0
> 2026 Aug 10 01:23:51.977441 ixre-egl-board201 NOTICE swss0#orchagent: :- addSystemPorts: Added system port 5d00000000006a for ixre-egl-board202|asic0|Ethernet-IB0
> 2026 Aug 10 01:23:51.987263 ixre-egl-board201 NOTICE swss0#orchagent: :- addSystemPorts: Added system port 5d000000000036 for ixre-egl-board201|asic0|Ethernet-IB0
> 2026 Aug 10 01:23:52.889775 ixre-egl-board201 NOTICE swss0#portmgrd: :- doTask: Configure Ethernet-IB0 MTU to 9100
> 2026 Aug 10 01:23:52.893911 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: nlmsg type:16 key:Ethernet-IB0 admin:1 oper:0 addr:18:c3:00:8d:f1:ee ifindex:119 master:0 flags:4099
> 2026 Aug 10 01:23:52.894533 ixre-egl-board201 NOTICE swss0#portsyncd: :- onMsg: Publish Ethernet-IB0(ok:down) to state db
> 2026 Aug 10 01:23:52.895114 ixre-egl-board201 NOTICE swss0#portmgrd: :- doTask: Configure Ethernet-IB0 admin status to up
> 2026 Aug 10 01:23:53.934628 ixre-egl-board201 NOTICE swss0#orchagent: :- addRouterIntfs: Create router interface Ethernet-IB0 MTU 9100
> 2026 Aug 10 01:23:55.019746 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0|3.3.3.1/32', 'SET', (('state', 'ok'),))'
> 2026 Aug 10 01:23:55.023811 ixre-egl-board201 DEBUG bgp0#bgpcfgd: Received message : '('Ethernet-IB0|3333::3:1/128', 'SET', (('state', 'ok'),))'
> 2026 Aug 10 01:23:58.813371 ixre-egl-board201 NOTICE swss0#orchagent: :- addInbandNeighbor: Created inband neighbor 18:c3:00:8d:f1:ee on Ethernet-IB0
> 2026 Aug 10 01:23:58.817592 ixre-egl-board201 NOTICE swss0#orchagent: :- addInbandNeighbor: Created inband neighbor 18:c3:00:8d:f1:ee on Ethernet-IB0
> 2026 Aug 10 01:23:58.870282 ixre-egl-board201 NOTICE swss0#orchagent: :- doPortTask: Port Ethernet-IB0: deferring admin-up until SI settings are notified
> 2026 Aug 10 01:23:58.872848 ixre-egl-board201 NOTICE pmon#ledd[37]: Received PORT table event: key=Ethernet-IB0, state=down
> 2026 Aug 10 01:23:58.872908 ixre-egl-board201 NOTICE pmon#ledd[37]: Received PORT table event: key=Ethernet-IB0, state=down
> 2026 Aug 10 01:24:07.032595 ixre-egl-board201 NOTICE swss1#orchagent: :- addSystemPorts: Added system port 15d00000000006a for ixre-egl-board202|asic0|Ethernet-IB0
> 2026 Aug 10 01:24:07.109942 ixre-egl-board201 NOTICE swss1#orchagent: :- addSystemPorts: Added system port 15d000000000055 for ixre-egl-board201|asic0|Ethernet-IB0
> 2026 Aug 10 01:24:09.151645 ixre-egl-board201 NOTICE swss1#orchagent: :- addRouterIntfs: Create router interface ixre-egl-board201|asic0|Ethernet-IB0 MTU 9100
> 2026 Aug 10 01:24:22.053109 ixre-egl-board201 NOTICE swss0#orchagent: :- addRouterIntfs: Create router interface ixre-egl-board202|asic0|Ethernet-IB0 MTU 9100
> 2026 Aug 10 01:24:22.054334 ixre-egl-board201 NOTICE swss1#orchagent: :- addRouterIntfs: Create router interface ixre-egl-board202|asic0|Ethernet-IB0 MTU 9100``` 
> ```

@snider-nokia 
I have been testing it, our internal branches are not exactly 1:1 with master, I will check if some patch got missed.

File: sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py (hunk 12 above)
Comment (arpit-nexthop) [reply sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py:169]: The decision no longer relies only on the process-local flags. Added XcvrTableHelper.is_si_settings_synced() which reconciles the retained APPL_DB si_settings_notification with the STATE_DB si_settings_ack

File: sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py (hunk 13 above)
Comment (arpit-nexthop) [reply sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py:1016]: The decision no longer relies only on the process-local flags. Added XcvrTableHelper.is_si_settings_synced() which reconciles the retained APPL_DB si_settings_notification with the STATE_DB si_settings_ack

File: sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py [hunk 17]
```diff
@@ -1049,6 +1120,139 @@ def handle_cmis_dp_deinit_state(self, lport):
...
+        if not api.scs_apply_datapath_init(host_lanes_mask):
+            self.log_notice("{}: unable to set application and stage DP init".format(lport))
+            self.force_cmis_reinit(lport, retries + 1)
+            return False
+
+        # Check if SI settings notification number was cached during INSERTED state
+        notification_number = self.port_dict[lport].get('si_notification_number')
+
+        if notification_number is not None:
+            self.log_notice("{}: SI settings notified to OA with number {}, waiting for ASIC to apply".format(
```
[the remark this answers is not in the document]
arpit-nexthop: The cached si_notification_number is now authoritative at AP_CONF because it's recovered earlier: a restart always re-enters CMIS_STATE_INSERTED (via force_cmis_reinit), and AP_CONF is strictly downstream of INSERTED.

sonic-xcvrd/xcvrd/sff_mgr.py
arpit-nexthop: SFF publish guarded by is_si_settings_synced(); replayed TRANSCEIVER_INFO won't re-notify a synced module.

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
arpit-nexthop: Added the DEFAULT:<N+1> write via shared reset_si_settings_notification_to_default().

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
arpit-nexthop: Done

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -567,8 +564,6 @@ def task_worker(self, stopping_event, sfp_error_event):
                                 if rc != SFP_EEPROM_NOT_READY:
                                     self.dom_db_utils.post_port_dom_thresholds_to_db(logical_port)
                                     self.vdm_db_utils.post_port_vdm_thresholds_to_db(logical_port)
```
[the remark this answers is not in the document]
arpit-nexthop: Added notify_media_setting_on_insert() at all insert sites; publishes for non-CMIS independent of --enable_sff_mgr, no-ops for CMIS/synced.

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
arpit-nexthop: Done

sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py
arpit-nexthop: Return value is consumed by handle_cmis_inserted_state, keeping this form so early returns still give None

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvr_table_helper.py
arpit-nexthop: Done

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvr_table_helper.py
arpit-nexthop: Added DEFAULT_NOTIFICATION_NUMBER = 1.

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvr_table_helper.py
arpit-nexthop: Added format comment; parse via split(':',1) + try/except ValueError.

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
arpit-nexthop: Revisited the restart logic

sonic-xcvrd/tests/test_xcvrd.py
arpit-nexthop: Applied to test_notify_media_setting_mixed_settings and _custom_only.

rlhui:
@mihirpat1 - please re-review if all comments are addressed?

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvr_table_helper.py
```diff
@@ -177,74 +182,99 @@ def get_firmware_info_tbl(self, asic_id):
...
+        if notified is None:
+            return False
+        found, ack = self.state_port_tbl[asic_id].hget(port_name, 'si_settings_ack')
+        if not found or not ack:
+            return False
+        try:
+            prefix, count = ack.split(':', 1)
+            # OA acks either SI_SYNC_DONE:<N> (real SI applied) or SI_SETTINGS_DEFAULT:<N>
+            # (defaults applied); both mean notification <N> has been fully processed.
+            return prefix in ('SI_SYNC_DONE', 'SI_SETTINGS_DEFAULT') and int(count) == notified
```
longhuan-cisco: **High: `SI_SETTINGS_DEFAULT:<N>` is not proof that SI settings for the current module were applied**

`get_current_si_notification_number()` discards the APPL_DB notification type and returns only `N`; this function then accepts either `SI_SYNC_DONE:N` or `SI_SETTINGS_DEFAULT:N` as synchronized. Those states have different meanings. `DONE:N` means OA applied a `NOTIFIED:N` request. `DEFAULT:N` is the removal/reset handshake; OA's DEFAULT path does not program new-module SerDes settings, it only echoes the DEFAULT ack.

A normal OIR sequence can therefore produce a false positive:

```text
old module removed
APPL_DB  = SI_SETTINGS_DEFAULT:8
STATE_DB = SI_SETTINGS_DEFAULT:8
new module inserted
is_si_settings_synced() returns True from matching number 8
SFF/direct path skips notification, or CMIS marks SI synced
new module's SI settings are never published/applied
```

There is a second ordering failure: if insertion observes `DEFAULT:8` before OA has acknowledged it, CMIS treats the bare number as an outstanding SI notification and waits for `SI_SYNC_DONE:8`; OA will only answer `SI_SETTINGS_DEFAULT:8`, causing timeout/reinit instead of a fresh notification.

Shall we preserve and validate both type and counter? Restart-skip should require the exact pair `APPL_DB = SI_SETTINGS_NOTIFIED:N` and `STATE_DB = SI_SYNC_DONE:N`. Only `NOTIFIED:N` should be considered outstanding. `DEFAULT:N` must force a fresh SI decision on the next insertion, while `UNAVAIL` remains a separate terminal no-settings result. The tests should reject `DEFAULT:N` + `DEFAULT:N` as applied SI and cover removal followed immediately by insertion.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -870,14 +869,50 @@ def retry_eeprom_reading(self):
...
+        physical_port = physical_port_list[0]
+        if not common._wrapper_get_presence(physical_port):
+            return
+        try:
+            api = platform_chassis.get_sfp(physical_port).get_xcvr_api()
+        except (NotImplementedError, AttributeError):
+            api = None
+        # CMIS modules are handled by CmisManagerTask, and SFF modules by SffManagerTask
+        # when it is enabled; only publish here for a non-CMIS module with the SFF manager off.
+        if common.is_cmis_api(api) or self.enable_sff_mgr:
```
longhuan-cisco: **High regression: the new SI publisher split leaves valid CMIS and retry paths without a terminal notification**

This is broader than the non-CMIS ownership issue discussed in https://github.com/sonic-net/sonic-platform-daemons/pull/799#discussion_r3184835560. The added fallback covers non-CMIS modules when SFF manager is disabled, but the implementation still has paths for which no component owns SI publication.

**Required invariant**

For every present external port whose admin-up can be gated by OA, the insertion flow needs one responsible owner that eventually publishes one of these terminal results:

- `SI_SETTINGS_NOTIFIED:N` when settings must be applied.
- `SI_SETTINGS_UNAVAIL` when there are no applicable/matching settings.

The current ownership split violates that invariant in several distinct cases:

- **Flat-memory/non-paged CMIS:** the direct `SfpStateUpdateTask` fallback returns for every `common.is_cmis_api(api)`; `SffManagerTask` also skips every CMIS API; and `CmisManagerTask` sets flat-memory CMIS to READY and returns. Therefore none of the three paths publishes `NOTIFIED:N` or `UNAVAIL`. Flat-memory CMIS, including passive-copper CMIS, can still require platform-specific NPU SI based on cable length, lane speed, board trace, and ASIC.
- **CMIS with `--skip_cmis_mgr`:** the CMIS task does not exist, but the always-on fallback still skips all CMIS APIs. This is a supported configuration, not an error path, and it has no SI publisher.
- **Transient publisher-side information-read failure:** if the second `get_transceiver_info()` call below returns `None`, this path returns without adding the port back to the EEPROM retry set. The SFF and CMIS paths can similarly consume/clear their one-shot `notify_si_settings` flag before the publisher read succeeds. The initial EEPROM-ready retry does not guarantee a retry of this later read, so a temporary failure can become a permanent missing notification.

This is a behavioral regression. Before this PR, the always-on `SfpStateUpdateTask` directly called `media_settings_parser.notify_media_setting()` after successful normal EEPROM reads, including for flat-memory CMIS. The HLD defines a CMIS-state-machine-driven module as CMIS **and not flat-memory**; using `is_cmis_api()` as the ownership boundary is therefore too broad.

The resulting failure is:

```text
external port is configured admin-up
        -> OA defers NPU admin-up while waiting for SI
        -> selected xcvrd path publishes neither NOTIFIED:N nor UNAVAIL
        -> OA never releases the gate
        -> the configured-up port remains down indefinitely
```

Shall we assign an explicit owner to every path: paged CMIS with the manager enabled can remain owned by `CmisManagerTask`; flat-memory CMIS and CMIS when the manager is intentionally skipped need the always-on/direct publisher. A successful decision must produce `NOTIFIED:N` or `UNAVAIL`, while transient read failures must rearm/retry rather than consume the event. Please also ensure paged CMIS has only one publisher.

Tests should cover flat-memory CMIS copper, flat-memory CMIS with no matching media entry, `--skip_cmis_mgr`, transient information-read failure followed by recovery, and no double-publication for paged CMIS.

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvr_table_helper.py
```diff
@@ -46,18 +46,22 @@
...
 class XcvrTableHelper:
     def __init__(self, namespaces):
         self.int_tbl, self.dom_tbl, self.dom_threshold_tbl, self.status_tbl, self.app_port_tbl, \
 		self.cfg_port_tbl, self.state_port_tbl, self.pm_tbl, self.firmware_info_tbl = {}, {}, {}, {}, {}, {}, {}, {}, {}
         self.state_db = {}
         self.appl_db = {}
+        self.app_port_read_tbl = {}
+        # (asic_id, port_name) -> last SI notification number issued this process
+        # (keeps get_next_si_notification_number monotonic despite the PORT_TABLE read lag).
+        self.last_si_notification_number = {}
```
longhuan-cisco, line 64: **Medium: SI notification-number allocation is not serialized across xcvrd task owners**

`N` is effectively a transaction ID. It lets OA say, "I processed exactly notification N," rather than acknowledging an older optic's settings.

Several threads can allocate it:

| Task | Notification |
|---|---|
| `SfpStateUpdateTask` | `DEFAULT:N` on removal |
| `CmisManagerTask` | `NOTIFIED:N` on insertion; `DEFAULT:N` on timeout |
| `SffManagerTask` | `NOTIFIED:N` on insertion |

Each task constructs a separate `XcvrTableHelper`, with its own `last_si_notification_number` cache. Allocation performs:

```text
read current N from APPL_DB
next = max(database N, this helper's local N) + 1
update this helper's local cache
later publish the notification
```

Normal sequential processing is safe because the APPL_DB write is synchronous:

```text
SfpStateUpdateTask processes removal and publishes DEFAULT:10
          |
          v
CmisManagerTask processes a later insertion and reads 10
          |
          v
publishes NOTIFIED:11
```

The problem occurs when processing for the same physical module overlaps. Here module A is current when CMIS processing begins and becomes the old/removed optic during the race:

```text
1. Module A is inserted.
2. SfpStateUpdateTask processes SFP_STATUS_INSERTED and publishes A's TRANSCEIVER_INFO.
3. CmisManagerTask observes that update and starts processing CMIS_STATE_INSERTED for A.
4. CmisManagerTask reads APPL_DB N=10, allocates N=11, and prepares A's SI payload.
5. Module A is quickly removed before that payload is published.
6. SfpStateUpdateTask processes SFP_STATUS_REMOVED, also reads N=10 through its
   separate helper, allocates N=11, and publishes DEFAULT:11.
7. CmisManagerTask's unfinished CMIS_STATE_INSERTED work for A resumes and publishes
   A's stale SI payload as NOTIFIED:11.
8. Module B may subsequently be inserted into the same port.
```

The overlapping allocation looks like this:

```text
APPL_DB currently has N=10

CmisManagerTask processing A's         SfpStateUpdateTask processing A's
CMIS_STATE_INSERTED                    later SFP_STATUS_REMOVED event
--------------------------------       ----------------------------------
reads N=10
allocates N=11
prepares A's SI payload
                                       reads N=10
                                       allocates N=11
                                       publishes DEFAULT:11
publishes stale NOTIFIED:11
                                       |
                                       v
                            Final APPL_DB = NOTIFIED:11
```

OA can then apply tuning captured for removed module A and acknowledge `DONE:11`. On module B's insertion or an xcvrd restart, retained `APPL_DB = NOTIFIED:11` plus `STATE_DB = DONE:11` looks perfectly synchronized, even though it belongs to the wrong module generation. This remains possible even after fixing the separate `DEFAULT:N` matching issue.

Could SI publication be made single-owner per port, or protected by a shared per-port lock/generation mechanism covering presence revalidation, number allocation, and the APPL_DB write? Atomic `N` allocation alone would guarantee unique numbers but would not stop stale work for module A from publishing after its removal. A barrier-based test that forces `CMIS_STATE_INSERTED` notification to overlap `SFP_STATUS_REMOVED` would make this behavior deterministic.

judyjoseph:
@anamehra @snider-nokia @mihirpat1 to review again + @arpit-nexthop can you get the sonic-mgmt testrun and share results PR description

arpit-nexthop:
## Manual tests
| Test | Result |
|---|---|
| **SI handshake integrity** | ✅ 56/56 `NOTIFIED`→`SYNC_DONE`, 0 stranded, 0 counter mismatches, serdes programmed (gearbox + ASIC), recirc port correctly not gated (IB fix) |
| **Clean boot** | ✅ all ports come up after the ~handshake completes |
| **swss restart** | ✅ full port recovery in 44s, 0 stranded |
| **admin toggle** | ✅ port recovers, re-syncs (`NOTIFIED:1`→`SYNC_DONE:1`), serdes re-applied |
| **config reload** | ✅ ports healthy after reload |
| **warm reboot** | ✅ **no port flap** (`flap_sum` + `last_up_time` unchanged, data plane preserved) |

## Sonic mgmt tests summary
| Area | pass | flaky | real fail | Failure notes |
|---|--:|--:|--:|---|
| `tests/platform_tests` | 34 | 1 | 1 | Matches our baseline |
| `tests/arp` | 10 | 0 | 0 |  |
| `tests/route` | 8 | 0 | 1 | `test_route_flap.py` — known issue |
| `tests/fdb` | 5 | 0 | 0 |  |
| `tests/ipfwd` | 4 | 0 | 0 |  |
| `tests/pc` | 3 | 1 | 0 | `test_po_cleanup.py` — flaky (passed on rerun) |
| `tests/sub_port_interfaces` | 3 | 0 | 0 |  |
| `tests/test_interfaces.py` | 1 | 0 | 0 |  |
| **Total (unique tests)** | **68** | 2 | 2 | |

## Passing tests
- `tests/arp/test_arpall.py`
- `tests/arp/test_arp_extended.py`
- `tests/arp/test_arp_update.py`
- `tests/arp/test_neighbor_mac_aging.py`
- `tests/arp/test_neighbor_mac_noptf.py`
- `tests/arp/test_neighbor_mac.py`
- `tests/arp/test_stress_arp.py`
- `tests/arp/test_tagged_arp.py`
- `tests/arp/test_unknown_mac.py`
- `tests/arp/test_wr_arp.py`
- `tests/fdb/test_fdb_churn.py`
- `tests/fdb/test_fdb_flush.py`
- `tests/fdb/test_fdb_mac_learning.py`
- `tests/fdb/test_fdb_mac_move.py`
- `tests/fdb/test_fdb.py`
- `tests/ipfwd/test_dip_sip.py`
- `tests/ipfwd/test_dir_bcast.py`
- `tests/ipfwd/test_mtu.py`
- `tests/ipfwd/test_nhop_group.py`
- `tests/pc/test_lag_2.py`
- `tests/pc/test_lag_member.py`
- `tests/pc/test_po_cleanup.py`
- `tests/pc/test_retry_count.py`
- `tests/platform_tests/cisco/test_fwutil_cisco.py`
- `tests/platform_tests/cli/test_show_platform.py`
- `tests/platform_tests/counterpoll/test_counterpoll_watermark.py`
- `tests/platform_tests/daemon/test_fancontrol.py`
- `tests/platform_tests/daemon/test_ledd.py`
- `tests/platform_tests/daemon/test_pcied.py`
- `tests/platform_tests/daemon/test_psud.py`
- `tests/platform_tests/daemon/test_syseepromd.py`
- `tests/platform_tests/fwutil/test_fwutil.py`
- `tests/platform_tests/link_flap/test_cont_link_flap.py`
- `tests/platform_tests/link_flap/test_link_flap.py`
- `tests/platform_tests/mgmt/test_mgmt_flap.py`
- `tests/platform_tests/sfp/test_sfpshow.py`
- `tests/platform_tests/sfp/test_sfputil.py`
- `tests/platform_tests/sfp/test_show_intf_xcvr.py`
- `tests/platform_tests/test_auto_negotiation.py`
- `tests/platform_tests/test_chassis_reboot.py`
- `tests/platform_tests/test_cpu_memory_usage.py`
- `tests/platform_tests/test_first_time_boot_password_change/test_password_change.py`
- `tests/platform_tests/test_hw_watchdog.py`
- `tests/platform_tests/test_intf_fec.py`
- `tests/platform_tests/test_kdump.py`
- `tests/platform_tests/test_link_down.py`
- `tests/platform_tests/test_liquid_cooling_leakage_detection.py`
- `tests/platform_tests/test_memory_exhaustion.py`
- `tests/platform_tests/test_platform_info.py`
- `tests/platform_tests/test_pmon_grep_safe.py`
- `tests/platform_tests/test_port_toggle.py`
- `tests/platform_tests/test_reload_config.py`
- `tests/platform_tests/test_sai_ocp_version.py`
- `tests/platform_tests/test_sensors.py`
- `tests/platform_tests/test_sequential_restart.py`
- `tests/platform_tests/test_xcvr_info_in_db.py`
- `tests/route/test_default_route.py`
- `tests/route/test_duplicate_route.py`
- `tests/route/test_forced_mgmt_route.py`
- `tests/route/test_route_consistency.py`
- `tests/route/test_route_flow_counter.py`
- `tests/route/test_route_map_check.py`
- `tests/route/test_route_perf.py`
- `tests/route/test_route_programming_benchmark.py`
- `tests/sub_port_interfaces/test_show_subinterface.py`
- `tests/sub_port_interfaces/test_sub_port_interfaces.py`
- `tests/sub_port_interfaces/test_sub_port_l2_forwarding.py`
- `tests/test_interfaces.py`


@snider-nokia @mihirpat1 Results match our baseline for the tests run. Please help with final review and approval.

snider-nokia:
Now that battery of OC tests have run well, suggest merging these changes (along with sister OA PR https://github.com/sonic-net/sonic-swss/pull/4497) to master so that they can be easily integrated/pulled into various vendors' lab testing environments for soak.

arpit-nexthop:
@mihirpat1 @prgeor 
Can we get these changes merged in master so that they can be soaked and tested in master?

@snider-nokia please mark both the PRs for approval from your end.

snider-nokia (approved):
Now with battery of OC tests passing successfully, suggest merging these changes to master so that multiple vendors can integrate/pull easily into their lab environments for soak.

ymd-arista (approved):

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
anamehra: Thanks for adding the retained DB check. One follow-up to validate: when an outstanding notification is recovered with `self.port_dict[lport]["si_notification_number"] = outstanding`, the `if not need_update` path can still set the port to `CMIS_STATE_READY` before orchagent writes the matching `SI_SYNC_DONE:<N>`.

Should this case remain in `SI_SETTINGS_WAIT` while `si_notification_number` is present and `si_settings_synced` is false? This looks like the remaining part of the earlier restart concern.

anamehra (approved):

abdosi:
@mihirpat1 : can you approve this.

sonic-xcvrd/xcvrd/xcvrd.py
longhuan-cisco: **P1 residual: recoverable publisher-side read failures can still leave the OA gate unreleased**

Thanks, the latest update fixes the flat-memory CMIS and `--skip_cmis_mgr` ownership gaps, and it attempts to re-arm the direct publisher after a transient read failure. Two recoverable read-failure paths still violate the terminal-notification invariant:

- **SFF manager:** `notify_si_settings` is cleared before the publisher-side `get_transceiver_info()` call. If that call returns `None`, the code only logs the failure; no retry is armed and no later event is guaranteed to restore the flag.
- **Direct retry path:** `notify_media_setting_on_insert()` re-adds the port to `retry_eeprom_set` when its information read fails, but `retry_eeprom_reading()` then unconditionally adds the port to `retry_success_set` and executes `self.retry_eeprom_set -= retry_success_set`. The attempted re-arm is therefore cancelled.

The direct retry sequence is:

```text
port is in retry_eeprom_set
        |
post_port_sfp_info_to_db() succeeds
        |
notify_media_setting_on_insert() performs another information read
        |
that read fails and re-adds the port to retry_eeprom_set
        |
caller adds the port to retry_success_set
        |
set subtraction removes it again
        |
no future retry and no NOTIFIED:N / UNAVAIL
```

These are supported transient EEPROM/I2C failure paths, not missing-platform-API cases. Before this PR, the successful first read's `transceiver_dict` was passed directly to `notify_media_setting()`; the new owner split adds another read while OA now waits fail-closed for a terminal notification. During initial bring-up or logical-port addition, this can leave a configured-up port down until xcvrd is restarted or another insertion event occurs.

Could the publisher reuse the already successful information read, or return an explicit terminal/retry outcome so the retry entry is removed only after `NOTIFIED:N` or `UNAVAIL` is published? For the SFF path, the notify intent should remain armed on a transient failure. An integrated test should exercise the outer retry loop; the current helper test verifies the re-add but does not catch its immediate removal.

sonic-xcvrd/xcvrd/xcvrd_utilities/xcvr_table_helper.py
longhuan-cisco: **P2 residual: insertion can accept stale synchronized state while the removal reset is still queued**

Thanks, the exact `NOTIFIED:N` plus `SI_SYNC_DONE:N` validation fixes the original issue once `DEFAULT:N` has reached the canonical APPL_DB table. One lower-probability ordering remains because `ProducerStateTable` writes are not immediately visible through the plain `Table` used by `is_si_settings_synced()`:

```text
visible APPL_DB = NOTIFIED:10
visible STATE_DB = DONE:10
        |
physical removal queues DEFAULT:11 in the producer-state table
        |
OA is delayed and has not consumed that update
        |
physical insertion completes the 2-second insertion soak
        |
is_si_settings_synced() still reads visible NOTIFIED:10 / DONE:10
        |
new-module SI notification is skipped
        |
OA later consumes DEFAULT:11
```

At that point the reset is visible, but the insertion event has already been consumed and no later event is guaranteed to force a fresh SI decision. The new module can therefore miss its SI programming even though `DEFAULT` itself is no longer accepted as proof of synchronization.

This requires OA consumer lag to span the insertion soak, so it is unlikely during normal healthy OIR and I would treat it as P2 rather than the original High issue. Standard DPB is not the main concern because its workflow waits for old ASIC ports to be deleted before adding the new ports. The window is physical OIR while OA is stalled or significantly backlogged.

Could an observed removal invalidate the retained synchronization by module generation, so the following insertion cannot accept the old pair regardless of producer/consumer visibility? A test that pauses OA consumption across removal and insertion would make the ordering deterministic.

longhuan-cisco:
Thanks @arpit-nexthop for addressing several major concerns. For convenience, I’ve collected the remaining items from the earlier review threads, in case any were overlooked.
- P1: Recoverable SI-publisher read failures can still consume the only publication attempt, leaving OA’s deferred admin-up unreleased. [Discussion](https://github.com/sonic-net/sonic-platform-daemons/pull/799#discussion_r3823509404)
- P2: A physical reinsertion can accept the previous module’s synchronized pair while the removal reset is still pending OA consumption. [Discussion](https://github.com/sonic-net/sonic-platform-daemons/pull/799#discussion_r3823509393)
- P2: In-flight SI work for a removed module can publish after the removal reset and be mistaken as synchronized for a later module. [Discussion](https://github.com/sonic-net/sonic-platform-daemons/pull/799#discussion_r3824745539)

These require uncommon error paths or narrow timing windows, but once triggered, the current module can miss SI publication/application until another OIR or xcvrd restart starts a fresh synchronization cycle.

PR #801, opened 2026-04-21
Replace imp with importlib

Pruthvi-Gowni:
#### Description
Replaced the deprecated imp.load_source function with the modern importlib.util and importlib.machinery modules across test files. A custom load_source function was implemented using importlib.machinery.SourceFileLoader and importlib.util.spec_from_file_location to maintain backward compatibility with the existing test code structure

#### Motivation and Context
The imp module was deprecated in Python 3.4 and completely removed in Python 3.12. Debian Trixie (the target distribution for the PMON container upgrade) uses Python 3.12 or later, which no longer supports the imp module. This change is required to ensure the test suite runs successfully on the upgraded PMON container.

The changes modify the test infrastructure itself. To verify these changes:

- Run the test suite for each affected daemon component (chassisd, ledd, pcied, psud, sensormond, stormond, syseepromd, thermalctld)
- Ensure all tests pass with the new importlib-based implementation
- Verify that the custom load_source function correctly handles module loading and caching via sys.modules

### DIFF
--- a/sonic-ledd/setup.py
+++ b/sonic-ledd/setup.py
@@ -21,6 +21,13 @@ setup(
         'pytest',
         'pytest-cov'
     ],
+    extras_require={
+        'testing': [
+            'mock>=2.0.0; python_version < "3.3"',
+            'pytest',
+            'pytest-cov'
+        ]
+    },
     classifiers=[
         'Development Status :: 4 - Beta',
         'Environment :: No Input/Output (Daemon)',
--- a/sonic-ledd/setup.py
+++ b/sonic-ledd/setup.py
@@ -17,13 +17,11 @@ setup(
         'wheel'
     ],
     tests_require=[
-        'mock>=2.0.0; python_version < "3.3"',
         'pytest',
         'pytest-cov'
     ],
     extras_require={
         'testing': [
-            'mock>=2.0.0; python_version < "3.3"',
             'pytest',
             'pytest-cov'
         ]
--- a/sonic-chassisd/tests/conftest.py
+++ b/sonic-chassisd/tests/conftest.py
@@ -1,0 +1,29 @@
+import os
+import sys
+import importlib.util
+import importlib.machinery
+from mock import MagicMock
+
+tests_path = os.path.dirname(os.path.abspath(__file__))
+
+# Add mocked_libs path so that the file under test can load mocked modules from there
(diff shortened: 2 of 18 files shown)


tirupatihemanth:
@Pruthvi-Gowni @saiarcot895 

Both test_DaemonPcied.py and test_pcied.py call load_source("pcied", ...) which creates different module objects each time. When pytest imports both files, the second load_source call replaces sys.modules["pcied"] with a NEW module object. Then mock.patch("pcied.load_platform_pcieutil", ...) patches the NEW module, while DaemonPcied.__init__.__globals__ points to the OLD module's dict. So the mock doesn't affect the actual function that DaemonPcied.__init__ calls, causing it to run the real load_platform_pcieutil which returns None in the test environment (no platform paths available).

The fix: prevent double-loading by using conftest so we load the module only once across the tests for one daemon
We need to do this at least for all daemons where we have multiple test files.

Pruthvi-Gowni:
> @Pruthvi-Gowni @saiarcot895
> 
> Both test_DaemonPcied.py and test_pcied.py call load_source("pcied", ...) which creates different module objects each time. When pytest imports both files, the second load_source call replaces sys.modules["pcied"] with a NEW module object. Then mock.patch("pcied.load_platform_pcieutil", ...) patches the NEW module, while DaemonPcied.**init**.**globals** points to the OLD module's dict. So the mock doesn't affect the actual function that DaemonPcied.**init** calls, causing it to run the real load_platform_pcieutil which returns None in the test environment (no platform paths available).
> 
> The fix: prevent double-loading by using conftest so we load the module only once across the tests for one daemon We need to do this at least for all daemons where we have multiple test files.

Addressed

sonic-ledd/setup.py
```diff
@@ -21,6 +21,13 @@
         'pytest',
         'pytest-cov'
     ],
+    extras_require={
+        'testing': [
+            'mock>=2.0.0; python_version < "3.3"',
```
saiarcot895: Why is a dependency for Python versions less than 3.3 being added here?

Pruthvi-Gowni: I think it's not required. Reverted it

saiarcot895 (approved):

saiarcot895:
@tirupatihemanth can you review as well?

PR #802, opened 2026-04-24
[Multi-ASIC] Added fast-reboot support for multi-asic

YairRaviv:
#### Description

Add fast-reboot support for multi-asic devices, according to this HLD - [[Multi-ASIC] Added fast-reboot support for multi-asic
](https://github.com/sonic-net/SONiC/pull/2306)

#### Motivation and Context

The main changes in transceiver daemon are related to checking the fast-reboot state per ASIC instead of globally.

### DIFF
--- a/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
+++ b/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
@@ -62,6 +62,13 @@ class CmisManagerTask(threading.Thread):
         self.xcvr_table_helper = XcvrTableHelper(self.namespaces)
         # Cache of gearbox line lanes dict, refreshed once per task_worker iteration.
         self._gearbox_lanes_dict = None
+        self.fast_reboot_status = self.initialize_fast_reboot_status()
+
+    def initialize_fast_reboot_status(self):
+        fast_reboot_status = {}
+        for namespace in self.namespaces:
+            fast_reboot_status[namespace] = bool(common.is_fast_reboot_enabled(namespace))
+        return fast_reboot_status
 
     def log_debug(self, message):
         helper_logger.log_debug(message)
@@ -78,7 +85,7 @@ class CmisManagerTask(threading.Thread):
     def is_fast_reboot_enabled_for_lport(self, lport):
         asic_id = self.get_asic_id(lport)
         namespace = common.get_namespace_from_asic_id(asic_id) if asic_id >= 0 else ''
-        return bool(common.is_fast_reboot_enabled(namespace))
+        return self.fast_reboot_status.get(namespace, False)
 
     def update_port_transceiver_status_table_sw_cmis_state(self, lport, cmis_state_to_set):
         status_table = self.xcvr_table_helper.get_status_sw_tbl(self.get_asic_id(lport))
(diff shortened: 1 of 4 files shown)


sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -71,15 +70,14 @@ def log_notice(self, message):
...
-            self._is_fast_reboot_enabled = common.is_fast_reboot_enabled()
-        return self._is_fast_reboot_enabled
-
     def get_asic_id(self, lport):
         return self.port_dict.get(lport, {}).get("asic_id", -1)
 
+    def is_fast_reboot_enabled_for_lport(self, lport):
+        asic_id = self.get_asic_id(lport)
+        namespace = common.get_namespace_from_asic_id(asic_id) if asic_id >= 0 else ''
+        return bool(common.is_fast_reboot_enabled(namespace))
```
bgallagher-nexthop: The old code was caching the result of this check in-memory to avoid repeating the database read, maybe we could continue to memoize it?

```suggestion
import functools

    @functools.cache
    def is_fast_reboot_enabled_for_lport(self, lport):
        asic_id = self.get_asic_id(lport)
        namespace = common.get_namespace_from_asic_id(asic_id) if asic_id >= 0 else ''
        return bool(common.is_fast_reboot_enabled(namespace))
```

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -1052,11 +1064,8 @@ def init(self):
         # Initialize xcvr table helper
         self.xcvr_table_helper = XcvrTableHelper(self.namespaces)
 
-        if common.is_fast_reboot_enabled():
-            self.log_info("Skip loading media_settings.json and optics_si_settings.json in case of fast-reboot")
-        else:
-            media_settings_parser.load_media_settings()
-            optics_si_parser.load_optics_si_settings()
+        media_settings_parser.load_media_settings()
+        optics_si_parser.load_optics_si_settings()
```
bgallagher-nexthop, line 1072: Out of curiosity, why do we now load these settings even when fast reboot is enabled? I can see how skipping loading these files would help bootstrap `xcvrd` more quickly, since they can be relatively large JSON files

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -281,6 +281,16 @@ def __init__(self, namespaces, port_mapping, sfp_obj_dict, main_thread_stop_even
         self.dom_db_utils = DOMDBUtils(sfp_obj_dict, self.port_mapping, self.xcvr_table_helper, self.task_stopping_event, self.logger)
         self.vdm_db_utils = VDMDBUtils(sfp_obj_dict, self.port_mapping, self.xcvr_table_helper, self.task_stopping_event, self.logger)
 
+    def is_warm_fast_reboot_for_lport(self, logical_port):
```
bgallagher-nexthop, line 292: Like the above case in the state machine, it may be worth considering caching the result of this function in memory

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
YairRaviv: I changed the logic to check fast reboot state once for each ASIC namespace, instead of per port

sonic-xcvrd/xcvrd/xcvrd.py
YairRaviv: I changed the logic to check fast reboot state once for each ASIC namespace, instead of per port

sonic-xcvrd/xcvrd/xcvrd.py
YairRaviv: since there is a scenario in which 1 ASIC performs a cold reboot while the other ASICs perform a fast reboot.
So instead of completely skipping from loading these settings, we are now checking whether to notify media/si settings or not later in the flow based on the fast-reboot state of the relevant ASIC

PR #808, opened 2026-04-30
[thermalctld] Remove SFP temperature polling

vvolam:
#### Description
Remove SFP/transceiver temperature polling from thermalctld. SFP temperature reporting is owned by xcvrd and exposed via the `TRANSCEIVER_DOM_SENSOR`, `TANSCEIVER_DOM_TEMPERATURE` tables, so duplicating it in thermalctld is unnecessary.

Changes:
- Drop the chassis-level and per-module SFP enumeration loops in `TemperatureUpdater.update()`.
- Remove the now-unnecessary `'SFP' not in parent_name` guard around `PHYSICAL_ENTITY_INFO` updates in `_collect_thermals()`.
- Remove the SfpUtilHelper init, Redis-based SFP temperature/threshold readers, and `_collect_sfp_thermals()` helper.
- Update unit tests: drop SFP-specific tests and adjust the module thermal count expectation (no longer counts the SFP thermal).

Companion PR:
- sonic-net/sonic-utilities#4522 — `tempershow` reads SFP temperature directly from xcvrd-managed `TRANSCEIVER_DOM_TEMPERATURE` / `TRANSCEIVER_DOM_THRESHOLD` / `TRANSCEIVER_DOM_FLAG` tables instead of `TEMPERATURE_INFO`, so `show platform temperature` continues to display SFP modules after this change.

#### Motivation and Context
SFP/transceiver temperature was being read from Redis (`TRANSCEIVER_DOM_*` tables populated by xcvrd) inside thermalctld's update loop. On high-port-count platforms (e.g. 512 SFPs), this caused 3-5 Redis round-trips per SFP per cycle, which could delay fast-polling sensors (such as ASIC temperature) and added duplicate ownership of transceiver telemetry between xcvrd and thermalctld.

Other consumers of `TEMPERATURE_INFO` were reviewed and are unaffected:
- sonic-snmpagent `PhysicalSensorTableMIBUpdater` / `ThermalCacheUpdater` already filter out non-chassis thermals, and transceiver temperature is exposed via the dedicated `TRANSCEIVER_DOM_SENSOR` path (`update_xcvr_dom_data`).
- system-health `hardware_checker` only reads the `ASIC` entry.

#### How Has This Been Tested?
- `pytest sonic-thermalctld/tests/test_thermalctld.py` — all 89 tests pass.
[...]

PR #809, opened 2026-05-01
[thermalctld] Fix SFP DOM to use per-ASIC STATE_DB on multi-ASIC

vishal-nokia:
#### Description
thermalctld now reads SFP DOM tables (temperature, thresholds, sensor) from the
correct STATE_DB based on platform type: host namespace on single-ASIC (unchanged),
per-ASIC namespace on multi-ASIC. Table dictionaries are keyed by asic_id (integer),
matching the xcvrd/ycabled pattern.
#### Motivation and Context
thermalctld reads SFP temperature and thresholds from STATE_DB tables populated
by xcvrd. On multi-ASIC platforms these tables reside in per-ASIC namespace
databases. This change ensures thermalctld connects to the correct STATE_DB
for each ASIC, while preserving existing single-ASIC behavior.
Tested on single-asic and multi-asic platforms


vishal-nokia:
@mlok-nokia @judyjoseph for viz

vishal-nokia:
Closing this PR since it's no longer needed. 
#808 removed SFP temperature polling from thermalctld, and the companion [sonic-utilities#4522](https://github.com/sonic-net/sonic-utilities/pull/4522) now handles SFP temperature display in tempershow with multi-ASIC support.

PR #814, opened 2026-05-12
Removed unused is_warm_reboot_enabled utility

Ryangwaite:
#### Description

The `is_warm_reboot_enabled()` helper in sonic-platform-daemons/sonic-xcvrd/xcvrd/xcvrd_utilities/common.py is currently unused in production code. [PR #680](https://github.com/sonic-net/sonic-platform-daemons/pull/680) replaced its call sites with `is_syncd_warm_restore_complete()`, but the function is still defined and importable. It should be removed to prevent it being reintroduced as a footgun that risks causing data plane impact in warm-reboot.

The changes here remove the utility.

#### Motivation and Context

**Why this is a footgun**

`is_warm_reboot_enabled()` calls `WarmStart::checkWarmStart("xcvrd", "pmon", False)`, which has a critical side-effect -> on cold boot it writes `restore_count=0` into `STATE_DB:WARM_RESTART_TABLE|xcvrd`. That row is what makes the next warm reboot recognise xcvrd as a registered warm-restart participant.

After PR [PR #680](https://github.com/sonic-net/sonic-platform-daemons/pull/680), nothing in xcvrd calls this function on startup any more, so the row is no longer being seeded on cold boot. If a future PR re-introduces a call to `is_warm_reboot_enabled()`, it will:

   1. Run on a cold boot where the row was never seeded (because no other startup path seeds it now).
   2. Hit the `restore_count == ""``branch in `swss-common/common/warm_restart.cpp`.
   3. Log `"xcvrd doing warm start, but restore_count not found... fall back to cold start"`.
   4. Cache `m_enabled = false` and `m_systemWarmRebootEnabled = false` on the singleton.
   5. Cause xcvrd to **take the cold-start branch on what is actually a warm reboot — almost certainly a data-plane hit**.

CI unit tests

This is very-low risk to 202511 branch so no cherry-pick required

### DIFF
--- a/sonic-xcvrd/tests/test_xcvrd.py
+++ b/sonic-xcvrd/tests/test_xcvrd.py
@@ -6235,10 +6235,10 @@ class TestXcvrdScript(object):
     @patch('xcvrd.xcvrd.DaemonXcvrd.load_platform_util', MagicMock())
     @patch('xcvrd.xcvrd_utilities.port_event_helper.get_port_mapping', MagicMock(return_value=MockPortMapping))
     @patch('sonic_py_common.device_info.get_paths_to_platform_and_hwsku_dirs', MagicMock(return_value=('/tmp', '/tmp')))
-    @patch('xcvrd.xcvrd_utilities.common.is_warm_reboot_enabled', MagicMock(return_value=False))
     @patch('xcvrd.xcvrd.DaemonXcvrd.wait_for_port_config_done', MagicMock())
     @patch('xcvrd.xcvrd.DaemonXcvrd.initialize_sfp_obj_dict', MagicMock())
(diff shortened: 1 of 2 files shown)


rookie-who (approved):
LGTM for master. Clean removal of unused `is_warm_reboot_enabled()` — replaced by `is_syncd_warm_restore_complete()` in PR #680. Test mock update is correct (now patches the actual production code path).

**Cherry-pick warning:** Release branches 202411 and 202405 still have active call sites for `is_warm_reboot_enabled()` in `xcvrd.py` (lines ~1576 and ~2186). Do not cherry-pick this PR to those branches without also cherry-picking PR #680.

vaibhavhd (approved):

PR #815, opened 2026-05-14
[smartswitch][chassisd]: Add DPU recovery state machine and new DB fields

vvolam:
#### Description

Implement DPU failure detection and automated recovery state machine in
`chassisd` as per the [enhance-dpu-robustness HLD](https://github.com/sonic-net/SONiC/pull/2310).

Changes include:

- New CHASSIS_STATE_DB `DPU_STATE` fields: `ready_status`, `recovery_status`,
  `reset_count`, `last_down_time`, `last_ready_time`
- DPU recovery state machine with states: Booting, Ready, SWFailure, PowerCycle,
  ManualIntervention, Offline, Unrecoverable
- Configurable thresholds from `platform.json`: `dpu_reboot_timeout`, `dpu_reset_limit`
- `dpu-auto-recovery` feature flag check from CONFIG_DB FEATURE table
- Recovery initialization on chassisd startup with NPU kernel crash detection
- State tracking in main poll loop with per-DPU recovery actions
- Mark all DPUs not-ready on chassisd shutdown (deinit)

State machine behavior:
- **SWFailure**: Entered when control plane goes down with midplane still up;
  transient state before PowerCycle or ManualIntervention
- **PowerCycle**: Auto-recovery enabled — chassisd power-cycles the DPU
- **ManualIntervention**: Auto-recovery disabled — operator must recover manually
- **Unrecoverable**: `reset_count` reached `reset_limit` — no further auto-recovery

Depends on sonic-buildimage change: https://github.com/sonic-net/sonic-buildimage/pull/27385

#### Motivation and Context

SmartSwitch DPUs can fail due to software crashes, hardware faults, PCIe failures,
or NPU kernel crashes. Without automated recovery, failed DPUs require manual
operator intervention. This change adds automated failure detection and power-cycle
recovery to `chassisd`, reducing DPU downtime and operational burden.

- Unit tests (existing chassisd tests pass)
- Manual review of state machine transitions against HLD scenarios

 Unit Tests (46 tests passing)
- Location: `sonic-chassisd/tests/test_dpu_auto_recovery.py`
- Covers: state transitions, NPU crash detection, feature flag, reset limits, multi-DPU, full recovery cycles

 Integration Testbed Script
- Location: `sonic-chassisd/tests/testbed/test_dpu_auto_recovery.py`
- Usage: `sudo python3 test_dpu_auto_recovery.py --dpu DPU0 [--skip-destructive]`
- Non-destructive tests: DB field validation, chassisd running check, recovery_status validity
- Destructive tests: admin-down behavior, feature flag toggle, power-cycle recovery, chassisd restart/stop

Verified on the smartswitch testbed with the integrated testbed script

HLD: https://github.com/vvolam/SONiC/blob/b43d0412f933e0df4a2df06be5f6754f7666c715/doc/smart-switch/pmon/enhance-dpu-robustness.md


sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1183,279 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+        """Power-cycle a DPU and increment reset_count."""
+        recovery = self.dpu_recovery_state[module_name]
+        recovery['reset_count'] += 1
+        self._set_reset_count(module_name, recovery['reset_count'])
+        self.log_warning("{}: Initiating power-cycle (reset_count={})".format(
+            module_name, recovery['reset_count']))
+
+        try:
+            module = self.chassis.get_module(module_index)
+            try_get(module.set_admin_state, MODULE_ADMIN_DOWN, default=False)
```
gpunathilell, line 1269: @vvolam is there a scenario where this can collide with user issued reboot/admin state change?

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1183,279 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            if oper_status == str(ModuleBase.MODULE_STATUS_OFFLINE) and \
+                    current_state not in (DPU_STATE_BOOTING, DPU_STATE_POWER_CYCLE,
+                                          DPU_STATE_SW_FAILURE,
+                                          DPU_STATE_MANUAL_INTERVENTION):
+                self.log_warning(
+                    "{}: Hardware failure detected (oper_status: Offline)".format(name))
+                self._set_ready_status(name, 'false')
+                self._set_last_down_time(name)
+                if auto_recovery_enabled:
+                    self._power_cycle_dpu(name, module_index)
```
gpunathilell, line 1393: @vvolam same question as in HLD, Do we need to power off DPU, whihc is already powered off? and one more question, the autorecovery should also have a wait time, to check if the DPU recovers by itself (which is expected in case of kernel panic / mem exhaustion) before we try to power it on with this feature right?

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1183,279 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+                    else:
+                        recovery['state'] = DPU_STATE_MANUAL_INTERVENTION
+                elif dp_state == 'down':
+                    # Data plane down only: mark not ready but no recovery.
+                    self._set_ready_status(name, 'false')
+                    self._set_last_down_time(name)
+                else:
+                    # All states up — ensure ready_status reflects readiness
+                    # (covers recovery from transient dp-only down).
+                    self._set_ready_status(name, 'true')
```
gpunathilell: do we need to check current status to confirm if it s already set to true before writing every time? Because this is the steady state behavior

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1183,279 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+                continue
+
+            if current_state == DPU_STATE_SW_FAILURE:
```
gpunathilell: when is this value set?

sonic-chassisd/scripts/chassisd
vvolam: The `_is_planned_transition_in_progress()` check at line 1379 explicitly suppresses recovery during user-initiated operations. It reads `state_transition_in_progress` from STATE_DB and skips recovery processing when a planned shutdown/reboot is active.

sonic-chassisd/scripts/chassisd
vvolam: The DPU now gets a configurable wait period to self-recover before power-cycling. The state machine also only power-cycles when the DPU transitions to PowerCycle state after the grace period expires.

sonic-chassisd/scripts/chassisd
```diff
@@ -109,6 +109,47 @@ DP_STATE = 'dpu_data_plane_state'
...
+DEFAULT_DPU_SELF_RECOVERY_TIMEOUT = 300
+DEFAULT_DPU_RESET_LIMIT = 2
+MINIMUM_SELF_RECOVERY_GRACE_PERIOD = 30
+MINIMUM_SELF_RECOVERY_POLL_COUNT = 3
+
+# Reboot cause file used to detect NPU kernel crash on chassisd startup
+REBOOT_CAUSE_FILE = "/host/reboot-cause/reboot-cause.txt"
+NPU_CRASH_REBOOT_CAUSE_KEYWORDS = ("kernel panic", "unknown")
+
+# Planned operation tracking (STATE_DB CHASSIS_MODULE_TABLE)
```
judsonwilson-nvidia, line 142:
```suggestion
# Name of DB field for planned operation tracking (STATE_DB CHASSIS_MODULE_TABLE)
```

sonic-chassisd/scripts/chassisd
```diff
@@ -109,6 +109,47 @@ DP_STATE = 'dpu_data_plane_state'
...
+STATE_TRANSITION_IN_PROGRESS = 'state_transition_in_progress'
+
+# DPU recovery states
+DPU_STATE_BOOTING = 'Booting'
+DPU_STATE_READY = 'Ready'
+DPU_STATE_WAIT_FOR_SELF_RECOVERY = 'WaitForSelfRecovery'
+DPU_STATE_POWER_CYCLE = 'PowerCycle'
+DPU_STATE_MANUAL_INTERVENTION = 'ManualIntervention'
+DPU_STATE_OFFLINE = 'Offline'
+DPU_STATE_UNRECOVERABLE = 'Unrecoverable'
```
judsonwilson-nvidia, line 152: PlannedShutdown? PlannedReboot?  Those are in the HLD

sonic-chassisd/scripts/chassisd
```diff
@@ -711,6 +752,7 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
 
         self.hostname_table = swsscommon.Table(self.chassis_state_db, CHASSIS_MODULE_HOSTNAME_TABLE)
         self.module_reboot_table = swsscommon.Table(self.chassis_state_db, CHASSIS_MODULE_REBOOT_INFO_TABLE)
+        self.dpu_state_table = swsscommon.Table(self.chassis_state_db, 'DPU_STATE')
```
judsonwilson-nvidia, line 755: Nit: why not use a constant named DPU_STATE_TABLE for 'DPU_STATE' like is done for the other tables in the few lines above it, for consistency?

sonic-chassisd/scripts/chassisd
```diff
@@ -719,22 +761,47 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+        self.dpu_boot_timeout = DEFAULT_DPU_BOOT_TIMEOUT
+        self.dpu_self_recovery_timeout = DEFAULT_DPU_SELF_RECOVERY_TIMEOUT
+
         if os.path.isfile(PLATFORM_JSON_FILE):
             try:
                 with open(PLATFORM_JSON_FILE, 'r') as f:
                     platform_cfg = json.load(f)
-                # Extract the "dpu_reboot_timeout" if it exists
                 self.dpu_reboot_timeout = int(platform_cfg.get("dpu_reboot_timeout", DEFAULT_DPU_REBOOT_TIMEOUT))
+                self.reset_limit = int(platform_cfg.get("dpu_reset_limit", DEFAULT_DPU_RESET_LIMIT))
```
judsonwilson-nvidia, line 775: Nit: use constants for these field names? It looks like we do that for others...

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+        """Check if dpu-auto-recovery feature is enabled in CONFIG_DB FEATURE table.
+
+        Returns True only when ``state == 'enabled'``. Both ``'disabled'`` and
+        ``'always_disabled'`` suppress automatic power-cycle recovery; chassisd
+        still updates DPU state fields in CHASSIS_STATE_DB but will not
+        initiate a power-cycle.
+        """
+        try:
+            config_db = daemon_base.db_connect("CONFIG_DB")
+            feature_table = swsscommon.Table(config_db, 'FEATURE')
```
judsonwilson-nvidia, line 1205: I'm not that familiar with the db APIs, but should we be caching the feature table in `self` during the init, like we do for other tables? It looks like this is called every time the state machine is updated...

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            self.log_error("Failed to read dpu-auto-recovery feature state: {}".format(e))
+        return False  # Default to disabled
+
+    def _get_dpu_states(self, module_name):
+        """Get current DPU states from CHASSIS_STATE_DB."""
+        dpu_key = "DPU_STATE|" + module_name
+        mp_state = self.get_dpu_midplane_state(dpu_key)
+        _, cp_state = self.dpu_state_table.hget(module_name, CP_STATE)
+        _, dp_state = self.dpu_state_table.hget(module_name, DP_STATE)
+        return mp_state, cp_state, dp_state
```
judsonwilson-nvidia, line 1221: Store in a typing.NamedTuple to help not mix up which state is which?

```
from typing import NamedTuple

class DpuStates(NamedTuple):
    mp_state: str
    cp_state: str 
    dp_state:str


...
    return DpuStates(mp_state=mp_state, cp_state=cp_state, dp_state=dp_state)
```

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+        except Exception as e:
+            self.log_warning("{}: Failed to read state_transition_in_progress: {}".format(
+                module_name, e))
+        return False
+
+    def _power_cycle_dpu(self, module_name, module_index):
+        """Power-cycle a DPU and increment reset_count."""
+        recovery = self.dpu_recovery_state[module_name]
+        recovery['reset_count'] += 1
+        self._set_reset_count(module_name, recovery['reset_count'])
```
judsonwilson-nvidia, line 1263: Should we use the constant RESET_COUNT here?

sonic-chassisd/scripts/chassisd
vvolam: They are covered in HLD to have complete scenario. They were already part of code. I am not planning to refactor the code for now, but update new DB fields as part of the planned operations path too.

vvolam:
@judsonwilson-nvidia As they are all cosmetic comments, I will address them as a follow-up comments. Thank you for the detailed review.

@gpunathilell, @judsonwilson-nvidia, @dgsudharsan @rameshraghupathy We are planning to merge the changes. We could toggle the feature to disable/enable to disable/enable the code path.

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+        self.log_warning("{}: Initiating power-cycle (reset_count={})".format(
+            module_name, recovery['reset_count']))
+
+        try:
+            module = self.chassis.get_module(module_index)
+            try_get(module.set_admin_state, MODULE_ADMIN_DOWN, default=False)
```
judsonwilson-nvidia, line 1269: Would it make more sense to use ModuleHelper.reboot_module(), rather than set_admin_state MODULE_ADMIN_UP/DOWN ?   That is what the reboot_smartswitch_helper script calls.

Are you thinking the admin up/down has stronger power-cycling semantics?

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            try_get(module.set_admin_state, MODULE_ADMIN_UP, default=False)
+        except Exception as e:
+            self.log_error("{}: Power-cycle failed: {}".format(module_name, e))
+
+        if recovery['reset_count'] >= self.reset_limit:
```
judsonwilson-nvidia, line 1274: Shouldn't this check come before rebooting?

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            recovery['state'] = DPU_STATE_UNRECOVERABLE
+            self._set_recovery_status(module_name, RECOVERY_UNRECOVERABLE)
+            self.log_error("{}: Reached reset_limit ({}). Marking as unrecoverable.".format(
+                module_name, self.reset_limit))
+        else:
+            recovery['state'] = DPU_STATE_POWER_CYCLE
+            recovery['boot_start_time'] = time.time()
```
judsonwilson-nvidia, line 1281: Nit/future:

It looks like this is currently in one place, so maybe don't need to change it right now, but in a transition from state A to B, it's dangerous to put the setup code for B in the handler for A.  It would be better to package it with B, or at least create some helper functions for transitions.

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            self.dpu_recovery_state[name]['boot_start_time'] = time.time()
+
+            if not npu_crash:
+                continue
+
+            # NPU kernel crash recovery: unconditionally power-cycle
+            # admin-up DPUs to guarantee a known-good starting state.
+            oper_status = self.get_module_current_status(name)
+            if oper_status == str(ModuleBase.MODULE_STATUS_OFFLINE):
+                # Admin-down DPU; leave powered off
```
judsonwilson-nvidia, line 1314: This looks like it's the oper status, not the admin status? Should this be fetching the admin status?

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            fvs = self.module_table.get(module_name)
+            if isinstance(fvs, list) and fvs[0] is True:
+                fvs = dict(fvs[-1])
+                return fvs.get(STATE_TRANSITION_IN_PROGRESS, 'False') == 'True'
+        except Exception as e:
+            self.log_warning("{}: Failed to read state_transition_in_progress: {}".format(
+                module_name, e))
+        return False
+
+    def _power_cycle_dpu(self, module_name, module_index):
```
judsonwilson-nvidia, line 1259: personal opinion:

Rename this to _power_cycle_dpu_or_go_to_unrecoverable()

Currently it's hiding a state machine transition which is confusing while reading the code.

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+                continue
+            if not self._is_auto_recovery_enabled():
+                self.dpu_recovery_state[name]['state'] = DPU_STATE_MANUAL_INTERVENTION
+                continue
+            self._power_cycle_dpu(name, module_index)
+
+    def _npu_crash_on_last_boot(self):
+        """Return True if the last NPU reboot cause looks like a crash/OOM/unknown."""
+        try:
+            if not os.path.isfile(REBOOT_CAUSE_FILE):
```
judsonwilson-nvidia, line 1324: Is this logic correct - no file -> not a crash?  If that is correct, it would be nice if there were a short comment explaining why.

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+        whether the dpu-auto-recovery feature flag is enabled. Only the
+        actual power-cycle action is gated by the flag — when disabled, the
+        DPU transitions to ManualIntervention and waits for operator action.
+
+        Recovery is suppressed while a planned operation (shutdown/reboot) is
+        in progress for a DPU (state_transition_in_progress == True).
+        """
+        auto_recovery_enabled = self._is_auto_recovery_enabled()
+
+        for module_index in range(0, self.num_modules):
```
judsonwilson-nvidia, line 1347: Nit:

The body of this for-loop is huge and probably makes sense in its own helper function?

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            all_up = (mp_state == 'up' and cp_state == 'up' and dp_state == 'up')
+
+            # DPU is administratively down — no automatic recovery; just
+            # reflect the offline state and wait for `module startup`.
+            # This check must precede the Unrecoverable check so that an
+            # operator can shut down an unrecoverable DPU and then restart it.
+            if admin_status == 'down':
+                if current_state != DPU_STATE_OFFLINE:
+                    if current_state == DPU_STATE_UNRECOVERABLE:
+                        recovery['was_unrecoverable'] = True
```
judsonwilson-nvidia, line 1367: nit: 'was_unrecoverable' should be a constant

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            oper_status = self.get_module_current_status(name)
+            admin_status = self.get_module_admin_status(name)
+
+            all_up = (mp_state == 'up' and cp_state == 'up' and dp_state == 'up')
+
+            # DPU is administratively down — no automatic recovery; just
+            # reflect the offline state and wait for `module startup`.
+            # This check must precede the Unrecoverable check so that an
+            # operator can shut down an unrecoverable DPU and then restart it.
+            if admin_status == 'down':
```
judsonwilson-nvidia, line 1364: Should mention this way of handling admin down in the HLD if it's not.  I was wondering about this while reading  it.

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+                if current_state != DPU_STATE_OFFLINE:
+                    if current_state == DPU_STATE_UNRECOVERABLE:
+                        recovery['was_unrecoverable'] = True
+                    recovery['state'] = DPU_STATE_OFFLINE
+                    self._set_ready_status(name, 'false')
+                continue
+
+            # Terminal: requires chassisd reset or operator action
+            if current_state == DPU_STATE_UNRECOVERABLE:
```
judsonwilson-nvidia, line 1373: How does operator action get us out of this state?

sonic-chassisd/scripts/chassisd
gpunathilell, line 1269: shoultdnt it be `set_admin_state_gracefully` here? just to have the transition being set, so that user cannot invoke the admin state change while this is happening

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            mp_state, cp_state, dp_state = self._get_dpu_states(name)
+            oper_status = self.get_module_current_status(name)
+            admin_status = self.get_module_admin_status(name)
+
+            all_up = (mp_state == 'up' and cp_state == 'up' and dp_state == 'up')
+
+            # DPU is administratively down — no automatic recovery; just
+            # reflect the offline state and wait for `module startup`.
+            # This check must precede the Unrecoverable check so that an
+            # operator can shut down an unrecoverable DPU and then restart it.
```
judsonwilson-nvidia, line 1363:
```suggestion
            # operator can get the DPU out of the UNRECOVERABLE state by setting the DPU module to admin-down (which transitions it to the OFFLINE state).
```

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            if admin_status == 'down':
+                if current_state != DPU_STATE_OFFLINE:
+                    if current_state == DPU_STATE_UNRECOVERABLE:
+                        recovery['was_unrecoverable'] = True
+                    recovery['state'] = DPU_STATE_OFFLINE
+                    self._set_ready_status(name, 'false')
+                continue
+
+            # Terminal: requires chassisd reset or operator action
```
judsonwilson-nvidia, line 1372: Maybe mention here that the operater action can break out of this state by triggering the admin-down check above?

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            if current_state == DPU_STATE_UNRECOVERABLE:
+                continue
+
+            # Suppress recovery while a planned operation (shutdown/reboot)
+            # is in progress. The control plane goes down during planned
+            # operations; we must not misinterpret that as a failure.
+            if self._is_planned_transition_in_progress(name):
```
judsonwilson-nvidia, line 1379: I think this still allows race conditions, because it doesn't set the entry in the DB.

Would it be safer to set the field during resets in the same way a graceful reboot would?

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+                continue
+
+            # Admin-up DPU went Offline at the hardware level — treat as a
+            # hardware failure.
+            if oper_status == str(ModuleBase.MODULE_STATUS_OFFLINE) and \
+                    current_state not in (DPU_STATE_BOOTING, DPU_STATE_POWER_CYCLE,
+                                          DPU_STATE_WAIT_FOR_SELF_RECOVERY,
+                                          DPU_STATE_MANUAL_INTERVENTION):
```
judsonwilson-nvidia, line 1387: This is confusing to read and understand which states it does/doesn't happen in. Should this functionality be packaged into a helper function that is called at the beginning of certain state handlers, below?

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+                    "{}: Hardware failure detected (oper_status: Offline)".format(name))
+                self._set_ready_status(name, 'false')
+                self._set_last_down_time(name)
+                if auto_recovery_enabled:
+                    self._power_cycle_dpu(name, module_index)
+                else:
+                    recovery['state'] = DPU_STATE_MANUAL_INTERVENTION
+                continue
+
+            # Admin transitioned from down -> up; restart the recovery FSM.
```
judsonwilson-nvidia, line 1398: Mention that there is an admin-down check, above, and if execution reaches here, it must be admin up.

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            # If DPU was previously Unrecoverable, operator module startup
+            # resets recovery_status and reset_count.
+            if current_state == DPU_STATE_OFFLINE:
```
judsonwilson-nvidia, line 1401: Should we rename this state to "WaitForAdminUp" ? I think that would be more clear.

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+                continue
+
+            if current_state == DPU_STATE_MANUAL_INTERVENTION:
+                # Operator may have recovered the DPU manually.
+                if all_up:
+                    recovery['state'] = DPU_STATE_READY
+                    self._set_ready_status(name, 'true')
+                    self._set_last_ready_time(name)
+                    self.log_info("{}: DPU recovered (manual intervention)".format(name))
+                elif auto_recovery_enabled:
```
gpunathilell, line 1521: isnt this a loop? the only exit out of `DPU_STATE_MANUAL_INTERVENTION` is to `DPU_STATE_READY`, but assuming DPU is damaged we constantly retry powercycle of the DPU (ignoring reset_count)

vvolam: _power_cycle_dpu() always increments `reset_count` and checks it against `reset_limit`. After the call, the state transitions to either `PowerCycle` (if under the limit) or `Unrecoverable `(if at the limit). So the DPU never stays in `ManualIntervention` after auto-recovery is re-enabled — it immediately moves to another state, and the `reset_count` is fully respected. There's no infinite loop.

sonic-chassisd/scripts/chassisd
gpunathilell: same question, why are we setting state to unrecoverable after we try power cycle

sonic-chassisd/scripts/chassisd
gpunathilell: ignore, I see that the state is changed in power cycle function

sonic-chassisd/scripts/chassisd
judsonwilson-nvidia: Update:  I think this is more informative:

`_enter_power_cycle_or_unrecoverable()`

sonic-chassisd/scripts/chassisd
judsonwilson-nvidia: I see we renamed it to AdminDown in the HLD 👍

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+                    recovery['state'] = DPU_STATE_WAIT_FOR_SELF_RECOVERY
+                    recovery['self_recovery_start_time'] = time.time()
+                    recovery['self_recovery_poll_count'] = 0
+                    self._set_ready_status(name, 'false')
+                    self._set_last_down_time(name)
+                elif dp_state == 'down':
+                    # Data plane down only: mark not ready but no recovery.
+                    # Only write to DB on transition to avoid redundant writes.
+                    if recovery.get('dp_was_down') is not True:
+                        self._set_ready_status(name, 'false')
```
judsonwilson-nvidia, line 1456: It's kind of confusing that in the READY state the ready_status can be false. In hindsight, I would have preferred a name like like "CTRL_READY", or "RESPONSIVE" or maybe even just "OK" or "ONLINE".

Anyways, enough bikeshedding.

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+                    else:
+                        recovery['state'] = DPU_STATE_MANUAL_INTERVENTION
+                continue
+
+            if current_state == DPU_STATE_POWER_CYCLE:
+                if all_up:
+                    recovery['state'] = DPU_STATE_READY
+                    self._set_ready_status(name, 'true')
+                    self._set_last_ready_time(name)
+                    self.log_info("{}: DPU recovered after power-cycle".format(name))
```
judsonwilson-nvidia, line 1503: I think this should be encapsulated in a helper function that is also used by the BOOTING handler?

sonic-chassisd/scripts/chassisd
```diff
@@ -1125,6 +1192,338 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+                    self.log_warning("{}: Boot timeout ({}s) expired after power-cycle".format(
+                        name, self.dpu_boot_timeout))
+                    if auto_recovery_enabled:
+                        self._power_cycle_dpu(name, module_index)
+                    else:
+                        recovery['state'] = DPU_STATE_MANUAL_INTERVENTION
+                continue
+
+            if current_state == DPU_STATE_MANUAL_INTERVENTION:
+                # Operator may have recovered the DPU manually.
```
judsonwilson-nvidia, line 1515: Maybe mention in comment that code above can handle and transition from this state to AdminDown. For this state it's especially useful to know.

vvolam:
@judsonwilson-nvidia I will address all these minor comments as a follow-up PR this week.

judsonwilson-nvidia (approved):
Approving under the following context / conditions:

1) Feature should be harmless while it is disabled, with the main side effects being database updates at low frequency (on the order of 3 x num-dpus per every 10 seconds).

2) Feature is currently disabled in sonic-mgmt tests and will address feedback this week, per this comment, or set the feature default-disabled in general:

https://github.com/sonic-net/sonic-platform-daemons/pull/815#issuecomment-4597752129

3) Specifically, we need to resolve the state-transition handling for preventing race conditions with graceful-shutdown enabled features (admin-up, admin-down, reboot, ...). Gagan and I both believe that this code needs to set the state-transition status to true in the db before the power-cyle and clear it after the power cycle.

https://github.com/sonic-net/sonic-platform-daemons/pull/815/changes#r3336904708

image

gpunathilell (approved):

vvolam:
Merging the PR today. All open comments will be addressed as follow-up PR this week.

Issue #816, opened 2026-05-18
Regression in chassis_db_init causing PMON Failure


prajjwal-arista:
Regression seen in chassis_db_init due to this https://github.com/sonic-net/sonic-platform-daemons/pull/783 PR 
This prevents other processes inside pmon from starting as well and redis isn't populated with the correct data.

Please see: https://github.com/sonic-net/sonic-platform-daemons/pull/783#issuecomment-4474948147

    # Determine switch host serial number
    if platform_chassis.is_bmc():
        # BMC must also populate the switch host serial number
        switch_host_serial = try_get(platform_chassis.get_switch_host_serial)
    root@dut8:~# docker exec -it pmon /usr/local/bin/chassis_db_init
    Traceback (most recent call last):
    File "/usr/local/bin/chassis_db_init", line 113, in
    main()
    File "/usr/local/bin/chassis_db_init", line 110, in main
    provision_db(platform_chassis, log)
    File "/usr/local/bin/chassis_db_init", line 77, in provision_db
    switch_host_serial = try_get(platform_chassis.get_switch_host_serial)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AttributeError: 'Chassis' object has no attribute 'get_switch_host_serial'

judyjoseph:
@chander-nexthop @roger-nexthop f.y.i

chander-nexthop:
The following diff should fix this. Pasting it here to unblock.

```shell
[ip-172-31-47-43 sonic-platform-daemons ((b825806...))]> g diff
diff --git i/sonic-chassisd/scripts/chassis_db_init w/sonic-chassisd/scripts/chassis_db_init
index 80a1b52..ccd1e26 100644
--- i/sonic-chassisd/scripts/chassis_db_init
+++ w/sonic-chassisd/scripts/chassis_db_init
@@ -74,7 +74,8 @@ def provision_db(platform_chassis, log):
     # Determine switch host serial number
     if platform_chassis.is_bmc():
         # BMC must also populate the switch host serial number
-        switch_host_serial = try_get(platform_chassis.get_switch_host_serial)
+        get_switch_host_serial = getattr(platform_chassis, "get_switch_host_serial", None)
+        switch_host_serial = try_get(get_switch_host_serial) if get_switch_host_serial else NOT_AVAILABLE
     else:
         switch_host_serial = NOT_AVAILABLE
 
[...]

PR #817, opened 2026-05-20
Fix Issue #816: Regression in chassis_db_init causing PMON Failure

chander-nexthop:
Use getattr to ensure if the chassis object defines a get_switch_host_serial method before invoking the same. Nexthop's platforms implement the API to read the switch host serial number in the Chassis object that represents the BMC card. This API may not be present in other vendor's code, so check before invoking the same. Ideally this API must be part of the BMC chassis object as that object has collection of all system objects. Where the switch host's serial number is present may differ between vendors implementation and hence having it as part of chassis object (which can then get to the appropriate module object and read) is better IMHO rather than assuming that the CPU card will have it. @judyjoseph Let me know if this makes sense.

Issue #816

### DIFF
--- a/sonic-chassisd/scripts/chassis_db_init
+++ b/sonic-chassisd/scripts/chassis_db_init
@@ -11,6 +11,7 @@ try:
     import sys
 
     from sonic_py_common import daemon_base, logger
+    from sonic_platform_base.module_base import ModuleBase
 
     # If unit testing is occurring, mock swsscommon and module_base
     if os.getenv("CHASSIS_DB_INIT_UNIT_TESTING") == "1":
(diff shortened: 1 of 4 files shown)


sonic-chassisd/scripts/chassis_db_init
```diff
@@ -74,7 +74,8 @@ def provision_db(platform_chassis, log):
     # Determine switch host serial number
     if platform_chassis.is_bmc():
         # BMC must also populate the switch host serial number
-        switch_host_serial = try_get(platform_chassis.get_switch_host_serial)
+        get_switch_host_serial = getattr(platform_chassis, "get_switch_host_serial", None)
```
judyjoseph: @chander-nexthop since we don't have this API get_switch_host_serial() defined in platform API .. we would need to first get the module with type SWITCH-HOST first and call the get_serial() function [inherited from device_base.py] which needs to be implemented in the Module class in respective platform modules.

judyjoseph (dismissed):
LGTM

judyjoseph:
@chander-nexthop, could you fix the test failures

judyjoseph (approved):
LGTM

PR #818, opened 2026-05-21
[PMON] Bmcctld improvements & thermalctld: push thermals from Switch-Host --> BMC

judyjoseph:
#### Why I did it
Bmcctld improvements & thermalctld: push thermals from Switch-Host --> BMC

##### Work item tracking
- Microsoft ADO **(number only)**: N/A

#### How I did it

**sonic-bmcctld/scripts/bmcctld**

- Conditional SWITCH_HOST_POWER_ON_DELAY - applied only on REBOOT_CAUSE_POWER_LOSS; skipped on warm/fast/soft reboot. 
- GracefulShutdownHandler.execute() cleaned up
- Event-log de-duplication across handler functions
- Rack manager alert : skip when incoming alert matches to existing alert level, prevents unnecessary pub/sub events
- chassis module admin_status : skip when admin_status don't change, prevents unnecessary pub/sub events
- Uses self.chassis.is_liquid_cooled() platform API.
- Update the init_host_state API to derive the device_power_state from get_oper_status platform API which will either ONLINE/OFFLINE.
- Set the default graceful_shutdown_timeout = 0 till we support GNOI graceful shutdown

**sonic-thermalctld/scripts/thermalctld**

- EventLogger class added - stores critical events to syslog and /host/bmc/event.log.
- Switch-Host ↔ BMC, thermals are pushed to BMC, via TEMPERATURE_INFO_STATE_DB table
      - critical temperature breaches logged into BMC event log.
- LiquidCoolingUpdater improvements: 
                 - writes SYSTEM_LEAK_STATUS only on transitions 
                 - Per-sensor state updates also only on transitions
- Added the Leak severity escalation logic (MINOR → CRITICAL after max_minor_duration_sec).
- Critical leaks logged via event_logger.
- Add logic to use self.chassis.is_liquid_cooled() also to identify liquid cooled platforms"

#### How to verify it
All Unit tests pass and sonic-mgmt tests for bmcctld and thermalctld pass

- [ ] 201811
- [ ] 201911
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511
- [x] 202605

### DIFF
--- a/sonic-bmcctld/scripts/bmcctld
+++ b/sonic-bmcctld/scripts/bmcctld
@@ -227,7 +227,8 @@ DEFAULT_RACK_MGR_MINOR_ALERT_ACTION = ACTION_SYSLOG_ONLY
 
 # Default timing values (seconds)
 DEFAULT_POWER_ON_DELAY_SECS = 0     # 0 seconds - power on Switch-Host immediately by default
-DEFAULT_SHUTDOWN_DELAY_SECS = 120   # 2 minutes - wait for graceful shutdown before force
+DEFAULT_SHUTDOWN_DELAY_SECS = 0     # default 0 until real GNOI graceful shutdown is supported;
+                                    # operator may override via CHASSIS_MODULE|SWITCH-HOST/graceful_shutdown_timeout
 
 # Switch-Host module name key in CHASSIS_MODULE table
 SWITCH_HOST_MODULE_KEY = "SWITCH-HOST"
@@ -559,7 +560,7 @@ class PolicyReader(EventLogger):
         return DEFAULT_POWER_ON_DELAY_SECS
 
     def get_graceful_shutdown_timeout(self):
-        """Return graceful_shutdown_timeout in seconds from CHASSIS_MODULE|SWITCH-HOST (default: 120)."""
+        """Return graceful_shutdown_timeout in seconds from CHASSIS_MODULE|SWITCH-HOST (default: 0)."""
         raw = self._get_chassis_module_entry().get(FIELD_GRACEFUL_SHUTDOWN_TIMEOUT)
         if raw is not None:
             try:
--- a/sonic-bmcctld/scripts/bmcctld
+++ b/sonic-bmcctld/scripts/bmcctld
@@ -350,8 +350,8 @@ class SwitchHostController(EventLogger):
         Poll get_oper_status() until it matches expected_status or timeout_secs elapses.
         Returns True when confirmed, False on timeout.
         """
-        deadline = time.time() + timeout_secs
-        while time.time() < deadline:
+        deadline = time.monotonic() + timeout_secs
+        while time.monotonic() < deadline:
             if self.get_oper_status() == expected_status:
                 self.log_info("{}: oper_status confirmed {}".format(context, expected_status))
                 return True
@@ -968,17 +968,20 @@ class BmcEventHandler(EventLogger):
     def _handle_rack_mgr_alert(self, key, fvs):
         """Process RACK_MANAGER_ALERT change from Rack Manager (via redfish docker)."""
         severity = fvs.get(FIELD_SEVERITY) or fvs.get(FIELD_LEAK, ALERT_SEVERITY_NORMAL)
-        if self._last_rack_mgr_alert_severity.get(key) == severity:
-            return
-        self._last_rack_mgr_alert_severity[key] = severity
-
         policy = self.policy_reader.get_leak_control_policy()
 
         if policy.get("rack_mgr_leak_policy") == "disabled":
-            self._event_log.log_notice(
-                "RACK_MGR_ALERT handler done: key={} severity={} result=SKIPPED (policy=disabled)".format(
-                    key, severity))
+            # Log only on transition; do NOT update dedup state so that a later
+            # policy re-enable + same severity re-publish still dispatches the action.
+            if self._last_rack_mgr_alert_severity.get(key) != severity:
+                self._event_log.log_notice(
+                    "RACK_MGR_ALERT handler done: key={} severity={} result=SKIPPED (policy=disabled)".format(
+                        key, severity))
             return
+
+        if self._last_rack_mgr_alert_severity.get(key) == severity:
+            return
+        self._last_rack_mgr_alert_severity[key] = severity
 
         if severity == ALERT_SEVERITY_CRITICAL:
             action = policy.get("rack_mgr_critical_alert_action", DEFAULT_RACK_MGR_CRITICAL_ALERT_ACTION)
--- a/sonic-bmcctld/scripts/bmcctld
+++ b/sonic-bmcctld/scripts/bmcctld
@@ -81,13 +81,18 @@ class EventLogger(logger.Logger):
                                  os.path.basename(log_file).replace(".", "_"))
         self._file_logger = logging.getLogger(py_name)
         if not self._file_logger.handlers and os.getenv("BMCCTLD_UNIT_TESTING") != "1":
-            handler = logging.FileHandler(log_file)
-            handler.setFormatter(logging.Formatter(
-                "%(asctime)s %(name)s[%(process)d] %(levelname)s: %(message)s",
-                datefmt="%Y-%m-%dT%H:%M:%S",
-            ))
-            self._file_logger.addHandler(handler)
-            self._file_logger.setLevel(logging.DEBUG)
+            try:
+                handler = logging.FileHandler(log_file)
+                handler.setFormatter(logging.Formatter(
+                    "%(asctime)s %(name)s[%(process)d] %(levelname)s: %(message)s",
+                    datefmt="%Y-%m-%dT%H:%M:%S",
+                ))
+                self._file_logger.addHandler(handler)
+                self._file_logger.setLevel(logging.DEBUG)
+            except (OSError, IOError) as e:
+                self.log_error(
+                    "EventLogger: failed to open log file {} ({}); events will go to syslog only".format(
+                        log_file, repr(e)))
 
     def log_debug(self, msg, also_print_to_console=False):
         super(EventLogger, self).log_debug(msg, also_print_to_console)
--- a/sonic-bmcctld/scripts/bmcctld
+++ b/sonic-bmcctld/scripts/bmcctld
@@ -50,6 +50,7 @@ try:
     from datetime import datetime
 
     from sonic_py_common import daemon_base, logger
+    from sonic_platform_base.chassis_base import ChassisBase
 
     try:
         from sonic_py_common import is_liquid_cooled
@@ -598,7 +599,7 @@ class PolicyReader(EventLogger):
 # CriticalEventChecker
 # ============================================================
 
-class CriticalEventChecker(EventLogger):
+class CriticalEventChecker(logger.Logger):
     """Checks current system and Rack Manager leak/alert status from STATE_DB."""
 
     # Alert keys in RACK_MANAGER_ALERT table
@@ -610,7 +611,7 @@ class CriticalEventChecker(EventLogger):
     ]
 
     def __init__(self, policy_reader):
-        super(CriticalEventChecker, self).__init__(SYSLOG_IDENTIFIER, CRITICAL_EVENT_LOG_FILE)
+        super(CriticalEventChecker, self).__init__(SYSLOG_IDENTIFIER)
         self.state_db = daemon_base.db_connect("STATE_DB")
         self.policy_reader = policy_reader
 
@@ -620,7 +621,7 @@ class CriticalEventChecker(EventLogger):
         result = tbl.get(SYSTEM_LEAK_STATUS_KEY)
         if result and result[0]:
             if fvs_to_dict(result[1]).get(FIELD_DEVICE_LEAK_STATUS) == SYSTEM_LEAK_CRITICAL:
-                _event_logger.log_error("CRITICAL system leak detected in {}: {}={}".format(
+                self.log_error("CRITICAL system leak detected in {}: {}={}".format(
                     SYSTEM_LEAK_STATUS_TABLE, FIELD_DEVICE_LEAK_STATUS, SYSTEM_LEAK_CRITICAL))
                 return True
         return False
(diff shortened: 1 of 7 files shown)


sonic-thermalctld/scripts/thermalctld
Yakiv-Huryk: Can you please also wrap the `get_leak_max_minor_duration_sec()` value in `str()`? 
Currently, we get:
```
pmon#supervisord: thermalctld   File "/usr/local/bin/thermalctld", line 545, in __init__
pmon#supervisord: thermalctld     fvs = swsscommon.FieldValuePairs([
pmon#supervisord: thermalctld           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pmon#supervisord: thermalctld   File "/usr/lib/python3/dist-packages/swsscommon/swsscommon.py", line 224, in __init__
pmon#supervisord: thermalctld     _swsscommon.FieldValuePairs_swiginit(self, _swsscommon.new_FieldValuePairs(*args))
pmon#supervisord: thermalctld                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pmon#supervisord: thermalctld TypeError: Wrong number or type of arguments for overloaded function 'new_FieldValuePairs'.
pmon#supervisord: thermalctld   Possible C/C++ prototypes are:
pmon#supervisord: thermalctld     std::vector< std::pair< std::string,std::string > >::vector()
pmon#supervisord: thermalctld     std::vector< std::pair< std::string,std::string > >::vector(std::vector< std::pair< std::string,std::string > > const &)
pmon#supervisord: thermalctld     std::vector< std::pair< std::string,std::string > >::vector(std::vector< std::pair< std::string,std::string > >::size_type)
pmon#supervisord: thermalctld     std::vector< std::pair< std::string,std::string > >::vector(std::vector< std::pair< std::string,std::string > >::size_type,std::vector< std::pair< std::string,std::string > >::value_type const &)
```

judyjoseph: Done

sonic-bmcctld/scripts/bmcctld
```diff
@@ -1267,7 +1273,7 @@ class BmcctldDaemon(daemon_base.DaemonBase):
             self._event_log.log_notice("STARTUP: Switch-Host already ONLINE (daemon restart); "
                                        "skipping boot sequence")
             self.controller.refresh_host_state()
-        elif not is_liquid_cooled():
+        elif not self.chassis.is_liquid_cooled():
             self._event_log.log_notice("STARTUP: non-liquid-cooled system — "
                                        "immediate power_on, no additional checks")
             self.controller.power_on()
```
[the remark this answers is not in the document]
judyjoseph: do we really neat as in abstract implementation it returns False : https://github.com/sonic-net/sonic-platform-common/blob/ef6286f36f30024bc700e117b576d6b159be9049/sonic_platform_base/chassis_base.py#L177

judyjoseph:
@fraserg-arista can you review as well .. there are changes in thermalctld

sonic-bmcctld/scripts/bmcctld
```diff
@@ -942,60 +961,39 @@ class BmcEventHandler(EventLogger):
...
-            if above:
-                self._event_log.log_notice(
-                    "Switch-Host thermal sensors above threshold ({} sensor(s)):\n{}".format(
-                        len(above), "\n".join(above)))
-            else:
-                self._event_log.log_notice(
-                    "Switch-Host: no thermal sensors above threshold at time of alert")
-        except Exception as e:
-            self.log_warning("Failed to read Switch-Host thermal sensors: {}".format(repr(e)))
+            self._event_log.log_notice(
```
Yakiv-Huryk, line 960: I think we need to consider lowering the severity here, since this way we will get 2 notice messages per second: https://github.com/sonic-net/sonic-platform-daemons/issues/824

sonic-thermalctld/scripts/thermalctld
```diff
@@ -618,6 +681,17 @@ class LiquidCoolingUpdater(threading.Thread, logger.Logger):
                             max_minor = leak_start_time + timedelta(seconds=max_minor_secs)
                             if max_minor <= now:
                                 sensor_leak_severity = LeakSeverity.CRITICAL
+                                escalated_to_critical = True
+
+                    if new_leak and sensor_leak_severity == LeakSeverity.CRITICAL:
+                        self.event_logger.log_error(
+                            'CRITICAL leak reported by sensor {}'.format(sensor_name))
+                        self.critical_sensors.add(sensor_name)
+                    elif escalated_to_critical:
```
fraserg-arista: I think this also needs to check whether the sensor is already in critical_sensors or it might repeatedly log here?

sonic-thermalctld/scripts/thermalctld
```diff
@@ -628,28 +702,45 @@ class LiquidCoolingUpdater(threading.Thread, logger.Logger):
...
+            sensor_location = try_get(sensor.get_leak_sensor_location, 'unknown')
+            fvs_tuple = (
                 ('name', sensor_name),
                 ('leaking', leaking),
                 ('leak_sensor_status', 'Good' if sensor_is_ok else 'Fault'),
-                ('type', str(sensor.get_leak_sensor_type())),
-                ('location', str(sensor.get_leak_sensor_location())),
+                ('type', str(sensor_type)),
+                ('location', str(sensor_location)),
                 ('severity', str(sensor_leak_severity)),
```
fraserg-arista: I just noticed that this field name is wrong; per the HLD, it should be 'leak_severity', not 'severity'. Would you mind fixing this as a drive-by?

sonic-bmcctld/scripts/bmcctld
judyjoseph: I hope this is already taken care of @Yakiv-Huryk  can you check

sonic-thermalctld/scripts/thermalctld
judyjoseph: updated

yxieca (changes requested):
## Review summary

Substantial PR with mostly good direction. One blocking bug + several non-blocking but high-value asks before merge.

---

### 🔴 Blocking

**B2 — `_handle_rack_mgr_alert` dedup-state poisoning (state-corrupting bug)**

The new dedup map (`self._last_rack_mgr_alert_severity`) is updated **before** the `policy.rack_mgr_leak_policy == "disabled"` early-return:

```python
def _handle_rack_mgr_alert(self, key, fvs):
    severity = fvs.get(FIELD_SEVERITY) or fvs.get(FIELD_LEAK, ALERT_SEVERITY_NORMAL)
    if self._last_rack_mgr_alert_severity.get(key) == severity:
        return
    self._last_rack_mgr_alert_severity[key] = severity   # ← mutated BEFORE policy check

    policy = self.policy_reader.get_leak_control_policy()
    if policy.get("rack_mgr_leak_policy") == "disabled":
        ...
        return
```

**Failure mode (operator workflow):**
1. Operator disables `rack_mgr_leak_policy` for maintenance.
2. Rack Manager publishes `severity=CRITICAL` → dedup state := CRITICAL, policy disabled, no action.
3. Operator re-enables the policy.
4. Rack Manager re-publishes the same `severity=CRITICAL` → dedup says "already saw it" → handler returns early → **CRITICAL alert silently dropped despite active policy**. No log emitted.

**Suggested fix:** move dedup state mutation after the policy gate so disabled-policy paths don't poison the map:

```python
def _handle_rack_mgr_alert(self, key, fvs):
    severity = fvs.get(FIELD_SEVERITY) or fvs.get(FIELD_LEAK, ALERT_SEVERITY_NORMAL)
    policy = self.policy_reader.get_leak_control_policy()

    if policy.get("rack_mgr_leak_policy") == "disabled":
        # Log only on transition, but do NOT update dedup state.
        if self._last_rack_mgr_alert_severity.get(key) != severity:
            self._event_log.log_notice(
                "RACK_MGR_ALERT handler done: key={} severity={} result=SKIPPED (policy=disabled)".format(
                    key, severity))
        return

    if self._last_rack_mgr_alert_severity.get(key) == severity:
        return
    self._last_rack_mgr_alert_severity[key] = severity
    # ... rest of dispatch
```

Please add a test covering: policy=disabled + CRITICAL arrives → policy re-enabled + same CRITICAL re-arrives → action MUST dispatch.

---

### 🟡 High-priority asks (non-blocking but please address before merge)

**H1 — `LiquidCoolingUpdater.__init__` startup seed bypasses transition guard**

```python
self.system_table.set('system', swsscommon.FieldValuePairs([
    ('device_leak_status', 'None'),
    ('timestamp', datetime.now().strftime('%Y%m%d %H:%M:%S')),
]))
```

This unconditional seed runs on every thermalctld restart. With the new transition-only writes in `_refresh_leak_status`, on a daemon restart **during an active CRITICAL leak**, subscribers see `device_leak_status: None` published transiently and don't transition back to CRITICAL until the next refresh cycle (up to `interval` seconds later). During that gap, `bmcctld.CriticalEventChecker.system_leak_critical()` reads `None` and returns `False` — admin commands / rack-manager retries could pass the leak check.

**Suggested fix:** seed only if the row is absent.
```python
existing = self.system_table.get('system')
if not (existing and existing[0]):
    self.system_table.set('system', ...)
```

**H2 — `severity` → `leak_severity` rename is a schema break; please link cross-repo companion PR(s)**

The test asserts `severity` is **removed** from the FVS (`assert 'severity' not in fvp.fv_dict`), not just renamed-with-alias. Consumers that need to be ported:
- `system-health hardware_checker`
- `leakageshow` CLI in sonic-utilities

The PR description acknowledges `leak_status` back-compat (for the `leaking` value) but doesn't link the companion PR(s) that port consumers from `severity` → `leak_severity`. Please link them, and document the deprecation plan for `leak_status` alias if it's intended to be temporary.

**H3 — Please link the `pmon-bmc-design` HLD**

Code references `pmon-bmc-design §2.4.1 (phase 1: syslog-only, no action)` but the PR description doesn't link the HLD. Please link it and confirm HLD consensus has been reached — the architectural shape ("BMC mirrors Switch-Host TEMPERATURE_INFO via remote DB connector, then re-reads its own STATE_DB to log critical events") deserves HLD-level review before code-level approval.

**H4 — `/host/bmc/event.log` has no rotation policy**

`logging.FileHandler` (not `RotatingFileHandler`) and no `logrotate.d` config visible. Indefinitely growing log file = disk-fill risk on long-running BMC. Please add either:
- A companion sonic-buildimage PR with a logrotate config, or
- Switch to `RotatingFileHandler` with sensible size/backup limits.

**H5 — `_bmc_table_set` / `_init_bmc_temperature_table` — syslog flood with no backoff**

When the BMC link is down, every thermal cycle × every sensor invokes `db_connect_remote` with no backoff:
- `update_interval=5s` × ~10 sensors → ~120 reconnect attempts/min, each potentially logging a warning.

**Suggested fix:** track `_last_init_attempt` and respect a min reconnect interval (e.g., 30s) when the previous init failed.

**H6 — Deleted `_log_host_thermal_sensors_above_threshold` — please document the rationale**

Pre-PR `BmcEventHandler._log_host_thermal_sensors_above_threshold` was called from `_handle_rack_mgr_alert` on CRITICAL severity to dump current Switch-Host thermal sensor state to event log — a forensic snapshot at action time. This PR deletes it without rationale.

The new BMC-side `ThermalMonitor._check_switch_host_thermals` is a **different** signal (continuous monitoring of TEMPERATURE_INFO, not a snapshot triggered by rack-manager-alert). If the intent is "continuous monitoring supersedes the snapshot," please say so in the PR body. Otherwise, restore the snapshot call.

---

### Grounded nits

1. **`STATE_DB_ID = 6`** — use `swsscommon.STATE_DB` symbolic constant.
2. **`time.monotonic()` migration in bmcctld is incomplete** — only the boot-delay loop was migrated; other `time.time()` deadline computations (e.g., `_verify_oper_status`) remain. Sweep or leave a TODO.
3. **`_check_switch_host_thermals` recovery is silent**, but `LiquidCoolingUpdater` logs both leak detect AND clear — asymmetric. Is the thermal-recovery silence intentional? Operators chasing an incident benefit from a recovery line to bracket the window.
4. **`fvs_to_dict` (bmcctld) vs inline `dict(fvs)` (thermalctld)** — minor style inconsistency.
5. **Test coverage gaps** worth adding alongside the fixes:
   - CRITICAL → MINOR sensor downgrade transition (`log_notice('... downgraded from CRITICAL to ...')`)
   - CRITICAL → MINOR → CRITICAL flap (verify re-log once after downgrade)
   - Empty-string `admin_status` in `_handle_chassis_module` (must not poison dedup map)
6. **PR description trailing-quote typo** at end of the thermalctld bullet (`...liquid cooled platforms"`).

---

Overall: solid direction, good test additions for the new state machines, and the transition-only writes are a clear improvement. The B2 dedup poisoning is the one bug I'd hold the merge for; everything else is "let's get this right before it ships."

fraserg-arista:
The copilot issues look worth addressing but other than that, the thermalctld changes LGTM.

yxieca (approved):
Thanks for the updates — the rack-mgr-alert dedup reordering looks correct now (dedup state is no longer mutated under `policy=disabled`, so a later re-enable + same severity still dispatches), and the new regression test for that path is exactly right. The `time.monotonic()` migration on `_verify_oper_status`, the seed-if-absent guard on `SYSTEM_LEAK_STATUS`, the BMC reconnect warning suppression, and the empty-`admin_status` dedup guard all look good. Approving.

Two non-blocking follow-ups for visibility (don't need to hold the merge):
- The `severity` → `leak_severity` field rename touches a STATE_DB schema field consumed downstream (system-health hardware_checker, leakageshow CLI). The added `leak_status` back-compat field is appreciated — could you link the companion consumer PR(s) so they land in lockstep?
- This change set scrubbed the inline `pmon-bmc-design` reference. Could you link the HLD in the PR description and confirm consensus on it? Helps future readers trace the architectural intent (BMC mirroring Switch-Host `TEMPERATURE_INFO`).

judyjoseph:
Thanks @yxieca. will raise a follow up PR to fix the remaining no blocking comments.

PR #820, opened 2026-05-22
thermalctld: convert leak data to strings before writing to db

### LINKED ISSUE #27515: Bug: Leakage detection feature broken by newly changes
### Is it platform specific

generic

### Importance or Severity

Critical

### Description of the bug

the change https://github.com/sonic-net/sonic-platform-daemons/pull/776 introduces 2 issues:
1. the legacy leak_status field had been removed, instead, the leaking had been used, this db scheme is not aligned with HLD anymore, causes CLI and system-health failure since they still looking for the leak_status field
2. during initialization, there isn't any prevent of non-implemented API such as get_leak_sensor_type and get_leak_sensor_location, the FieldValuePairs does NOT accpet None as parameter which causes thread directly crash.

### Steps to Reproduce

the problem is 100% reproduce, on liquid cooling system the CLI will report liquid cooling not detected.

feature should work as before

### Relevant log output

```shell

```

```shell

```

### Attach files (if any)

_No response_

fraserg-arista:
thermalctld: convert leak data to strings before writing to db

#### Description

This helps avoid issues where implementations return non-string data, which will fault when trying to write to the state db.

#### Motivation and Context

Partially fixes https://github.com/sonic-net/sonic-buildimage/issues/27515

Test suite for thermalctld was run.

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -564,7 +564,7 @@ class LiquidCoolingUpdater(threading.Thread, logger.Logger):
         for profile in self.liquid_cooling.get_all_profiles():
             fvs = swsscommon.FieldValuePairs([
                 ('type', profile.get_type()),
-                ('max_minor_duration_sec', profile.get_leak_max_minor_duration_sec()),
+                ('max_minor_duration_sec', str(profile.get_leak_max_minor_duration_sec())),
             ])
             self.profile_table.set(profile.get_type(), fvs)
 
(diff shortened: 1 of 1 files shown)


fraserg-arista:
@judyjoseph

judyjoseph (approved):
LGTM

keboliu:
@fraserg-arista, since it's not able to be cherry-picked to 202605, would you please raise a separate PR to the 202605 branch?

fraserg-arista:
This merged before the 202605 branch was created and doesn't need to be cherry-picked.

PR #825, opened 2026-05-28
Remove PENDING flag check in CMIS SM

### LINKED ISSUE #27582: Bug: CMIS SM checks datapath init pending as necessary condition to continue the SM
### Is it platform specific

generic

### Importance or Severity

Critical

### Description of the bug

 To configure the module, CMIS SM has a condition that checks whether DP_INIT_PENDING flag is ON. If not, it fails the SM an raises the number of retires by 1. 
https://github.com/sonic-net/sonic-platform-daemons/blob/master/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py#L1187

this condition is not needed because:
1. CMIS spec doesn’t define the DP_INIT_PENDING flag to be a solid indicator for the host to verify DP_INIT state. - it’s only additional information and NOS cannot rely on it to continue/stop the CMIS SM. 
DP_INIT_PENDING flag indicates the HW configuration between successful apply until DataPathInit state. 

2. CMIS spec does not define a minimal time for PENDING flag to remain raised.
The duration of DP_INIT_PENDING flag as ON, is dependent on HW configuration speed -  It is not guaranteed that NOS could catch the PENDING flag. In case the module is fast, the value is written to HW, and module state is moved to the next (DP_INIT/DataPathInitialized)

3. In the next CMIS state - TXON, we check if the state is DataPathInitialized. This is much stronger check, and dince it happens, it makes the DP_INIT_PENDING not needed.

Attached is the DP_INIT_PENDING from CMIS spec:

Image

### Steps to Reproduce

Run CMIS SM with a very fast module with CMIS rev 5 and above. 
you might "miss" the pending flag on, as it was already off and moved to next state.
even if you missed it, in TXON state you will see the module is in the right state and can continue to finish module configuration. 

N/A

### Relevant log output

```shell

```

```shell

```

### Attach files (if any)

_No response_

noaOrMlnx:
#### Description
Removed the check in CMIS SM that waits for PENDING flag. 
This PR aims to fix https://github.com/sonic-net/sonic-buildimage/issues/27582

#### Motivation and Context
The removed code block is not accurate according to CMIS spec - this flag is not a good indicator for states transition. 
1. The DP_INIT_PENDING flag is not defined as a reliable indicator for validating the DP_INIT state. It is intended only as supplementary information and should not be used by the NOS to control state progression.
2. The specification does not define a minimum duration for which the PENDING flag must remain asserted. On faster modules, the transition may complete before the NOS has a chance to observe this flag.
3.  In the next CMIS state (TXON), we will check whether the module state is DataPathInitialized. This is the required indicator, making the PENDING check unnecessary.

This is only informational flag, so removing it cannot fail anything. 
The next state check is the right indicator and already exists. 
this was tested with CMIS host mgmt. modules.

### DIFF
--- a/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
+++ b/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
@@ -1181,14 +1181,6 @@ class CmisManagerTask(threading.Thread):
                     self.force_cmis_reinit(lport)
                     return
 
-                if hasattr(api, 'get_cmis_rev'):
-                    # Check datapath init pending on module that supports CMIS 5.x
-                    majorRev = int(api.get_cmis_rev().split('.')[0])
-                    if majorRev >= 5 and not self.check_datapath_init_pending(api, host_lanes_mask):
-                        self.log_notice("{}: datapath init not pending".format(lport))
-                        self.force_cmis_reinit(lport, retries + 1)
-                        return
-
                 # Ensure the Datapath is NOT Activated unless the host Tx siganl is good.
                 # NOTE: Some CMIS compliant modules may have 'auto-squelch' feature where
                 # the module won't take datapaths to Activated state if host tries to enable


prgeor (changes requested):
@noaOrMlnx we discussed this that module vendor should fix this in firmware.

noaOrMlnx:
@prgeor correct. 
Once the fix will be pushed to module FW, I will close this PR. for now, I moved it to draft

PR #829, opened 2026-06-02
[chassisd]: Enhance DPU auto-recovery robustness and race condition handling

vvolam:
#### Description

Enhance the DPU auto-recovery state machine in chassisd to fix race conditions between concurrent operations (shutdown/reboot vs power-cycle), improve correctness of the default behavior, and refactor for maintainability.

Key changes:
- Read `dpu_auto_recovery` from `DEVICE_METADATA` instead of `FEATURE` table
- Default to **disabled** when the `dpu_auto_recovery` field is missing (safer default)
- Direct transition to `ManualIntervention` when auto-recovery is disabled (skip `WaitForSelfRecovery`)
- PCIe detach/reattach around power-cycle for clean hardware reset
- Check `reset_limit` before issuing power-cycle to avoid wasted resets
- Acquire state transition lock during power-cycle to prevent races with gnoi shutdown/reboot
- Re-check `transition_in_progress` after lock acquisition to close TOCTOU window
- Abort power-cycle if lock cannot be acquired (another operation in progress)
- Unify field name: use `transition_in_progress` (drop legacy `state_transition_in_progress`)
- Ignore `transition_type == 'recovery'` so chassisd does not suppress itself
- Refactor: extract `_process_single_dpu_recovery` to reduce `update_dpu_recovery_state` size
- Refactor: extract `_enter_booting`/`_enter_ready`/`_enter_wait_for_self_recovery` transition helpers
- Refactor: extract `_should_handle_hw_offline` for offline-state guard clarity
- Refactor: extract `_handle_boot_timeout_expired` shared by Booting and PowerCycle states

Depends on sonic-net/sonic-platform-common PR for `get_module_state_transition`/`set_module_state_transition`/`clear_module_state_transition` platform API.

#### Motivation and Context

Addresses review comments from PR #815. The DPU auto-recovery feature had several race conditions:
1. **Power-cycle recovery vs planned shutdown**: If gnoi initiated a shutdown while chassisd was about to power-cycle, both operations would conflict, potentially bricking the DPU.
2. **TOCTOU between check and lock**: A gap between checking if a transition was in progress and acquiring the lock allowed concurrent operations to slip through.
3. **Default-enabled risk**: Missing `dpu_auto_recovery` field defaulted to enabled, which could cause unexpected power-cycles on systems without explicit configuration.

1. **Unit tests**: Comprehensive unit tests added in `tests/test_dpu_auto_recovery.py` covering all state transitions, race conditions, lock acquisition failures, and edge cases.

2. **On-device testing** on a **202605/master** image (SmartSwitch), via the sonic-mgmt `smartswitch/platform_tests` suite, which exercises the DPU failure-mode / auto-recovery scenarios:
   - Control-plane-down → auto-recovery power-cycle → DPU returns to `ready_status=true`
   - Auto-recovery disabled → direct `ManualIntervention` (no power-cycle)
   - Power-cycle with lock contention (gnoi shutdown holding the transition lock)
   - TOCTOU re-check after lock acquisition
   - Self-suppression (`transition_type == 'recovery'` exclusion)
   - Unrecoverable state after `reset_limit` exceeded; admin-down/up clears it
   - NPU crash → bulk power-cycle of all admin-up DPUs

   **Final results (202605/master):** the end-to-end auto-recovery path was confirmed — chassisd power-cycled the failed DPU and it returned to `ready_status=true`. After the accompanying fixes (device `DEVICE_METADATA.dpu_auto_recovery` YANG leaf, the `module_base` `recovery` transition-timeout default, and a recovery-timeout bump), `test_dpu_failure_modes` finished **5 passed / 2 failed / 1 skipped (0 errored)**. Both remaining failures were environmental/test-harness — a `databasedpu` container inactive at test start, and a config-reload case that ran on pre-fix test code — not auto-recovery regressions.

3. **Verified**: No regressions on remaining DPUs (remained in Ready state throughout DPU0 destructive testing)

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -1434,7 +1434,15 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
                 return False
             with open(REBOOT_CAUSE_FILE, 'r') as f:
                 cause = f.read().strip().lower()
-            return any(kw in cause for kw in NPU_CRASH_REBOOT_CAUSE_KEYWORDS)
+
+            # Match explicit crash signals only. This avoids classifying
+            # normal operator-initiated reboot strings as crashes.
+            if 'kernel panic' in cause:
+                return True
+
+            # Keep supporting legacy "unknown" reboot-cause content, but only
+            # as the full value/prefix (not arbitrary substring matches).
+            return cause == 'unknown' or cause.startswith('unknown:') or cause.startswith('unknown ')
         except Exception as e:
             self.log_warning("Failed to read {}: {}".format(REBOOT_CAUSE_FILE, e))
             return False
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -1314,19 +1314,6 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
             recovery['boot_start_time'] = time.time()
             return
 
-        # Re-check transition_in_progress after acquiring our lock to close
-        # the TOCTOU window where gnoi/CLI sets transition_in_progress
-        # between _is_planned_transition_in_progress() and here.
-        if self._is_planned_transition_in_progress(module_name, module_index):
-            self.log_warning("{}: transition_in_progress detected "
-                            "after lock acquisition — aborting power-cycle".format(module_name))
-            try_get(module.clear_module_state_transition, module_name, default=False)
-            recovery['reset_count'] -= 1
-            self._set_reset_count(module_name, recovery['reset_count'])
-            recovery['state'] = DPU_STATE_POWER_CYCLE
-            recovery['boot_start_time'] = time.time()
-            return
-
         try:
             # Keep the same graceful sequence as module_base:
             # pre-shutdown -> admin down -> admin up -> post-startup.
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -151,6 +151,9 @@ NPU_CRASH_REBOOT_CAUSE_KEYWORDS = ("kernel panic")
 # Internal recovery dict key for tracking prior Unrecoverable state
 WAS_UNRECOVERABLE_KEY = 'was_unrecoverable'
 
+# Transition type value used by chassisd when acquiring the state-transition lock
+TRANSITION_TYPE_RECOVERY = 'recovery'
+
 # DPU recovery states
 DPU_STATE_BOOTING = 'Booting'
 DPU_STATE_READY = 'Ready'
@@ -1263,7 +1266,7 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
         Uses the platform API get_module_state_transition() which handles
         timeout-based auto-clearing. Returns True when a non-recovery
         transition is active, suppressing auto-recovery.
-        Ignores transition_type == 'recovery' so chassisd does not suppress itself.
+        Ignores transition_type == TRANSITION_TYPE_RECOVERY so chassisd does not suppress itself.
         """
         module = self.chassis.get_module(module_index)
         in_progress = try_get(module.get_module_state_transition, module_name, default=False)
@@ -1274,7 +1277,7 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
             fvs = self.module_table.get(module_name)
             if isinstance(fvs, list) and fvs[0] is True:
                 fvs = dict(fvs[-1])
-                return fvs.get('transition_type') != 'recovery'
+                return fvs.get('transition_type') != TRANSITION_TYPE_RECOVERY
         except Exception:
             pass
         return True
@@ -1302,12 +1305,14 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
         # shutdown/reboot from operating on this DPU during power-cycle.
         # If the lock cannot be acquired (another operation holds it),
         # abort the power-cycle — the other operation owns this DPU.
-        lock_acquired = try_get(module.set_module_state_transition, module_name, "recovery", default=False)
+        lock_acquired = try_get(module.set_module_state_transition, module_name, TRANSITION_TYPE_RECOVERY, default=False)
         if not lock_acquired:
             self.log_warning("{}: Cannot acquire state transition lock; "
                             "another operation in progress — skipping power-cycle".format(module_name))
             recovery['reset_count'] -= 1
             self._set_reset_count(module_name, recovery['reset_count'])
+            recovery['state'] = DPU_STATE_POWER_CYCLE
+            recovery['boot_start_time'] = time.time()
             return
 
         # Re-check transition_in_progress after acquiring our lock to close
@@ -1319,6 +1324,8 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
             try_get(module.clear_module_state_transition, module_name, default=False)
             recovery['reset_count'] -= 1
             self._set_reset_count(module_name, recovery['reset_count'])
+            recovery['state'] = DPU_STATE_POWER_CYCLE
+            recovery['boot_start_time'] = time.time()
             return
 
         try:
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -128,8 +128,8 @@ RECOVERY_UNRECOVERABLE = 'unrecoverable'
 
 # DPU auto-recovery field in CONFIG_DB DEVICE_METADATA|localhost
 DPU_AUTO_RECOVERY_FIELD = 'dpu_auto_recovery'
-DPU_AUTO_RECOVERY_ENABLED = 'enabled'
-DPU_AUTO_RECOVERY_DISABLED = 'disabled'
+DPU_AUTO_RECOVERY_ENABLED = 'enable'
+DPU_AUTO_RECOVERY_DISABLED = 'disable'
 
 # Platform.json field names for recovery thresholds
 PLATFORM_CFG_DPU_REBOOT_TIMEOUT = 'dpu_reboot_timeout'
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -1325,10 +1325,17 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
             return
 
         try:
-            try_get(module.pci_detach, default=False)
+            # Keep the same graceful sequence as module_base:
+            # pre-shutdown -> admin down -> admin up -> post-startup.
+            # Treat NotImplemented pre/post hooks as no-op success.
+            if not try_get(module.module_pre_shutdown, default=True):
+                self.log_warning("{}: module_pre_shutdown() failed".format(module_name))
+
             try_get(module.set_admin_state, MODULE_ADMIN_DOWN, default=False)
             try_get(module.set_admin_state, MODULE_ADMIN_UP, default=False)
-            try_get(module.pci_reattach, default=False)
+
+            if not try_get(module.module_post_startup, default=True):
+                self.log_warning("{}: module_post_startup() failed".format(module_name))
         except Exception as e:
             self.log_error("{}: Power-cycle failed: {}".format(module_name, e))
         finally:
@@ -1406,9 +1413,9 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
 
             # NPU kernel crash recovery: unconditionally power-cycle
             # admin-up DPUs to guarantee a known-good starting state.
-            oper_status = self.get_module_current_status(name)
-            if oper_status == str(ModuleBase.MODULE_STATUS_OFFLINE):
-                # Admin-down DPU; leave powered off
+            admin_status = self.get_module_admin_status(name)
+            if admin_status == 'down':
+                # Admin-down DPU; leave powered off.
                 continue
             if not self._is_auto_recovery_enabled():
                 self.dpu_recovery_state[name]['state'] = DPU_STATE_MANUAL_INTERVENTION
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -146,7 +146,7 @@ MINIMUM_SELF_RECOVERY_POLL_COUNT = 3
 
 # Reboot cause file used to detect NPU kernel crash on chassisd startup
 REBOOT_CAUSE_FILE = "/host/reboot-cause/reboot-cause.txt"
-NPU_CRASH_REBOOT_CAUSE_KEYWORDS = ("kernel panic", "unknown")
+NPU_CRASH_REBOOT_CAUSE_KEYWORDS = ("kernel panic")
 
 # Internal recovery dict key for tracking prior Unrecoverable state
 WAS_UNRECOVERABLE_KEY = 'was_unrecoverable'
@@ -1208,15 +1208,15 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
                 keys = self.chassis_state_db.keys(pattern)
                 if keys:
                     for key in keys:
-                        if DPU_STATE_TABLE not in key and not "REBOOT_CAUSE" in key:
+                        if DPU_STATE_TABLE not in key and "REBOOT_CAUSE" not in key:
                             self.chassis_state_db.delete(key)
         return
 
     def _is_auto_recovery_enabled(self):
         """Check if dpu_auto_recovery is enabled in CONFIG_DB DEVICE_METADATA|localhost.
 
-        Returns True when the field is 'enabled'.
-        Returns False when the field is 'disabled' (or absent, defaulting to disabled).
+        Returns True when the field is 'enable'.
+        Returns False otherwise.
         """
         try:
             fvs = self.device_metadata_table.get('localhost')
@@ -1225,7 +1225,8 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
                 state = fvs.get(DPU_AUTO_RECOVERY_FIELD, DPU_AUTO_RECOVERY_DISABLED)
                 return state == DPU_AUTO_RECOVERY_ENABLED
         except Exception as e:
-            self.log_error("Failed to read dpu_auto_recovery from DEVICE_METADATA: {}".format(e))
+            self.log_error("Failed to read {} from DEVICE_METADATA: {}".format(
+                DPU_AUTO_RECOVERY_FIELD, e))
         return False  # Default to disabled
 
     def _get_dpu_states(self, module_name):
@@ -1279,11 +1280,7 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
         return True
 
     def _enter_power_cycle_or_unrecoverable(self, module_name, module_index):
-        """Attempt to power-cycle a DPU, or mark it unrecoverable if reset_limit reached.
-
-        Checks reset_limit BEFORE issuing the power-cycle to avoid a
-        wasted reset that would immediately be marked unrecoverable.
-        """
+        """Attempt power-cycle; mark unrecoverable if reset_limit is reached."""
         recovery = self.dpu_recovery_state[module_name]
 
         # Check limit before issuing power-cycle
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -146,7 +146,6 @@ MINIMUM_SELF_RECOVERY_POLL_COUNT = 3
 
 # Reboot cause file used to detect NPU kernel crash on chassisd startup
 REBOOT_CAUSE_FILE = "/host/reboot-cause/reboot-cause.txt"
-NPU_CRASH_REBOOT_CAUSE_KEYWORDS = ("kernel panic")
 
 # Internal recovery dict key for tracking prior Unrecoverable state
 WAS_UNRECOVERABLE_KEY = 'was_unrecoverable'
@@ -1368,14 +1367,14 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
         recovery['self_recovery_poll_count'] = 0
 
     def _should_handle_hw_offline(self, current_state):
-        """Return True if the DPU should react to oper_status Offline.
+        """Return True only if a previously-Ready DPU should treat
+        oper_status Offline as a new hardware failure.
 
-        States already handling failure (Booting, PowerCycle,
-        WaitForSelfRecovery, ManualIntervention) should not re-trigger.
+        Allow-list (Ready only) is fail-safe for future states and avoids a
+        `module startup` race where a stale Offline snapshot in AdminDown
+        would power-cycle instead of cleanly restarting.
         """
-        return current_state not in (
-            DPU_STATE_BOOTING, DPU_STATE_POWER_CYCLE,
-            DPU_STATE_WAIT_FOR_SELF_RECOVERY, DPU_STATE_MANUAL_INTERVENTION)
+        return current_state == DPU_STATE_READY
 
     def _handle_boot_timeout_expired(self, name, module_index, recovery,
                                      auto_recovery_enabled, context):
@@ -1426,7 +1425,7 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
             self._enter_power_cycle_or_unrecoverable(name, module_index)
 
     def _npu_crash_on_last_boot(self):
-        """Return True if the last NPU reboot cause looks like a crash/OOM/unknown."""
+        """Return True if the last NPU reboot cause looks like a kernel crash."""
         try:
             if not os.path.isfile(REBOOT_CAUSE_FILE):
                 # No reboot-cause file means this is likely a fresh boot or the
@@ -1435,14 +1434,9 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
             with open(REBOOT_CAUSE_FILE, 'r') as f:
                 cause = f.read().strip().lower()
 
-            # Match explicit crash signals only. This avoids classifying
-            # normal operator-initiated reboot strings as crashes.
-            if 'kernel panic' in cause:
-                return True
-
-            # Keep supporting legacy "unknown" reboot-cause content, but only
-            # as the full value/prefix (not arbitrary substring matches).
-            return cause == 'unknown' or cause.startswith('unknown:') or cause.startswith('unknown ')
+            # Match kernel crashes only. "unknown" is reported on first boot
+            # and must not trigger a power-cycle.
+            return 'kernel panic' in cause
         except Exception as e:
             self.log_warning("Failed to read {}: {}".format(REBOOT_CAUSE_FILE, e))
             return False
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -18,6 +18,7 @@ try:
     import json
     import glob
     from datetime import datetime, timezone
+    from typing import NamedTuple
 
     from sonic_py_common import daemon_base, logger, device_info
     from sonic_py_common.task_base import ProcessTaskBase
@@ -105,13 +106,16 @@ MODULE_ADMIN_UP = 1
 MODULE_REBOOT_CAUSE_DIR = "/host/reboot-cause/module/"
 MAX_HISTORY_FILES = 10
 
+# DPU_STATE table in CHASSIS_STATE_DB
+DPU_STATE_TABLE = 'DPU_STATE'
+
 DP_STATE = 'dpu_data_plane_state'
 DP_UPDATE_TIME = 'dpu_data_plane_time'
 CP_STATE = 'dpu_control_plane_state'
 CP_UPDATE_TIME = 'dpu_control_plane_time'
 MP_STATE = 'dpu_midplane_link_state'
 
-# New DPU recovery DB fields (CHASSIS_STATE_DB DPU_STATE table)
+# DPU recovery DB fields (CHASSIS_STATE_DB DPU_STATE table)
 READY_STATUS = 'ready_status'
 RECOVERY_STATUS = 'recovery_status'
 RESET_COUNT = 'reset_count'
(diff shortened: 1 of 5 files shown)


sonic-chassisd/scripts/chassisd
```diff
@@ -1240,45 +1256,127 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+                            "after lock acquisition — aborting power-cycle".format(module_name))
+            try_get(module.clear_module_state_transition, module_name, default=False)
+            recovery['reset_count'] -= 1
+            self._set_reset_count(module_name, recovery['reset_count'])
+            return
 
         try:
-            module = self.chassis.get_module(module_index)
+            try_get(module.pci_detach, default=False)
             try_get(module.set_admin_state, MODULE_ADMIN_DOWN, default=False)
```
gpunathilell, line 1324: @vvolam when will this be modified to graceful usage?

vvolam: @gpunathilell Use set_admin_state_gracefully() blocks the call for 3 mins and does not make sense here to wait for a dead DPU. It will also have a race, with a planned operations skip in SmartSwitchModuleUpdater. Please let me know your thoughts?

chartsai-nvidia:
Hi Vasundhara

When the reboot cause is unknown, the log still reports a kernel crash as the reboot cause.
Please update the logic so that an unknown reboot cause is reported correctly instead of being classified as a kernel crash. Thanks

vvolam:
> Hi Vasundhara
> 
> When the reboot cause is unknown, the log still reports a kernel crash as the reboot cause. Please update the logic so that an unknown reboot cause is reported correctly instead of being classified as a kernel crash. Thanks

@chartsai-nvidia for 4280 sku, nvidia is reporting unknown for kernel crashes and unable to distinguish. Once it is fixed, we could tighten the chassisd code.

chartsai-nvidia:
> > Hi Vasundhara
> > When the reboot cause is unknown, the log still reports a kernel crash as the reboot cause. Please update the logic so that an unknown reboot cause is reported correctly instead of being classified as a kernel crash. Thanks
> 
> @chartsai-nvidia for 4280 sku, nvidia is reporting unknown for kernel crashes and unable to distinguish. Once it is fixed, we could tighten the chassisd code.

I understand, at this time we also report unknown reboot cause for first boot, and it would cause additional unwanted power cycle for DPU on each first boot. This will break test flows. Until there's a clarification on the reboot cause we need to fix this issue.

sonic-chassisd/scripts/chassisd
```diff
@@ -1188,37 +1208,33 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
                 keys = self.chassis_state_db.keys(pattern)
                 if keys:
                     for key in keys:
-                        if not "DPU_STATE" in key and not "REBOOT_CAUSE" in key:
+                        if DPU_STATE_TABLE not in key and not "REBOOT_CAUSE" in key:
```
judsonwilson-nvidia:
```suggestion
                        if DPU_STATE_TABLE not in key and "REBOOT_CAUSE" not in key:
```

sonic-chassisd/scripts/chassisd
```diff
@@ -1188,37 +1208,33 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
     def _is_auto_recovery_enabled(self):
-        """Check if dpu-auto-recovery feature is enabled in CONFIG_DB FEATURE table.
+        """Check if dpu_auto_recovery is enabled in CONFIG_DB DEVICE_METADATA|localhost.
 
-        Returns True only when ``state == 'enabled'``. Both ``'disabled'`` and
-        ``'always_disabled'`` suppress automatic power-cycle recovery; chassisd
-        still updates DPU state fields in CHASSIS_STATE_DB but will not
-        initiate a power-cycle.
+        Returns True when the field is 'enabled'.
+        Returns False when the field is 'disabled' (or absent, defaulting to disabled).
```
judsonwilson-nvidia:
```suggestion
        Returns False otherwise.
```

sonic-chassisd/scripts/chassisd
```diff
@@ -1188,37 +1208,33 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            fvs = self.device_metadata_table.get('localhost')
             if isinstance(fvs, list) and fvs[0] is True:
                 fvs = dict(fvs[-1])
-                state = fvs.get('state', DPU_AUTO_RECOVERY_STATE_ENABLED)
-                return state == DPU_AUTO_RECOVERY_STATE_ENABLED
+                state = fvs.get(DPU_AUTO_RECOVERY_FIELD, DPU_AUTO_RECOVERY_DISABLED)
+                return state == DPU_AUTO_RECOVERY_ENABLED
         except Exception as e:
-            self.log_error("Failed to read dpu-auto-recovery feature state: {}".format(e))
+            self.log_error("Failed to read dpu_auto_recovery from DEVICE_METADATA: {}".format(e))
```
judsonwilson-nvidia: Nit: use the DPU_AUTO_RECOVERY_FIELD variable in the error message instead of "dpu_auto_recovery"?

sonic-chassisd/scripts/chassisd
```diff
@@ -1240,45 +1256,134 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
-            self.log_warning("{}: Failed to read state_transition_in_progress: {}".format(
-                module_name, e))
-        return False
+                return fvs.get('transition_type') != 'recovery'
+        except Exception:
+            pass
+        return True
+
+    def _enter_power_cycle_or_unrecoverable(self, module_name, module_index):
+        """Attempt to power-cycle a DPU, or mark it unrecoverable if reset_limit reached.
```
judsonwilson-nvidia:
```suggestion
```

sonic-chassisd/scripts/chassisd
```diff
@@ -1240,45 +1256,134 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
 
-    def _power_cycle_dpu(self, module_name, module_index):
-        """Power-cycle a DPU and increment reset_count."""
+        Checks reset_limit BEFORE issuing the power-cycle to avoid a
+        wasted reset that would immediately be marked unrecoverable.
```
judsonwilson-nvidia:
```suggestion
```

Delete this AI over-chatty-ness.

sonic-chassisd/scripts/chassisd
```diff
@@ -1240,45 +1256,134 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+        # Re-check transition_in_progress after acquiring our lock to close
+        # the TOCTOU window where gnoi/CLI sets transition_in_progress
+        # between _is_planned_transition_in_progress() and here.
+        if self._is_planned_transition_in_progress(module_name, module_index):
+            self.log_warning("{}: transition_in_progress detected "
+                            "after lock acquisition — aborting power-cycle".format(module_name))
+            try_get(module.clear_module_state_transition, module_name, default=False)
+            recovery['reset_count'] -= 1
+            self._set_reset_count(module_name, recovery['reset_count'])
+            return
```
judsonwilson-nvidia: Will this happen?  I see that set_module_state_transition uses a file lock. I think in a correct implementation all code that writes this database key should use the same file lock (I don't know if they actually do).

sonic-chassisd/scripts/chassisd
```diff
@@ -1240,45 +1256,134 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
-        if recovery['reset_count'] >= self.reset_limit:
-            recovery['state'] = DPU_STATE_UNRECOVERABLE
-            self._set_recovery_status(module_name, RECOVERY_UNRECOVERABLE)
-            self.log_error("{}: Reached reset_limit ({}). Marking as unrecoverable.".format(
-                module_name, self.reset_limit))
+        finally:
+            try_get(module.clear_module_state_transition, module_name, default=False)
+
+        recovery['state'] = DPU_STATE_POWER_CYCLE
+        recovery['boot_start_time'] = time.time()
```
judsonwilson-nvidia, line 1335: Should we set this state earlier before early-return paths? I'm afraid what happens if there is a state transition in progress, etc, and return early without updating the state. It might be simplest to understand if it goes to the POWER_CYCLE state anyway, even if it didn't power cycle.  (This state is more about waiting for recovery, as far as I can tell.)

I see the following cases:

transition is coming up or a reset:  Then being in this state is an acceptable way to wait for full recovery.
transition is going down: Eventually the admin-down catcher will handle this at the beginning of the state machine logic function.
anything else: this state will wait for the DPU to come up, it probably won't happen, and will retry a reset later.

I haven't fully thought through the implications for any counters/timeouts.

sonic-chassisd/scripts/chassisd
```diff
@@ -1240,45 +1256,134 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            DPU_STATE_BOOTING, DPU_STATE_POWER_CYCLE,
+            DPU_STATE_WAIT_FOR_SELF_RECOVERY, DPU_STATE_MANUAL_INTERVENTION)
+
+    def _handle_boot_timeout_expired(self, name, module_index, recovery,
+                                     auto_recovery_enabled, context):
+        """Handle boot timeout expiry — power-cycle or manual intervention."""
+        self.log_warning("{}: Boot timeout ({}s) expired {}".format(
+            name, self.dpu_boot_timeout, context))
+        self._set_ready_status(name, 'false')
+        self._set_last_down_time(name)
```
judsonwilson-nvidia: Just calling this out to double check: do we want to set the last down time every retry of a reset? Or do we want it to be set only when initially detected and not update it while trying to reset?

To put it another way, would we only set it while leaving the ready state? (And maybe some other corner case like startup...?)

vvolam:
> > > Hi Vasundhara
> > > When the reboot cause is unknown, the log still reports a kernel crash as the reboot cause. Please update the logic so that an unknown reboot cause is reported correctly instead of being classified as a kernel crash. Thanks
> > 
> > 
> > @chartsai-nvidia for 4280 sku, nvidia is reporting unknown for kernel crashes and unable to distinguish. Once it is fixed, we could tighten the chassisd code.
> 
> I understand, at this time we also report unknown reboot cause for first boot, and it would cause additional unwanted power cycle for DPU on each first boot. This will break test flows. Until there's a clarification on the reboot cause we need to fix this issue.

@chartsai-nvidia Removed "unknown" reboot cause.

sonic-chassisd/scripts/chassisd
vvolam: Yes, set/clear/get all calls use the same lock which is defined here https://github.com/sonic-net/sonic-platform-common/blob/b76791d551a20eb118bd1c4ec3791c4e8539dda4/sonic_platform_base/module_base.py#L31

sonic-chassisd/scripts/chassisd
vvolam: You are right, better to keep initial down_time. Made the changes and added unittests accordingly.

sonic-chassisd/scripts/chassisd
```diff
@@ -1340,189 +1442,178 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
-            recovery = self.dpu_recovery_state[name]
-            current_state = recovery['state']
-            mp_state, cp_state, dp_state = self._get_dpu_states(name)
-            oper_status = self.get_module_current_status(name)
-            admin_status = self.get_module_admin_status(name)
+        # Terminal: requires operator to set admin-down then admin-up,
+        # which triggers the admin-down check above and transitions to
+        # AdminDown state, clearing the Unrecoverable condition.
+        if current_state == DPU_STATE_UNRECOVERABLE:
+            return
```
judsonwilson-nvidia, line 1477: Allow a transition to READY if all_up? This might be a nice way to handle the user issuing reboot without relying on the admin state to go down in that case?

sonic-chassisd/scripts/chassisd
```diff
@@ -1240,45 +1257,129 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+            try_get(module.clear_module_state_transition, module_name, default=False)
+
+        recovery['state'] = DPU_STATE_POWER_CYCLE
+        recovery['boot_start_time'] = time.time()
+
+    def _enter_booting(self, name, recovery):
+        """Transition a DPU into the Booting state."""
+        recovery['state'] = DPU_STATE_BOOTING
+        recovery['boot_start_time'] = time.time()
+        recovery.pop('boot_dp_warning_logged', None)
```
judsonwilson-nvidia, line 1341: This string literal is repeated on ~3 lines. Would be good to turn this into a CONSTANT

sonic-chassisd/scripts/chassisd
```diff
@@ -1340,189 +1442,178 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
-                        if not recovery.get('boot_dp_warning_logged'):
-                            self.log_warning("{}: data plane not up after {}s".format(
-                                name, self.dpu_boot_timeout))
-                            recovery['boot_dp_warning_logged'] = True
-                    else:
-                        # CP or midplane still down — genuine boot failure.
-                        self.log_warning("{}: Boot timeout ({}s) expired without reaching Ready".format(
-                            name, self.dpu_boot_timeout))
-                        self._set_ready_status(name, 'false')
-                        self._set_last_down_time(name)
```
judsonwilson-nvidia, line 1433: I see _set_last_down_time is no longer here. Not sure if that was intended or not.

sonic-chassisd/scripts/chassisd
vvolam: Done. I replaced the repeated recovery transition-type string with a dedicated constant, TRANSITION_TYPE_RECOVERY, and updated usages to reference it for consistency and maintainability.

sonic-chassisd/scripts/chassisd
vvolam: Thanks, good call. I updated the logic so both early-abort paths in _enter_power_cycle_or_unrecoverable (lock not acquired and transition detected after lock acquisition) now move the FSM into PowerCycle and set boot_start_time, while still rolling back reset_count because no reset was actually issued. This lets us wait for recovery and avoids repeatedly re-entering the same path each poll cycle.

sonic-chassisd/scripts/chassisd
vvolam: Agreed. This is now covered in the PowerCycle handler: when all planes are up, we transition directly to Ready. With the early-abort path now entering PowerCycle as well, this also handles user-issued reboot/shutdown completion cleanly without depending only on admin status transitions.

sonic-chassisd/scripts/chassisd
vvolam: Great question. This was intentional: we preserve the initial outage timestamp across retries instead of refreshing it on each boot-timeout retry. The goal is to keep the original failure start time for observability and avoid masking outage duration by repeated retries.

chartsai-nvidia:
> > > > Hi Vasundhara
> > > > When the reboot cause is unknown, the log still reports a kernel crash as the reboot cause. Please update the logic so that an unknown reboot cause is reported correctly instead of being classified as a kernel crash. Thanks
> > > 
> > > 
> > > @chartsai-nvidia for 4280 sku, nvidia is reporting unknown for kernel crashes and unable to distinguish. Once it is fixed, we could tighten the chassisd code.
> > 
> > 
> > I understand, at this time we also report unknown reboot cause for first boot, and it would cause additional unwanted power cycle for DPU on each first boot. This will break test flows. Until there's a clarification on the reboot cause we need to fix this issue.
> 
> @chartsai-nvidia Removed "unknown" reboot cause.

Thank you, Vasundhara.

Just wanted to confirm my understanding. With the "unknown" removed, I assume kdump will also be enabled and functioning properly so that kernel crashes can still be captured and analyzed.

Otherwise, I'm concerned that some crashes may no longer be reported as "unknown" while also not generating a kdump for investigation.

Could you please confirm?

Thanks,
Charles

vvolam:
> > > > > Hi Vasundhara
> > > > > When the reboot cause is unknown, the log still reports a kernel crash as the reboot cause. Please update the logic so that an unknown reboot cause is reported correctly instead of being classified as a kernel crash. Thanks
> > > > 
> > > > 
> > > > @chartsai-nvidia for 4280 sku, nvidia is reporting unknown for kernel crashes and unable to distinguish. Once it is fixed, we could tighten the chassisd code.
> > > 
> > > 
> > > I understand, at this time we also report unknown reboot cause for first boot, and it would cause additional unwanted power cycle for DPU on each first boot. This will break test flows. Until there's a clarification on the reboot cause we need to fix this issue.
> > 
> > 
> > @chartsai-nvidia Removed "unknown" reboot cause.
> 
> Thank you, Vasundhara.
> 
> Just wanted to confirm my understanding. With the "unknown" removed, I assume kdump will also be enabled and functioning properly so that kernel crashes can still be captured and analyzed.
> 
> Otherwise, I'm concerned that some crashes may no longer be reported as "unknown" while also not generating a kdump for investigation.
> 
> Could you please confirm?
> 
> Thanks, Charles

@chartsai-nvidia With unknown removed, we expect platform reboot cause is properly reported as kernel panic. Thanks!

sonic-chassisd/scripts/chassisd
gpunathilell: @vvolam Re: the ~3 minute delay flagged earlier — I dug into the root cause and filed sonic-net/sonic-buildimage#28093.

The short version: set_admin_state_gracefully() → _graceful_shutdown_handler() sets gnoi_halt_in_progress and then waits for gnoi-shutdown-daemon to clear it, up to the halt_services timeout. But that daemon only triggers on a CONFIG_DB CHASSIS_MODULE|admin_status change — it doesn't watch the flag. So any shutdown that doesn't go through a CONFIG_DB admin_status change (boot-time power-off, and a recovery power-cycle) never wakes the daemon, the gNOI HALT never runs, and the handler dead-waits the full timeout before falling back to a hard power-off. On sn4280 that timeout is overridden to 180s, which is exactly the ~3 min we're seeing.

Given that, I'd still push to route this through set_admin_state_gracefully() rather than a separate hand-rolled power-cycle path. It keeps admin-state changes in one place, so any fix (including #28093, after which the graceful path stops dead-waiting and actually performs the gNOI HALT) lands once and benefits every caller, instead of us maintaining and re-fixing the same lock + HALT + PCIe detach/reattach sequence in multiple spots. Does that make sense?

sonic-chassisd/scripts/chassisd
```diff
@@ -137,21 +146,31 @@ MINIMUM_SELF_RECOVERY_POLL_COUNT = 3
 
 # Reboot cause file used to detect NPU kernel crash on chassisd startup
 REBOOT_CAUSE_FILE = "/host/reboot-cause/reboot-cause.txt"
-NPU_CRASH_REBOOT_CAUSE_KEYWORDS = ("kernel panic", "unknown")
+NPU_CRASH_REBOOT_CAUSE_KEYWORDS = ("kernel panic")
```
gpunathilell: Variable is unused, (and type of variable is string right now, not the tuple type previously used)

sonic-chassisd/scripts/chassisd
```diff
@@ -1300,32 +1408,41 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
-            return any(kw in cause for kw in NPU_CRASH_REBOOT_CAUSE_KEYWORDS)
+
+            # Match explicit crash signals only. This avoids classifying
+            # normal operator-initiated reboot strings as crashes.
+            if 'kernel panic' in cause:
+                return True
+
+            # Keep supporting legacy "unknown" reboot-cause content, but only
+            # as the full value/prefix (not arbitrary substring matches).
+            return cause == 'unknown' or cause.startswith('unknown:') or cause.startswith('unknown ')
```
gpunathilell: @vvolam This needs to be removed right?

sonic-chassisd/scripts/chassisd
```diff
@@ -1240,45 +1260,133 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+        recovery['self_recovery_start_time'] = time.time()
+        recovery['self_recovery_poll_count'] = 0
+
+    def _should_handle_hw_offline(self, current_state):
+        """Return True if the DPU should react to oper_status Offline.
+
+        States already handling failure (Booting, PowerCycle,
+        WaitForSelfRecovery, ManualIntervention) should not re-trigger.
+        """
+        return current_state not in (
```
gpunathilell: @vvolam can you explain all the states which would trigger hw_offline here?
and an Allow-list is clearer and fail-safe for future states instead of a deny list please confirm if admin down is also part of this list

sonic-chassisd/scripts/chassisd
```diff
@@ -1340,189 +1457,178 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
-            if admin_status == 'down':
-                if current_state != DPU_STATE_OFFLINE:
-                    if current_state == DPU_STATE_UNRECOVERABLE:
-                        recovery['was_unrecoverable'] = True
-                    recovery['state'] = DPU_STATE_OFFLINE
-                    self._set_ready_status(name, 'false')
-                continue
+        # Admin-up DPU went Offline at the hardware level — treat as a
+        # hardware failure.
+        if oper_status == str(ModuleBase.MODULE_STATUS_OFFLINE) and \
```
gpunathilell, line 1487: @vvolam is this a valid race condition?J
In _process_single_dpu_recovery(), the hw-offline check (guard 4) runs before the AdminDown → Booting restart (guard 5), and _should_handle_hw_offline(AdminDown) returns True.

On config chassis-modules startup, the "startup" lock is held through the (blocking) set_admin_state(True), so guard 3 suppresses recovery during boot. But the FSM reads oper status from the STATE_DB snapshot refreshed by module_db_update() at the top of the loop, not live — while set_admin_state runs in a separate thread. So there's a narrow window: module_db_update() writes Offline mid-boot, the worker thread then clears the lock, and the FSM runs with guard 3 clear + a stale Offline → guard 4 power-cycles (and bumps reset_count) instead of going to Booting. Low probability, but it relies on blocking set_admin_state semantics that aren't in the ModuleBase contract.

Suggest making it timing-independent: check guard 5 before guard 4, or tighten _should_handle_hw_offline() to return current_state == DPU_STATE_READY so AdminDown falls through to the clean AdminDown → Booting path.

sonic-chassisd/scripts/chassisd
vvolam: Removed it in latest commit

sonic-chassisd/scripts/chassisd
vvolam: Removed it in latest commit

sonic-chassisd/scripts/chassisd
vvolam: you are right, fixed it.

sonic-chassisd/scripts/chassisd
vvolam: Good catch — yes, it's a valid race given `set_admin_state` isn't contractually blocking. Fixed in ab47050 by tightening `_should_handle_hw_offline()` to `current_state == DPU_STATE_READY`, so a stale `Offline` snapshot while still `AdminDown` falls through to the clean `AdminDown → Booting` path instead of power-cycling. Added `test_hardware_offline_from_admin_down_restarts_booting` to cover it (asserts `reset_count` unchanged).

vvolam:
@gpunathilell I have addressed the comments in the latest changes.

sonic-chassisd/scripts/chassisd
```diff
@@ -1240,45 +1259,133 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+        # between _is_planned_transition_in_progress() and here.
+        if self._is_planned_transition_in_progress(module_name, module_index):
+            self.log_warning("{}: transition_in_progress detected "
+                            "after lock acquisition — aborting power-cycle".format(module_name))
+            try_get(module.clear_module_state_transition, module_name, default=False)
+            recovery['reset_count'] -= 1
+            self._set_reset_count(module_name, recovery['reset_count'])
+            recovery['state'] = DPU_STATE_POWER_CYCLE
+            recovery['boot_start_time'] = time.time()
+            return
```
gpunathilell: Can you walk me through the TOCTOU this re-check is meant to close?

`set_module_state_transition()` in `module_base` is an atomic compare-and-acquire: it takes the `_transition_operation_lock` flock and does the `transition_in_progress` read-modify-write inside it. So `lock_acquired = set_module_state_transition(...)` above already *is* the check — if a gnoi shutdown/reboot raced us, it either grabbed the flag first (our call returns `False` -> we abort in the block above) or we grabbed it first (their call returns `False`). Per contract every actor acquires via `transition_in_progress` before proceeding, so both paths go through the same flock + same DB flag — I don't see a window left between the acquire and this re-check for anyone to slip in.

On top of that, once we've acquired, `transition_in_progress=True` and `transition_type='recovery'`, and `_is_planned_transition_in_progress()` returns `False` for `'recovery'` — so this condition can never be True on this path, and the abort body under it is unreachable. The unit test only reaches that body by mocking `_is_planned_transition_in_progress -> True`, so it isn't exercising a real scenario.

I think the acquire above is sufficient and this re-check can be dropped.

sonic-chassisd/scripts/chassisd
```diff
@@ -1340,189 +1451,178 @@ class SmartSwitchModuleUpdater(ModuleUpdater):
...
+                    if not recovery.get('boot_dp_warning_logged'):
+                        self.log_warning("{}: data plane not up after {}s".format(
+                            name, self.dpu_boot_timeout))
+                        recovery['boot_dp_warning_logged'] = True
+                else:
+                    # CP or midplane still down — genuine boot failure.
+                    self._handle_boot_timeout_expired(
+                        name, module_index, recovery,
+                        auto_recovery_enabled, "without reaching Ready")
+            return
```
gpunathilell, line 1531: In the `BOOTING` handler, when the boot timeout expires with `cp` and `mp` up but `dp` still down, we only log a warning and stay in `BOOTING` — we never transition to `READY`.

Is it expected that a DPU whose `dp_state` never comes up stays in `BOOTING` indefinitely (with `ready_status` pinned `false`)? The `READY` handler tolerates `dp_state == 'down'` and keeps the DPU Ready, so a DPU that converges once and then loses dp stays Ready, but one that never brings dp up on first boot is wedged in `BOOTING`. Wanted to confirm that asymmetry is intentional and not a case we should promote to `READY` on `cp`+`mp` up.

vvolam: @gpunathilell This is intentional, as we don't want to kick-off recovery for DP state down.

sonic-chassisd/scripts/chassisd
vvolam: Good catch — you're right, and I've dropped it in e98f98d.

Walking through it once more to confirm: `set_module_state_transition()` is an atomic compare-and-acquire — the `transition_in_progress` read/modify/write runs inside `_transition_operation_lock()`, which is an `fcntl.flock(LOCK_EX)` on `/var/lock/{module}_transition.lock`, so it serializes across processes (chassisd / CLI / reboot helper), not just threads. And every writer goes through that API — there are no direct DB writes of `transition_in_progress`/`transition_type` outside `module_base.py` (the reboot/shutdown helpers go through `ModuleHelper` → the platform API, plus module_base's own graceful admin-state path; the gnoi-shutdown-daemon doesn't touch the flag at all).

So the acquire itself closes the window: if a planned op raced us between the earlier `_is_planned_transition_in_progress()` gate and the acquire, either they won and our `set_module_state_transition('recovery')` returns `False` (we abort in `if not lock_acquired`), or we won and theirs returns `False`. Once we hold the lock `transition_type == 'recovery'`, so `_is_planned_transition_in_progress()` returns `False` by construction — the re-check could only fire when mocked.

Removed the re-check block and the `test_power_cycle_aborted_when_legacy_field_set_after_lock` test that only reached it via mocking; kept the earlier suppression gate and the `if not lock_acquired` abort. Thanks!

sonic-chassisd/scripts/chassisd
gpunathilell: so it is expected to stay booting? assuming dp never comes up?

gpunathilell:
Hi @vvolam, I think I only have two points, everything else looks good, I want to use graceful shutdown/startup use since that automatically covers the transition lock etc, so Please update that once we have fix for this command: https://github.com/sonic-net/sonic-platform-daemons/pull/829/changes#r3475789446

vvolam:
> Hi @vvolam, I think I only have two points, everything else looks good, I want to use graceful shutdown/startup use since that automatically covers the transition lock etc, so Please update that once we have fix for this command: https://github.com/sonic-net/sonic-platform-daemons/pull/829/changes#r3475789446

Sure, we could modify the path once we have the fix.

sonic-chassisd/scripts/chassisd
vvolam: "yes, it stays in the `Booting `state as long as dp stays down and `ready_state` will be False.

gpunathilell (approved):

vaibhavhd:
202605 testing results are not updated

vvolam:
> 202605 testing results are not updated

@vaibhavhd Tests ran on both master and 202605 images on a smartswitch. Updated the PR description with the same.

vaibhavhd:
> Both remaining failures were environmental/test-harness — a databasedpu container inactive at test start, and a config-reload case that ran on pre-fix test code — not auto-recovery regressions.

What about this? The failures likely are due to setup/state of tb, but that also means that these cases cannot be run and change is not evaluated.

vaibhavhd:
Are the failing cases not relevant to the change being made here? Do you want to rerun to ensure you get a green run for these cases?

vvolam:
> Are the failing cases not relevant to the change being made here? Do you want to rerun to ensure you get a green run for these cases?

Failure cases were because of PSU not ok on the DPUs which is fixed by [[system-health]: Honor 'psu' and 'pdb' device ignore independently in hardware checker by vvolam · …](https://github.com/sonic-net/sonic-buildimage/pull/28415) and this makes all tests run clean.

PR #831, opened 2026-06-05
[xcvrd] Don't crash when no physical port is found

### LINKED ISSUE #832: xcvrd can crash if no physical port is found for a logical port
If we fail to find a physical port for a given logical port in `_init_port_sfp_status_sw_tbl`, `xcvrd` will crash:
```
    def _init_port_sfp_status_sw_tbl(self, port_mapping, xcvr_table_helper, stop_event=threading.Event()):
        ...
        for logical_port_name in logical_port_list:
            ...
            physical_port_list = port_mapping.logical_port_name_to_physical_port_list(logical_port_name)
            if physical_port_list is None:
                helper_logger.log_error("No physical ports found for logical port '{}' during sfp status table init".format(logical_port_name))
                common.update_port_transceiver_status_table_sw(logical_port_name, xcvr_table_helper.get_status_sw_tbl(asic_index), sfp_status_helper.SFP_STATUS_REMOVED)

            for physical_port in physical_port_list: # We will crash here if physical_port_list is None
```

bgallagher-nexthop:
Fixes #832.

#### Description
If no physical port is found for a logical port, `physical_port_list` will be `None`. `xcvrd` marks the transceiver as removed in this case, but does not skip to the next logical port and instead attempts to execute `for physical_port in physical_port_list`, which will crash since `None` is not iterable.

This PR fixes the issue by just simply continuing to the next logical port after marking the transceiver removed.

#### Motivation and Context
This change prevents `xcvrd` from crashing when encountering this bug.

A failing unit-test was written and then made pass with the change in this PR.

### DIFF
--- a/sonic-xcvrd/tests/test_xcvrd.py
+++ b/sonic-xcvrd/tests/test_xcvrd.py
@@ -1535,6 +1535,31 @@ class TestXcvrdScript(object):
         task = SfpStateUpdateTask(DEFAULT_NAMESPACE, port_mapping, mock_sfp_obj_dict, stop_event, sfp_error_event)
         task._init_port_sfp_status_sw_tbl(port_mapping, xcvr_table_helper, stop_event)
 
+    @patch('xcvrd.xcvrd_utilities.common._wrapper_get_presence', MagicMock(return_value=True))
+    @patch('xcvrd.xcvrd.XcvrTableHelper', MagicMock())
+    @patch('xcvrd.xcvrd.common.update_port_transceiver_status_table_sw')
+    def test_init_port_sfp_status_sw_tbl_no_physical_port_found(self, mock_update_status):
+        """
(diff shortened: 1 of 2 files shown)


mihirpat1:
@bgallagher-nexthop Thanks for the fix. Could you share the scenario where a logical port exists, but logical_port_name_to_physical_port_list() returns None?

bgallagher-nexthop:
> @bgallagher-nexthop Thanks for the fix. Could you share the scenario where a logical port exists, but logical_port_name_to_physical_port_list() returns None?

I haven't encountered this actual failure in the practice, I just noticed this code was missing a continue and decided to make the fix while I was there. It doesn't seem to be a failure mode that is likely to occur, but the fix is simple enough that the change seemed worth it to me.

mihirpat1 (approved):

prgeor (approved):

PR #836, opened 2026-06-15
[ycabled]: Add IPv6 only active-active SoC gRPC

zjswhhh:
## Why I did it

Active-active DualToR IPv6 only configurations can omit `soc_ipv4`. ycabled should continue to prefer IPv4 for dual-stack configurations, but connect to the SoC over IPv6 when IPv4 is absent.

## How I did it

- Prefer `soc_ipv4` when present.
- Fall back to `soc_ipv6` only when `soc_ipv4` is missing.
- Strip CIDR notation before building the gRPC target.
- Format IPv6 gRPC targets as `[addr]:port` for synchronous and asynchronous channels.
- Add ycabled tests for IPv4 preference, IPv6 fallback, and IPv6 target formatting.

Related native NIC simulator support used for validation: https://github.com/sonic-net/sonic-mgmt/pull/27812.

### Completed validation

- **Coverage follow-up:** added 19 test cases for malformed/missing addresses, IPv6 retry/startup, and synchronous/asynchronous secure/insecure channel construction. Local full-suite result: **296 passed**; helper PR diff coverage: **100% (35/35 changed executable lines)**; helper total line coverage: **80.6%**. Local validation used a shim for unavailable native swsscommon bindings and a harness that exits after pytest reports because legacy daemon tests leave threads alive; neither is included in the PR. CI confirmation is pending.
- **Running-daemon end-to-end validation passed on both ToRs**, using a NIC simulator with native IPv6 support. No proxy, synthetic RPC responses, or substituted connectivity callbacks were used.
- With both SoC address fields configured, ycabled selected IPv4 and reached `READY`.
- With `soc_ipv4` absent and `soc_ipv6` configured, the running ycabled process owned established IPv6 connections to the native simulator. Channel callbacks and Redis reported `READY`.
- Real `QueryAdminForwardingPortState` and `SetAdminForwardingPortState` RPCs succeeded over IPv6, including actual simulator forwarding-state updates.
- Stopping and starting the simulated NIC listeners caused `TRANSIENT_FAILURE` followed by automatic recovery to `READY`, without restarting either ycabled daemon. Successful IPv6 RPCs resumed afterward.
- Separate synchronous and asynchronous feature-client readiness/RPC checks also passed.
- Original daemon code and configuration were restored after testing; IPv4 connectivity and active forwarding state were verified.

### Scope

End-to-end validation covered one active-active port from both ToRs. It does not claim full IPv6-only link-prober/HA validation, all-port scale, TLS/mTLS, or long-duration stability. The address-family configuration change was followed by a ycabled restart; hot switching without a restart was not tested. The focused unit-test result is not a claim that the entire ycabled suite passed.

### DIFF
--- a/sonic-ycabled/tests/test_ycable.py
+++ b/sonic-ycabled/tests/test_ycable.py
@@ -440,6 +440,29 @@ class TestYcableAsyncScript(object):
 
 class TestYcableActiveActiveHelper(object):
 
+    def test_get_soc_ip_for_grpc_prefers_ipv4(self):
+        soc_ip, soc_ip_field = get_soc_ip_for_grpc({
+            'soc_ipv4': '192.168.0.1/32',
+            'soc_ipv6': 'fc02:1000::1/128'
+        })
+
+        assert(soc_ip == '192.168.0.1')
+        assert(soc_ip_field == 'soc_ipv4')
+
+    def test_get_soc_ip_for_grpc_falls_back_to_ipv6(self):
+        soc_ip, soc_ip_field = get_soc_ip_for_grpc({
+            'soc_ipv6': 'fc02:1000::1/128'
+        })
+
+        assert(soc_ip == 'fc02:1000::1')
+        assert(soc_ip_field == 'soc_ipv6')
+
+    def test_format_grpc_target_for_ipv6(self):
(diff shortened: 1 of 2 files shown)


zjswhhh:
Hi @Ndancejic @lolyu - please help review

PR #844, opened 2026-07-01
[BMC] Fix bmcctld startup race and honor admin_status on air-cooled boot

nats-nokia:
Fix bmcctld startup race and honor admin_status on air-cooled boot

#### Description
1) sync HOST_STATE and seed CHASSIS_MODULE admin_status dedup before starting the BmcEventHandler thread;
add BmcEventHandler.seed_chassis_module_admin_status() so CONFIG SubscriberStateTable replays are treated as already handled. 
2) For non-liquid-cooled platforms, gate automatic startup power_on on CONFIG admin_status=up and skip power_on if admin_status=down. Update run() docstring and add unit tests for seed/dedup, startup ordering and admin-down skip paths.

#### Motivation and Context
1) bmcctld start can enqueue ACTION_POWER_ON from CONFIG replay before HOST_STATE init and action-loop readiness; CLI changes are logged as "initiating ..." but may never reach "Executing action ...". With a single action-loop consumer, an early or stuck power_on blocks subsequent CLI-driven actions until bmcctld restart.
2) Also fix air-cooled boot ignoring if CONFIG_DB has admin_status=down.

 run() order is init_host_state() -> seed_chassis_module_admin_status() ->
event thread -> boot sequence -> action loop. Dedup seed records the current CONFIG admin_status so replay does not enqueue ACTION_POWER_ON. Non-liquid OFFLINE boot checks get_switch_host_admin_status() and skips power_on unless admin is up.

New unit tests cover CONFIG replay dedup after seed, HOST_STATE init before event thread, non-liquid skip when admin_status=down, and non-liquid power_on when admin_status=up. Nokia H6-128 BMC verified: first-install CLI shutdown/startup works without pmon restart; BMC reboot with admin down leaves Switch-Host powered off.

### DIFF
--- a/sonic-bmcctld/scripts/bmcctld
+++ b/sonic-bmcctld/scripts/bmcctld
@@ -855,6 +855,22 @@ class BmcEventHandler(EventLogger):
         self.sel.addSelectable(self.rack_mgr_alert_sub)
         self.sel.addSelectable(self.system_leak_sub)
 
+    def seed_chassis_module_admin_status(self):
+        """
+        Record the current CONFIG admin_status before the event loop starts.
+
+        SubscriberStateTable replays existing CHASSIS_MODULE rows on subscribe. Without
+        seeding, the replay can look like a new admin transition and enqueue
+        ACTION_POWER_ON before HOST_STATE is synchronized and before the main-thread
+        action loop is ready to process the queue.
+        """
+        admin_status = self.policy_reader._get_chassis_module_entry().get(FIELD_ADMIN_STATUS, "")
+        if admin_status:
+            self._last_chassis_module_admin_status[SWITCH_HOST_MODULE_KEY] = admin_status
+            self.log_info(
+                "Seeded CHASSIS_MODULE admin_status dedup for {}: {}".format(
+                    SWITCH_HOST_MODULE_KEY, admin_status))
+
     # ---- public ----
 
     def run_event_loop(self):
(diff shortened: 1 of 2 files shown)


nats-nokia:
@judyjoseph
We found an issue with latest bmcctl that CLI "config chassis modules <shutdown/startup> SWITCH-HOST" events are not captured by Eventhandler untill PMON is restarted once after fresh onie-nos-install. 

Also, in air-cooled systems, if switch-host is shutdown and saved in DB, this commit skips power-on the switch-host on startup. 

Could you please review the changes?

Error case Logs:
2026 Apr 13 19:51:57.282641 h8-128-taj-bmc NOTICE pmon#bmcctld: bmcctld starting up
2026 Apr 13 19:51:57.286578 h8-128-taj-bmc NOTICE pmon#bmcctld: STARTUP: Switch-Host already ONLINE (daemon restart); skipping boot sequence
2026 Apr 13 19:51:57.287018 h8-128-taj-bmc INFO pmon#bmcctld.syslog[36]: Entering event subscription loop
2026 Apr 13 19:51:57.291825 h8-128-taj-bmc NOTICE pmon#bmcctld: CHASSIS_MODULE change: key=SWITCH-HOST admin_status=up
2026 Apr 13 19:51:57.292589 h8-128-taj-bmc INFO pmon#bmcctld.syslog[36]: NOTICE: CHASSIS_MODULE change: key=SWITCH-HOST admin_status=up
2026 Apr 13 19:52:01.790813 h8-128-taj-bmc INFO pmon#supervisord 2026-04-13 19:52:01,790 INFO success: bmcctld entered RUNNING state, process has stayed up for > than 10 seconds (startsecs)

Working case Logs:
2026 Apr 13 19:56:09.397595 sonic INFO pmon#bmcctld.syslog[29]: Entering event subscription loop
2026 Apr 13 19:56:09.398274 sonic NOTICE pmon#bmcctld: CHASSIS_MODULE change: key=SWITCH-HOST admin_status=up
2026 Apr 13 19:56:09.398983 sonic INFO pmon#bmcctld.syslog[29]: NOTICE: CHASSIS_MODULE change: key=SWITCH-HOST admin_status=up
2026 Apr 13 19:56:09.401134 sonic NOTICE pmon#bmcctld: STARTUP: Switch-Host already ONLINE (daemon restart); skipping boot sequence
2026 Apr 13 19:56:09.401429 sonic NOTICE pmon#bmcctld: CHASSIS_MODULE admin_up: key=SWITCH-HOST NOOP (Switch-Host already ONLINE)
2026 Apr 13 19:56:16.679964 sonic INFO pmon#supervisord 2026-04-13 19:56:16,679 INFO success: bmcctld entered RUNNING state, process has stayed up for > than 10 seconds (startsecs)

yxieca (approved):
### AI-assisted review — Approve

Solid fix. Seeding the CONFIG admin_status dedup before starting the event thread cleanly closes the replay-driven `ACTION_POWER_ON` race, and the ordering (`init_host_state` → `seed` → event thread → gate) is correct — the seed completes before `thread.start()`, so the seeded dedup state is guaranteed visible to the event thread. The description states root cause / mechanism / new behavior well, and the 4 new tests cover the meaningful invariants. Approving; a few non-blocking nits below.

**Non-blocking nits**

1. **`get_switch_host_admin_status` docstring vs. its air-cooled caller.** The accessor's docstring (`policy_reader`, ~L568-573) says *"Callers pass ADMIN_UP for air-cooled systems ... and ADMIN_DOWN for liquid-cooled,"* but the only air-cooled caller (startup gate, ~L1350) calls it with **no default** (so it falls back to `ADMIN_DOWN`) and instead relies on `initialize_chassis_module()` in `__init__` having already seeded CONFIG_DB with `admin_status=up`. Runtime behavior is correct (a fresh air-cooled box gets seeded `up` and powers on), but the docstring and the call disagree on paper, and the startup power-on decision is implicitly coupled to the earlier seed. Suggest either calling `get_switch_host_admin_status(default=ADMIN_UP)` at the air-cooled branch (self-consistent, and robust even if the seed path were ever skipped/reordered), or updating the docstring to describe the actual "seed-in-`__init__`, read-with-ADMIN_DOWN-default-here" contract. Right now the diff-in-isolation reads like "air-cooled with no CONFIG entry stays powered off," which sent me tracing `__init__` to confirm it doesn't — worth removing that trap for the next reader.

2. **Orphaned sentence in the `run()` docstring.** The rewritten docstring ends a paragraph with a dangling fragment: `"...regardless of the startup path.` / `startup power-on or the action loop."` — the last line looks like leftover edit debris; please drop it.

3. **Thread-safety comment on `_last_chassis_module_admin_status`.** It's a plain dict written by the main thread in `seed_chassis_module_admin_status()` and read/written by the event thread in `_handle_chassis_module()`. That's safe *only* because the seed strictly precedes `event_thread.start()` (after which the event thread owns it exclusively). A one-line comment to that effect would prevent a future refactor that moves the seed after `thread.start()` from silently introducing a data race.

judyjoseph (approved):
LGTM thanks for taking care of the case when admin was already up - we need not unnecessarily send POWER_ON.

PR #853, opened 2026-07-06
Add timestamp to the leak sensor data in DB, Fix to not escalate a minor leak if platform doesn't support it

### LINKED ISSUE #28218: Bug: BMC leak escalation logic is not aligned with the HLD
### Is it platform specific

generic

### Importance or Severity

Critical

### Description of the bug

According to the [HLD](https://github.com/sonic-net/SONiC/blob/master/doc/bmc/sonicBMC/pmon-bmc-design.md) for `get_leak_max_minor_duration_sec()`

_This API could return back 0 if a platform don't support this concept of minor severity leak gets critical over time_

While the current implementation is propagating the leak from MINOR to CRITICAL in 0 seconds:
https://github.com/sonic-net/sonic-platform-daemons/blob/c3b1d94d9cc91bd14da2be52ccb8906c81d4594a/sonic-thermalctld/scripts/thermalctld#L681-L686

Logs:
```
ERR pmon#thermalctld: Liquid cooling leakage sensor TestSensor reported leaking
ERR pmon#thermalctld: Leak on sensor TestSensor escalated from MINOR to CRITICAL after 0s
```

### Steps to Reproduce

Simulate a minor leak with `get_leak_max_minor_duration_sec()` = 0.

The leak severity escalation should not happen when the `get_leak_max_minor_duration_sec()` value is 0

### Relevant log output

```shell

```

```shell

```

### Attach files (if any)

_No response_

judyjoseph:
#### Description

Couple of enhancements/fixes in this PR 

1. Add timestamp to the leak sensor data in DB
2. Fix to not escalate a minor leak if platform doesn't support it
    Fixes https://github.com/sonic-net/sonic-buildimage/issues/28218

3. Also update the leak_severity in sensor table to in DB correctly as 'CRITICAL' 'MINOR' and 'None' strings

#### Motivation and Context
If the platform doesn't support escalation of a minor leak --> critical Skip it by checking the return value of platform API is 0
Add timestamp to sensor leak data entry in sensor table -- This will help triage when the leak started.

Unit tests added which mock the leak conditions and test

### DIFF
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -742,7 +742,7 @@ class LiquidCoolingUpdater(threading.Thread, logger.Logger):
                 ('leak_sensor_status', 'Good' if sensor_is_ok else 'Fault'),
                 ('type', str(sensor_type)),
                 ('location', str(sensor_location)),
-                ('leak_severity', str(sensor_leak_severity)),
+                ('leak_severity', sensor_leak_severity.value if sensor_leak_severity is not None else 'None'),
             )
             if self.last_sensor_fvs.get(sensor_name) != fvs_tuple:
                 fvs_with_ts = list(fvs_tuple) + [
--- a/sonic-thermalctld/scripts/thermalctld
+++ b/sonic-thermalctld/scripts/thermalctld
@@ -679,7 +679,9 @@ class LiquidCoolingUpdater(threading.Thread, logger.Logger):
                         leak_start_time = self.leaking_sensors[sensor_name]
                         profile = sensor.get_leak_profile()
                         max_minor_secs = profile.get_leak_max_minor_duration_sec() if profile else inf
-                        if max_minor_secs < inf:
+                        # A value of 0 (or negative) means the platform does not support
+                        # time-based escalation from MINOR to CRITICAL; skip escalation.
+                        if 0 < max_minor_secs < inf:
                             max_minor = leak_start_time + timedelta(seconds=max_minor_secs)
                             if max_minor <= now:
                                 sensor_leak_severity = LeakSeverity.CRITICAL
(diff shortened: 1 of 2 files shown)


Yakiv-Huryk:
LGTM.
@judyjoseph could you please also consider addressing the following in this PR?
https://github.com/sonic-net/sonic-mgmt/pull/25769#pullrequestreview-4632239583

judyjoseph:
> LGTM. @judyjoseph could you please also consider addressing the following in this PR? [sonic-net/sonic-mgmt#25769 (review)](https://github.com/sonic-net/sonic-mgmt/pull/25769#pullrequestreview-4632239583)

sure let me check @Yakiv-Huryk

Yakiv-Huryk (approved):

yxieca (approved):
**AI-assisted review — Approve**

Reviewed the thermalctld leak-sensor changes:

- **Escalation fix** (`if 0 < max_minor_secs < inf`): a `get_leak_max_minor_duration_sec()` of 0/negative now correctly means "platform does not support time-based MINOR→CRITICAL escalation" instead of escalating immediately. Fixes sonic-buildimage#28218.
- **`leak_severity` DB value** now written as `sensor_leak_severity.value` (`'MINOR'`/`'CRITICAL'`) with a `'None'` fallback, instead of `str(enum)` which produced `'LeakSeverity.MINOR'`. Applied consistently at all severity-write sites.
- **`timestamp`** added to the sensor-table write but deliberately excluded from the `fvs_tuple` diff-cache, so the advancing clock doesn't defeat change-detection and rewrite the row every poll.

Test coverage is solid — the no-escalation-when-duration-zero case, timestamp-on-transitions-only, and timestamp-not-in-diff-cache behaviors are all explicitly verified.

Looks good.

AI-assisted review, reviewed and approved by Ying Xie (@yxieca).

PR #855, opened 2026-07-09
Populating chassis_serial_number to state_db DEVICE_METADATA Table

gauravnagesh-msft:
#### Description

Populating `chassis_serial_number` into STATE_DB/DEVICE_METADATA/localhost/chassis_serial_number

#### Motivation and Context

The `DEVICE_METADATA` table is currently populated by snmp startup script. However, if we decide to shutdown snmp container service and rely on telemetry only, there is no way to easily fetch serial number from database. So, adding into Platform daemon in the same function where we save it in `CHASSIS_INFO` table.

Verified changes and checked Redis to verify the data population on one of the platforms.


judyjoseph (approved):
LGTM.. @gauravnagesh-msft can you add some unit test for his case in this PR itself? A sonic-mgmt test would help as this is applicable to both chassis and non-chassis (1RU) skus

vvolam (approved):

PR #856, opened 2026-07-16
Updated STATE_DB tables for c-cmis pm

nkanchi-nexthop:
#### Description

- Adds a new `TRANSCEIVER_PM_COUNTERS` STATE_DB table and routes raw FEC counter fields (page 34h counters and page 3Ah counters) from `get_transceiver_pm()` into it, while calculated ratios and metrics stay in `TRANSCEIVER_PM`.
- Adds the table constant, per-asic Table handle, and accessor to `xcvr_table_helper.py`.
- Splits the PM dict by a `PM_COUNTER_FIELDS` set in `post_port_pm_info_to_db` (`dom_mgr.py`), updates the caller to pass both table handles, and adds cleanup of the new table on logical port removal.

#### Motivation and Context

`TRANSCEIVER_PM` was only accumulating derived analog PM metrics (BER ratios, optical power/OSNR/etc.) from page 34h counters. Adding raw counters into their own table gives consumers a distinction between processed diagnostics and raw counters. Also, the new table gives a location for the new host counters from [C-CMIS 1.4](https://www.oiforum.com/wp-content/uploads/OIF-C-CMIS-01.4.pdf) page 3Ah.

- Ran the xcvrd unit tests (`tests/test_xcvrd.py`), focusing on `test_post_port_pm_info_to_db` (asserts non-counter fields land in `TRANSCEIVER_PM` and counter fields land in `TRANSCEIVER_PM_COUNTERS`).
- Verified on a DUT with changes from following PRS: [C-CMIS PM API/Memory Maps](https://github.com/sonic-net/sonic-platform-common/pull/719), [C-CMIS PM STATE_DB](https://github.com/sonic-net/sonic-platform-daemons/pull/856),  [C-CMIS PM CLI](https://github.com/sonic-net/sonic-utilities/pull/4699).

### DIFF
--- a/sonic-xcvrd/xcvrd/dom/dom_mgr.py
+++ b/sonic-xcvrd/xcvrd/dom/dom_mgr.py
@@ -36,6 +36,23 @@ SYSLOG_IDENTIFIER_DOMINFOUPDATETASK = "DomInfoUpdateTask"
 PORT_UPDATE_EVENT_SELECT_TIMEOUT_MSECS = 1000
 PORT_UPDATE_EVENT_SELECT_TIMEOUT_FAST_MSECS = 10
 
+# Raw FEC counter fields from get_transceiver_pm() that are published to the
+# TRANSCEIVER_PM_COUNTERS table instead of TRANSCEIVER_PM. This is a positive list:
+# any field not named here (calculated ratios, optical metrics) stays in TRANSCEIVER_PM.
+PM_COUNTER_FIELDS = frozenset({
+    # media (page 34h) RX FEC counters
+    'rx_bits_pm', 'rx_bits_subint_pm', 'rx_corr_bits_pm',
+    'rx_min_corr_bits_subint_pm', 'rx_max_corr_bits_subint_pm',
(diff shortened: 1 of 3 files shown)


sonic-xcvrd/xcvrd/dom/dom_mgr.py
```diff
@@ -36,6 +36,23 @@
 PORT_UPDATE_EVENT_SELECT_TIMEOUT_MSECS = 1000
 PORT_UPDATE_EVENT_SELECT_TIMEOUT_FAST_MSECS = 10
 
+# Raw FEC counter fields from get_transceiver_pm() that are published to the
+# TRANSCEIVER_PM_COUNTERS table instead of TRANSCEIVER_PM. This is a positive list:
+# any field not named here (calculated ratios, optical metrics) stays in TRANSCEIVER_PM.
+PM_COUNTER_FIELDS = frozenset({
```
prgeor, line 42: @nkanchi-nexthop there are streaming telemetry applications already using these counters from `TRANSCEIVER_PM` so this change will break those. To maintain backward compatibility, I would suggest to not remove these fields from old PM table

PR #861, opened 2026-07-28
[xcvrd] Verify that the module activated the requested CMIS application

tahmed-dev:
### Why I did it

`is_cmis_application_update_required()` compares the desired application against `get_application()`, which reads the **staged** control set. The staged value reflects what xcvrd requested, not what the module is actually running.

A module can accept a staged application, return success from `ApplyDataPathInit`, report `ConfigSuccess` on every lane and activate its datapath, while leaving the **Active Control Set** unchanged. xcvrd then sees a staged application equal to the desired one plus a healthy datapath, logs `no CMIS application update required...READY`, and marks the port READY.

The module keeps running its previous application on the media side, so the link never comes up. Nothing in the logs or in `TRANSCEIVER_STATUS` indicates why, because every value xcvrd inspects looks correct. The only sign is that `active_apsel_hostlane*` in `TRANSCEIVER_INFO` differs from the application xcvrd asked for.

This was hit on a 100G port where xcvrd correctly selected the `CAUI-4 C2M` / `100G PSM4` application, but the module stayed on a `400GBASE-DR4` application. The optical power was healthy in both directions while the media side ran the wrong modulation, so the port reported READY and never linked.

### How I did it

Added `is_active_apsel_matching_desired()`, which compares the desired application against `get_active_apsel_hostlane()` (the Active Control Set) for every host lane in the mask.

It is used in two places:

- `is_cmis_application_update_required()`: when the staged application matches and the datapath looks healthy, the active application is checked before the update is skipped. This covers xcvrd restarts, where the CMIS state is preserved.
- The `CMIS_STATE_DP_ACTIVATE` to `CMIS_STATE_READY` transition: the active application is verified before the port is declared READY.

On mismatch the datapath is reinitialized through the existing retry path. Once `CMIS_MAX_RETRIES` is exhausted the port transitions to `CMIS_STATE_FAILED`, which surfaces the problem instead of leaving the port silently on the wrong application.

The check is skipped when the active application code is not readable, for example the `'N/A'` reported for flat memory modules, so modules that do not expose it keep the current behaviour. On a compliant module the staged and active codes match after activation, so this is a no-op.

### How to verify it

Unit tests:

```
cd sonic-xcvrd && python3 -m pytest tests/test_xcvrd.py -k "is_active_apsel_matching_desired or is_cmis_application_update_required"
```

Added `test_CmisManagerTask_is_active_apsel_matching_desired`, covering a matching active application, a full mismatch, a partial per-lane mismatch, lanes outside the host lane mask, an unreadable `'N/A'` active code, and an empty result. Added `test_CmisManagerTask_is_cmis_application_update_required_active_apsel_mismatch`, which reproduces the reported case: staged application matches, `DataPathActivated` plus `ConfigSuccess` on all lanes, active application different, update still required.

On hardware, compare `active_apsel_hostlane*` in `TRANSCEIVER_INFO` against the application xcvrd selected. A module that fails to activate it now retries and ends in `CMIS_STATE_FAILED` with an explicit log, instead of reporting READY.

### Description for the changelog

Detect CMIS modules that fail to activate the requested application instead of reporting the port as READY.

### DIFF
--- a/sonic-xcvrd/tests/test_xcvrd.py
+++ b/sonic-xcvrd/tests/test_xcvrd.py
@@ -3800,6 +3800,57 @@ class TestXcvrdScript(object):
 
         assert task.is_cmis_application_update_required(mock_xcvr_api, app_new, host_lanes_mask) == expected
 
+    @pytest.mark.parametrize("appl, host_lanes_mask, active_apsel, expected", [
+        # The module activated the desired application
+        (1, 0x0F, {'ActiveAppSelLane1': 1, 'ActiveAppSelLane2': 1,
+                   'ActiveAppSelLane3': 1, 'ActiveAppSelLane4': 1}, True),
+        # The module is running a different application than the desired one
+        (12, 0x0F, {'ActiveAppSelLane1': 4, 'ActiveAppSelLane2': 4,
+                    'ActiveAppSelLane3': 4, 'ActiveAppSelLane4': 4}, False),
+        # Only a subset of the host lanes failed to activate the desired application
+        (12, 0x0F, {'ActiveAppSelLane1': 12, 'ActiveAppSelLane2': 12,
+                    'ActiveAppSelLane3': 12, 'ActiveAppSelLane4': 4}, False),
+        # Host lanes outside of the mask are ignored
+        (12, 0x03, {'ActiveAppSelLane1': 12, 'ActiveAppSelLane2': 12,
+                    'ActiveAppSelLane3': 4, 'ActiveAppSelLane4': 4}, True),
+        # The active application code is not readable
+        (12, 0x0F, {'ActiveAppSelLane1': 'N/A', 'ActiveAppSelLane2': 'N/A',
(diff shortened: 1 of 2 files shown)


tahmed-dev:
Closing this — the premise doesn't hold.

I opened it after finding a module reporting active application 4 while xcvrd had staged application 12, and concluded the port could never come up. That was wrong. The link in question is now up and passing traffic at 100G with exactly that staged/active mismatch still present. The actual fault was unrelated to CMIS: the switch port was cabled to a different host interface that was administratively down.

So a staged/active application mismatch is not on its own evidence of a broken datapath. Worse, this change would be actively harmful in that situation: it would force repeated datapath reinitialisation on a working port and drive it to `CMIS_STATE_FAILED` once `CMIS_MAX_RETRIES` is exhausted.

If there is still an appetite for surfacing a staged/active divergence, it should be a log or a state field rather than something that triggers reconfiguration — but I don't have a case that justifies it today.

Apologies for the noise. #862 is unaffected and stands on its own reproduction.

PR #862, opened 2026-07-28
[xcvrd] Bootstrap the CMIS state machine for ports that miss their PORT_SET

tahmed-dev:
### Why I did it

A port only enters the CMIS state machine once a `PORT_SET` event assigns it an initial state through `force_cmis_reinit()`. `get_cmis_state_from_state_db()` reports an absent `cmis_state` as `UNKNOWN`, and `process_single_lport()` returns immediately on `UNKNOWN`:

```python
state = common.get_cmis_state_from_state_db(...)
if state in CMIS_TERMINAL_STATES or state == CMIS_STATE_UNKNOWN:
    ...
    return
```

So `UNKNOWN` is treated as "already finished" rather than "not yet started". A port whose `PORT_SET` was never delivered is skipped on every iteration and left completely unmanaged: no CMIS state is ever written, the transceiver is never queried, and nothing is logged. There is no recovery path short of restarting xcvrd.

This was observed on a port created by a dynamic port breakout. It had no `cmis_state` field at all, while every other port on the switch reported `READY`. The transceiver stayed on the application it powered up with, and the link never came up. Restarting xcvrd fixed it, since the port was then present when the port map was built.

Reproduced against the current code with a fully configured port that never received a `PORT_SET`:

| Scenario | `cmis_state` | Transceiver queried |
|---|---|---|
| `PORT_SET` missed | never written | never |
| `PORT_SET` delivered (control) | `INSERTED` | yes |
| `PORT_SET` without `index` | `INSERTED`, stuck | no, and `KeyError('index')` |

### How I did it

Bootstrap the state machine in `process_single_lport()` when a port has no CMIS state but a complete configuration, instead of skipping it. The port then follows the normal `INSERTED` path. This closes the window regardless of why the event was lost, and is a no-op for ports that already have a state.

The completeness check is factored into `is_port_config_complete()`, which also replaces the existing inline validity check so both use the same rule. Ports without a usable configuration, for example a port whose physical index is not known yet, are still skipped exactly as before.

Also always record a physical index in `port_dict`. `on_port_update_event()` only stored it when `pport >= 0`, while `port_event_helper` defaults a missing `index` field to `-1`. A port could therefore be admitted to `port_dict` without an `index` key, and `is_decomm_pending()` and `is_decomm_lead_lport()` index into it directly, outside the `try` block in `process_cmis_state_machine()`. This part is defensive, as I did not establish a reachable path to that `KeyError` in the current flow.

### How to verify it

```
cd sonic-xcvrd && python3 -m pytest tests/test_xcvrd.py -k "is_port_config_complete or always_records_index or bootstraps_missing_cmis_state or skips_incomplete_config"
```

Added:

- `test_CmisManagerTask_is_port_config_complete`, a matrix over missing, invalid and non numeric `index`, `speed`, `lanes` and `subport`.
- `test_CmisManagerTask_on_port_update_event_always_records_index`, covering a `PORT_SET` with no index followed by one with a valid index.
- `test_CmisManagerTask_process_single_lport_bootstraps_missing_cmis_state`, which reproduces the reported case: a fully configured port with no CMIS state is now bootstrapped and its transceiver queried.
- `test_CmisManagerTask_process_single_lport_skips_incomplete_config`, which confirms a port without a usable configuration is still skipped and its transceiver never touched.

On hardware, create a port through a dynamic port breakout and confirm `cmis_state` appears in `TRANSCEIVER_STATUS` without restarting xcvrd.

### Description for the changelog

Bootstrap the CMIS state machine for ports that never received a PORT_SET event, so they are not left unmanaged until xcvrd restarts.

### DIFF
--- a/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
+++ b/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
@@ -122,6 +122,9 @@ class CmisManagerTask(threading.Thread):
             if lport not in self.port_dict:
                 self.port_dict[lport] = {"asic_id": port_change_event.asic_id,
                                          "forced_tx_disabled": False}
+            # Always record a physical index, even when it is still unknown, so that
+            # the consumers indexing into port_dict do not have to handle its absence.
+            self.port_dict[lport].setdefault('index', -1)
             if pport >= 0:
                 self.port_dict[lport]['index'] = pport
             if 'speed' in port_change_event.port_dict and port_change_event.port_dict['speed'] != 'N/A':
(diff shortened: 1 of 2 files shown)


sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -1249,8 +1252,41 @@ def process_cmis_state_machine(self, lport):
...
         state = common.get_cmis_state_from_state_db(lport, self.xcvr_table_helper.get_status_sw_tbl(self.get_asic_id(lport)))
+        if state == CMIS_STATE_UNKNOWN and self.is_port_config_complete(info):
+            # A port only enters the state machine once a PORT_SET event assigns it an
+            # initial CMIS state. When that event is missed, for instance because the
+            # port was created by a dynamic port breakout after the port map was built,
+            # the port keeps an empty CMIS state. Since an empty state reads back as
+            # UNKNOWN, and UNKNOWN is skipped below, the port would stay unmanaged until
+            # xcvrd is restarted. Bootstrap the state machine instead.
+            self.log_notice("{}: no CMIS state found, starting the CMIS state machine".format(lport))
+            self.update_port_transceiver_status_table_sw_cmis_state(lport, CMIS_STATE_INSERTED)
```
prgeor, line 1291: @tahmed-dev are you saying the port update event was missed by Xcvrd during port breakout? There are important port attributes like lanes, speed, subport values that are obtained when new port is created(that how self.port_dict[]) without that even if CMIS_STATE_INSERTED is forced, the port cannot be reliably initialized(its working in one case doesn't  invalidates the requirement). If Xcvrd is missing the event, how do we know that `process_single_lport()` is now operating on old port info versus newly created breakout port information?

PR #867, opened 2026-08-05
[chassisd]: Replay saved module admin down state on boot in chassisd

gupurush:
Issue: https://github.com/sonic-net/sonic-platform-daemons/issues/868

## Summary
- Add `set_initial_module_admin_state()` to replay persisted `admin_status: down` from CONFIG_DB at chassisd startup on modular chassis supervisors (8808 RP).
- For each present fabric/line module with saved admin down, call `set_admin_state(MODULE_ADMIN_DOWN)` so hardware shutdown is applied after reboot without a new config event.

#### Motivation and Context
After reboot, a module can remain `admin_status: down` in CONFIG_DB while oper stays Online because chassisd never replays the hardware shutdown. Z rails stay powered until a new shutdown config event is applied. This change replays the saved admin-down state once at supervisor boot, after config_manager starts and before the main loop.

- DUT 8808 RP: config chassis modules shutdown FABRIC-CARD1 → sudo config save -y → reboot
- After pmon/chassisd up: FC1 admin=down, oper=Offline, rails off, no swss@2/3

### DIFF
--- a/sonic-chassisd/scripts/chassisd
+++ b/sonic-chassisd/scripts/chassisd
@@ -410,6 +410,13 @@ class ModuleUpdater(logger.Logger):
             return fvs.get(CHASSIS_MODULE_INFO_OPERSTATUS_FIELD, ModuleBase.MODULE_STATUS_EMPTY)
         return ModuleBase.MODULE_STATUS_EMPTY
 
+    def get_module_current_presence(self, key):
+        fvs = self.module_table.get(key)
+        if isinstance(fvs, list) and fvs[0] is True:
+            fvs = dict(fvs[-1])
+            return fvs.get(CHASSIS_MODULE_INFO_PRESENCE_FIELD, 'False')
+        return None
+
     def get_module_admin_status(self, chassis_module_name):
         config_db = daemon_base.db_connect("CONFIG_DB")
         vtable = swsscommon.Table(config_db, CHASSIS_CFG_TABLE)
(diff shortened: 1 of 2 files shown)


rlhui:
@ymd-arista please help review

mlok-nokia:
Nokia chassis don't support shutdown a module (LC and Fabric card) or keep a module in down state.  How this works on Nokia platform?

sonic-chassisd/scripts/chassisd
```diff
@@ -410,6 +410,13 @@ class ModuleUpdater(logger.Logger):
             return fvs.get(CHASSIS_MODULE_INFO_OPERSTATUS_FIELD, ModuleBase.MODULE_STATUS_EMPTY)
         return ModuleBase.MODULE_STATUS_EMPTY
 
+    def get_module_current_presence(self, key):
+        fvs = self.module_table.get(key)
+        if isinstance(fvs, list) and fvs[0] is True:
+            fvs = dict(fvs[-1])
+            return fvs.get(CHASSIS_MODULE_INFO_PRESENCE_FIELD, 'False')
+        return None
```
ymd-arista, line 418: Is there a reason why we can't return 'False' here? If we look at other `get_*` functions, we return something rather than None.

sonic-chassisd/scripts/chassisd
```diff
@@ -420,6 +427,23 @@ class ModuleUpdater(logger.Logger):
...
+
+        if self.get_module_admin_status(module_name) != 'down':
+            return False
+
+        module = self.chassis.get_module(module_index)
+        if not try_get(module.get_presence, default=False):
+            return False
+
+        self.log_info("Replaying admin down state for module {}".format(module_name))
+        try_get(module.set_admin_state, MODULE_ADMIN_DOWN, default=False)
```
ymd-arista, line 444: We aren't using or checking the return value from `set_admin_state` anywhere right?

If setting admin state failed for example, do we still want to return `True`?

judyjoseph:
Points to consider in scenarios, add test results
1. LC reseat what is behavior 
2. The Supervisor config db parsing could be late .. resulting in LC to come up first and shutdown later.

Or a separate PR, share details.

PR #871, opened 2026-08-10
Fix build test failures

bgallagher-nexthop:
#### Description
Build failures have been observed in https://github.com/sonic-net/sonic-buildimage/pull/28794 as a result of a new import being added in https://github.com/sonic-net/sonic-platform-common/pull/700 that causes some unit-tests to fail.

#### Motivation and Context
This is required to fix test failures during the build and allow the submodule advancement PR to land.

```
  make SONIC_BUILD_JOBS=8 target/python-wheels/trixie/sonic_{bmcctld,sensormond,thermalctld,psud,\
  syseepromd,pcied,stormond,xcvrd,ycabled,chassisd}-1.0-py3-none-any.whl \
  target/python-wheels/trixie/sonic_ledd-1.1-py3-none-any.whl
```
All builds passed.

### DIFF
--- a/sonic-bmcctld/tests/mocked_libs/sonic_py_common/device_info.py
+++ b/sonic-bmcctld/tests/mocked_libs/sonic_py_common/device_info.py
@@ -1,8 +1,0 @@
-"""Test stub matching sonic_py_common.device_info."""
-
-
-def get_cpo_data():
-    # ChassisBase.__init__ calls this to decide whether to build CPO devices.
-    # The real implementation returns None when the platform has no cpo.json,
-    # which is the case on a test host.
-    return None
--- a/sonic-bmcctld/tests/mocked_libs/sonic_py_common/device_info.py
+++ b/sonic-bmcctld/tests/mocked_libs/sonic_py_common/device_info.py
@@ -1,0 +1,8 @@
+"""Test stub matching sonic_py_common.device_info."""
+
+
+def get_cpo_data():
+    # ChassisBase.__init__ calls this to decide whether to build CPO devices.
+    # The real implementation returns None when the platform has no cpo.json,
+    # which is the case on a test host.
+    return None
--- a/sonic-bmcctld/tests/mocked_libs/sonic_py_common/__init__.py
+++ b/sonic-bmcctld/tests/mocked_libs/sonic_py_common/__init__.py
@@ -1,2 +1,3 @@
 from . import daemon_base
+from . import device_info
 from . import logger
--- a/sonic-bmcctld/tests/mocked_libs/sonic_py_common/__init__.py
+++ b/sonic-bmcctld/tests/mocked_libs/sonic_py_common/__init__.py
@@ -1,3 +1,2 @@
 from . import daemon_base
-from . import device_info
 from . import logger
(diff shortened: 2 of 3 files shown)


bgallagher-nexthop:
cc @yijingyan2

prgeor (dismissed):

yijingyan2:
> cc @yijingyan2

Thanks for the quick fix!

sonic-bmcctld/tests/mocked_libs/sonic_py_common/device_info.py
```diff
@@ -0,0 +1,8 @@
+"""Test stub matching sonic_py_common.device_info."""
```
prgeor: @bgallagher-nexthop instead of repeating this fix for all future new daemons, can we instead fix the [PR](https://github.com/sonic-net/sonic-platform-common/pull/700) as follows:?
```
ChassisBase.__init__():
    ...
          try:
            cpo_data = device_info.get_cpo_data()
        except AttributeError:
            cpo_data = None
        if cpo_data:
            self.construct_cpo_devices(cpo_data)
```

bgallagher-nexthop: Can do. Note that the exception the sonic-bmcctld tests were experiencing wasn't `AttributeError` (`get_cpo_data` not implemented due to an older package version) but it was that they were getting `ImportError` because the mocked module contained no `device_info.py` file:
```
____________________ ERROR collecting tests/test_bmcctld.py ____________________
ImportError while importing test module '/sonic/src/sonic-platform-daemons/sonic-bmcctld/tests/test_bmcctld.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
scripts/bmcctld:53: in <module>
    from sonic_platform_base.chassis_base import ChassisBase
/usr/local/lib/python3.13/dist-packages/sonic_platform_base/__init__.py:1: in <module>
    from . import chassis_base
/usr/local/lib/python3.13/dist-packages/sonic_platform_base/chassis_base.py:9: in <module>
    from sonic_py_common import device_info
E   ImportError: cannot import name 'device_info' from 'sonic_py_common' (/sonic/src/sonic-platform-daemons/sonic-bmcctld/tests/mocked_libs/sonic_py_common/__init__.py)

During handling of the above exception, another exception occurred:
/usr/lib/python3.13/importlib/__init__.py:88: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_bmcctld.py:55: in <module>
    load_source('bmcctld', os.path.join(scripts_path, 'bmcctld'))
tests/test_bmcctld.py:22: in load_source
    spec.loader.exec_module(module)
scripts/bmcctld:60: in <module>
    raise ImportError(str(e) + " - required module not found")
E   ImportError: cannot import name 'device_info' from 'sonic_py_common' (/sonic/src/sonic-platform-daemons/sonic-bmcctld/tests/mocked_libs/sonic_py_common/__init__.py) - required module not found
- generated xml file: /sonic/src/sonic-platform-daemons/sonic-bmcctld/test-results.xml -
```

https://github.com/sonic-net/sonic-platform-common/pull/734 contains a fix to the library itself that catches ImportError so any tests missing the device_info file in their mock package do not fail, if you'd like to proceed with that fix. The downsides are:
- a real import failure will cause CPO device creation to be silently skipped
- we are changing production code to compensate for unit-tests having a mocking strategy that is brittle to dependency changes

prgeor (approved):

PR #876, opened 2026-08-14
[CPO] Add CpoManagerTask

bgallagher-nexthop:
#### Description
As outlined in the [xcvrd CPO HLD](https://github.com/sonic-net/SONiC/blob/c334d5cbb1cd7c76500bcdcdadbdab2e627869a8/doc/pmon/xcvrd_cpo.md), a CPO-specific subclass of `CmisManagerTask` is required to handle the CMIS state machine for CPO hardware.

#### Motivation and Context
Customization of the CMIS state machine is necessary to correctly handle CPO hardware.

Unit-tests have been added.

### DIFF
--- a/sonic-xcvrd/tests/test_cpo.py
+++ b/sonic-xcvrd/tests/test_cpo.py
@@ -1,9 +1,17 @@
 import contextlib
+import threading
 
 from unittest.mock import MagicMock, patch
 
 from sonic_py_common import device_info
+from xcvrd import xcvrd  # noqa: F401
+from xcvrd.cmis import cmis_manager_task
+from xcvrd.cmis.cmis_manager_task import CmisManagerTask
+from xcvrd.cpo.cpo_manager_task import CpoManagerTask
 from xcvrd.xcvrd_utilities import common
+from xcvrd.xcvrd_utilities.port_event_helper import PortChangeEvent, PortMapping
+
+DEFAULT_NAMESPACE = ['']
 
 
 class TestPortDeviceResolver:
@@ -130,6 +138,21 @@ PLATFORM_DATA = {
     },
 }
 
+# All three interfaces share the very same optical engine
+SINGLE_OE_CPO_DATA = {
+    'devices': {
+        'OE1': {'device_type': 'optical_engine', 'max_banks': 3},
+    },
+    'interfaces': {
+        'Ethernet0': {'associated_devices': [{'device_id': 'OE1', 'bank': 0}]},
+        'Ethernet8': {'associated_devices': [{'device_id': 'OE1', 'bank': 1}]},
+        'Ethernet16': {'associated_devices': [{'device_id': 'OE1', 'bank': 2}]},
+    },
+}
+
+# Logical port to physical port mapping matching PLATFORM_DATA
+CPO_PORTS = (('Ethernet0', 1), ('Ethernet8', 2), ('Ethernet16', 3))
+
 
 @contextlib.contextmanager
 def patched_topology(cpo_data=CPO_DATA, platform_data=PLATFORM_DATA):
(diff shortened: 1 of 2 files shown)


sonic-xcvrd/tests/test_cpo.py
```diff
@@ -1,9 +1,17 @@
 import contextlib
+import threading
 
 from unittest.mock import MagicMock, patch
 
 from sonic_py_common import device_info
+from xcvrd import xcvrd  # noqa: F401
+from xcvrd.cmis import cmis_manager_task
+from xcvrd.cmis.cmis_manager_task import CmisManagerTask
+from xcvrd.cpo.cpo_manager_task import CpoManagerTask
```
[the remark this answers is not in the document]
bgallagher-nexthop: This was necessary to deal with `cmis_manager_task` not being importable by itself.

`from xcvrd.cmis import cmis_manager_task` throws an exception due to this:
```
  --> START xcvrd.cmis.cmis_manager_task
      --> START xcvrd.xcvrd_utilities.media_settings_parser
        --> START xcvrd.xcvrd            ← re-enters the package from the top, which attempts to import CmisManagerTask
          ...
  FAILED: cannot import name 'CmisManagerTask' from partially initialized module
          'xcvrd.cmis.cmis_manager_task'
```

prgeor (approved):

sonic-xcvrd/xcvrd/cpo/cpo_manager_task.py
```diff
@@ -11,3 +12,91 @@ def __init__(self, namespaces, port_mapping, port_obj_dict, main_thread_stop_eve
...
+                if sibling_api is None:
+                    self.log_error("{}: no xcvr api available for sibling physical port "
+                                   "{}".format(lport, sibling_pport))
+                    done = False
+                    continue
+
+                lanes_mask = self.get_cmis_max_host_lanes_mask(sibling_api)
+                self.log_notice("{}: set datapath deinit and disable Tx output for all lanes "
+                                "of sibling physical port {}".format(lport, sibling_pport))
+                sibling_api.set_datapath_deinit(lanes_mask)
```
micas-net, line 63: can we skip it if this operation has been done before?

sonic-xcvrd/xcvrd/cpo/cpo_manager_task.py
```diff
@@ -11,3 +12,91 @@ def __init__(self, namespaces, port_mapping, port_obj_dict, main_thread_stop_eve
...
+                if not sibling_api.tx_disable_channel(lanes_mask, True):
```
micas-net, line 64: can we skip it if this operation has been done before?

sonic-xcvrd/xcvrd/cpo/cpo_manager_task.py
bgallagher-nexthop: I had intentionally left this to be repeated, even if another interface has performed the action already as it should not cause any problems other than potentially sometimes delaying the module from coming out of low power mode for a short time. To prevent the operation being repeated, we'd have to add some tracking state to either the `CmisManagerTask` or skip the write in the `CmisApi` if the write would be a no-op because the bit is already set to the desired value. The latter is probably the simpler place to do this dedup, but I am not confident it is worth the added complexity.

Let me know if you think it is worth it.

sonic-xcvrd/xcvrd/cpo/cpo_manager_task.py
bgallagher-nexthop: Same as above.

sonic-xcvrd/xcvrd/cpo/cpo_manager_task.py
micas-net: Could you please clarify the motivation behind deinitializing all ports associated with the same OE when only one port is being deinitialized?
In my understanding, we should be able to deinitialize ports individually rather than by OE. Is that correct?

bgallagher-nexthop: This is mirroring existing logic in the `CmisManagerTask` where even if we have multiple breakout interfaces on a single transceiver driving their own DPSM, if the module is in low power all lanes are deinitialized and disabled to prevent stale/default applications taking effect when the module is brought out of low power mode at the end of the function.
```
        # Deinit and disable all lanes if we are in ModuleLowPwr to avoid unintentional
        # initialization of other datapaths during transition to ModuleReady
        if self.check_module_state(api, ['ModuleLowPwr']):
            self.log_notice("{}: ModuleLowPwr detected, set datapath deinit and disable Tx output for all lanes".format(lport))
            deinit_host_lanes_mask = self.port_dict[lport]['max_host_lanes_mask']
            disable_media_lanes_mask = self.port_dict[lport]['max_media_lanes_mask']
```

This is extending the same solution to CPO hardware -- if you bring an optical engine out of low power mode without deinitializing and disabling all lanes on the device, all other interfaces sharing the OE may start initializing the datapath with a stale/default appsel.

micas-net (approved):

aditya-nexthop (approved):

sonic-xcvrd/xcvrd/cpo/cpo_manager_task.py
```diff
@@ -11,3 +12,91 @@ def __init__(self, namespaces, port_mapping, port_obj_dict, main_thread_stop_eve
...
+        Returns:
+            Boolean: True if state machine should continue to next state,
+                     False if processing should stop (return from caller)
+        """
+        api = self.port_dict[lport].get('api')
+        if self.check_module_state(api, ['ModuleLowPwr']):
+            self.log_notice("{}: ModuleLowPwr detected, deinitializing all lanes of the "
+                            "optical engine".format(lport))
+            if not self.deinit_oe_sibling_pports(lport):
+                self.port_dict[lport]['cmis_retries'] = self.port_dict[lport].get('cmis_retries', 0) + 1
```
tshalvi, line 97: LowPwrS is already included in this equation, taken from CMIS 5.3 OIF:`DPDeinitS = ( NOT ModuleReadyT ) OR LowPwrS OR DPDeinitT   (Eq. 6-6)`

Why do we also need to take it into consideration here? According to the CMIS logic, this is not necessary.

Also, why should sibling ports go through deinit as part of another port's state machine run? Once the CMIS state machine handles a sibling port, it will deinit that port's lanes.

aditya-nexthop: hi @tshalvi, does this comment above help answer your question?

https://github.com/sonic-net/sonic-platform-daemons/pull/876/changes#r3823585110

> This is mirroring existing logic in the `CmisManagerTask` where even if we have multiple breakout interfaces on a single transceiver driving their own DPSM, if the module is in low power all lanes are deinitialized and disabled to prevent stale/default applications taking effect when the module is brought out of low power mode at the end of the function.
> ```
>         # Deinit and disable all lanes if we are in ModuleLowPwr to avoid unintentional
>         # initialization of other datapaths during transition to ModuleReady
>         if self.check_module_state(api, ['ModuleLowPwr']):
>             self.log_notice("{}: ModuleLowPwr detected, set datapath deinit and disable Tx output for all lanes".format(lport))
>             deinit_host_lanes_mask = self.port_dict[lport]['max_host_lanes_mask']
>             disable_media_lanes_mask = self.port_dict[lport]['max_media_lanes_mask']
> ```
> 
> This is extending the same solution to CPO hardware -- if you bring an optical engine out of low power mode without deinitializing and disabling all lanes on the device, all other interfaces sharing the OE may start initializing the datapath with a stale/default appsel.

tshalvi: Just to make sure I understood the reason for this change correctly:

Assuming ports A and B belong to the same shared OE, the risk is that while handling A, xcvrd moves the shared OE from ModuleLowPwr to ModuleReady before B has been processed.

If B is not explicitly held in datapath deinit at that point, the module FW may start initializing B using whatever AppSel/configuration is currently programmed for B, before xcvrd gets a chance to process B and program the desired configuration.

So the purpose of holding the sibling port B in deinit is to prevent the module FW from acting on B until B's own CMIS state-machine iteration configures it correctly, and only then allow B's datapath initialization to proceed.

Is that the intended reason for this change?

bgallagher-nexthop: Yup, that is correct 👍

abelamit (approved):

prgeor (approved):

Issue #568, opened 2024-12-04
CMIS optical xcvrs (> 5.0) Breakout Issues


bobbymcgonigle:
# Problem
On optical transceivers across different vendors on multiple devices, some interfaces do not linkup in breakout mode. 

# Symptoms 
The symptoms below use an 800GBASE-FR4 in 8x100G mode.

1. Interfaces stay linkdown

```
 Ethernet72                               73     100G   9100    N/A  Ethernet10/1  routed      up       up       N/A  QSFP-DD      QDD-800G-2FR4         off        5
 Ethernet73                               74     100G   9100    N/A  Ethernet10/1  routed      up       up       N/A  QSFP-DD      QDD-800G-2FR4         off        5
 Ethernet74                               75     100G   9100    N/A  Ethernet10/1  routed      up       up       N/A  QSFP-DD      QDD-800G-2FR4         off        5
 Ethernet75                               76     100G   9100    N/A  Ethernet10/1  routed    down       up       N/A  QSFP-DD      QDD-800G-2FR4         off        4
 Ethernet76                               77     100G   9100    N/A  Ethernet10/1  routed      up       up       N/A  QSFP-DD      QDD-800G-2FR4         off        5
 Ethernet77                               78     100G   9100    N/A  Ethernet10/1  routed      up       up       N/A  QSFP-DD      QDD-800G-2FR4         off        5
 Ethernet78                               79     100G   9100    N/A  Ethernet10/1  routed    down       up       N/A  QSFP-DD      QDD-800G-2FR4         off        4
 Ethernet79                               80     100G   9100    N/A  Ethernet10/1  routed      up       up       N/A  QSFP-DD      QDD-800G-2FR4         off        5
```
2. Datapath on corresponding lanes are deactivated
3. Xcvr misprogrammed and CMIS state is FAILED
```
Ethernet72:
        CMIS State (SW): FAILED
        Data path state indicator on host lane 1: DataPathActivated
        Data path state indicator on host lane 2: DataPathActivated
        Data path state indicator on host lane 3: DataPathActivated
        Data path state indicator on host lane 4: DataPathDeactivated -> Ethernet75 down
        Data path state indicator on host lane 5: DataPathActivated
        Data path state indicator on host lane 6: DataPathActivated
        Data path state indicator on host lane 7: DataPathDeactivated -> Ethernet78 down
        Data path state indicator on host lane 8: DataPathActivated
        Tx output status on media lane 1: False
[...]

PR #878, opened 2026-08-20
[sonic-xcvrd/cmis]: Proceed with DP init on ConfigSuccess when DPInitPending is not asserted

skv0816-ciena:
#### Description

 Coherent CMIS 5.x 400ZR modules can leave a port stuck in CMIS_STATE_FAILED because xcvrd's DP_INIT handler waited for DPInitPending, which these modules do not re-assert when the application/ApSel is unchanged. The handler then looped via force_cmis_reinit until it exhausted CMIS_MAX_RETRIES and marked the port FAILED.
    
  Fix: In CmisManagerTask, CMIS_STATE_DP_INIT is only reached after the handler's guard has already confirmed ConfigSuccess on all host lanes. The DPInitPending check is therefore made informational only — when a CMIS 5.x module does not assert DPInitPending, xcvrd logs a notice and proceeds with datapath init instead of forcing a reinit. DPInitPending is advisory in CMIS 5.x, and the downstream DP_TXON/DP_ACTIVATE timeouts still retry if init genuinely fails. Modules that do assert DPInitPending are unaffected.
  

#### Motivation and Context
The observable symptom is:

show interfaces transceiver status Ethernet<N> → cmis_state = CMIS_STATE_FAILED
sfputil debug tx-output Ethernet<N> disable|enable → RC != 0
Repeated force_cmis_reinit entries in xcvrd's log, with the port never reaching CMIS_STATE_READY / DataPathActivated
Non-coherent modules (SR/DR/FR/LR QSFP-DD, QSFP+/28, SFP+/28) always assert DPInitPending on application-change or DP deinit-init cycles and therefore hit the existing branch — this change does not affect them.

   root@sonic:~# show interface transceiver status Ethernet1
Ethernet1: 
        CMIS State (SW): FAILED
-->

  Before the fix:

Check	                                                                 Result
show interfaces transceiver status Ethernet47	cmis_state = CMIS_STATE_FAILED
sfputil debug tx-output Ethernet47 disable	         non-zero RC
sfputil debug tx-output Ethernet47 enable	         non-zero RC
xcvrd log	                                                             repeated force_cmis_reinit, port never reaches DataPathActivated

After the fix:

Check	                                                                 Result
show interfaces transceiver status Ethernet47	cmis_state = CMIS_STATE_READY
Live get_xcvr_api() datapath state	                        DataPathActivated on all host lanes
sfputil debug tx-output Ethernet47 disable	        RC 0, Tx1 → false, output power → ~0 dBm
sfputil debug tx-output Ethernet47 enable	         RC 0, Tx1 → true, output power → nominal
xcvrd log	                                                               	no force_cmis_reinit retries; single-pass DP init, informational notice        logged when DPInitPending is not asserted

-->

CMIS 5.x reference: DPInitPending  is defined as "asserted while a data path initialization is in progress." Coherent module vendors interpret "no ApSel/config change" as "no init in progress" and correctly do not assert the bit; the host stack is responsible for detecting steady state via ConfigStatus == ConfigSuccess.
The change is scoped to the branch that already handles CMIS ≥ 5.0 modules (majorRev >= 5); pre-5 CMIS modules are unchanged.
Because CMIS_STATE_DP_INIT already returns unless all host lanes report ConfigSuccess. the DPInitPending result is now purely a diagnostic log before proceeding with datapath init.
Related upstream context: xcvrd has no dedicated coherent code path — the CMIS common flow is expected to handle CMIS 5.x coherent optics. This PR keeps that assumption and only relaxes one over-strict precondition.

### DIFF
--- a/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
+++ b/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
@@ -1206,9 +1206,8 @@ class CmisManagerTask(threading.Thread):
                     # Check datapath init pending on module that supports CMIS 5.x
                     majorRev = int(api.get_cmis_rev().split('.')[0])
                     if majorRev >= 5 and not self.check_datapath_init_pending(api, host_lanes_mask):
-                        self.log_notice("{}: datapath init not pending".format(lport))
-                        self.force_cmis_reinit(lport, retries + 1)
-                        return
+                        self.log_notice("{}: datapath init not pending but ConfigSuccess; "
+                                        "proceeding with datapath init".format(lport))
 
                 # Ensure the Datapath is NOT Activated unless the host Tx siganl is good.
                 # NOTE: Some CMIS compliant modules may have 'auto-squelch' feature where
--- a/sonic-xcvrd/tests/test_xcvrd.py
+++ b/sonic-xcvrd/tests/test_xcvrd.py
@@ -3510,6 +3510,140 @@ class TestXcvrdScript(object):
         mock_xcvr_api.set_datapath_deinit.assert_called_once_with(0xff)
         mock_xcvr_api.tx_disable_channel.assert_called_once_with(0xff, True)
 
+    def _build_dp_init_task(self, mock_get_status_sw_tbl, mock_xcvr_api):
+        """Helper: build a CmisManagerTask with Ethernet0 seeded into CMIS_STATE_DP_INIT."""
+        mock_sfp = MagicMock()
+        mock_sfp.get_presence = MagicMock(return_value=True)
+        mock_sfp.get_xcvr_api = MagicMock(return_value=mock_xcvr_api)
+
+        port_mapping = PortMapping()
+        port_mapping.handle_port_change_event(PortChangeEvent('Ethernet0', 1, 0, PortChangeEvent.PORT_ADD))
+        stop_event = threading.Event()
+        task = CmisManagerTask(DEFAULT_NAMESPACE, port_mapping, {1: mock_sfp}, stop_event)
+        task.xcvr_table_helper = XcvrTableHelper(DEFAULT_NAMESPACE)
+        task.xcvr_table_helper.get_status_sw_tbl.return_value = mock_get_status_sw_tbl
+
+        port_change_event = PortChangeEvent('Ethernet0', 1, 0, PortChangeEvent.PORT_SET,
+                                            {'speed': '400000', 'lanes': '1,2,3,4,5,6,7,8'})
+        task.on_port_update_event(port_change_event)
+
+        task.update_port_transceiver_status_table_sw_cmis_state('Ethernet0', CMIS_STATE_DP_INIT)
+        task.port_dict['Ethernet0']['host_tx_ready'] = 'true'
+        task.port_dict['Ethernet0']['admin_status'] = 'up'
+        task.port_dict['Ethernet0']['appl'] = 1
+        task.port_dict['Ethernet0']['host_lanes_mask'] = 0xff
+        task.port_dict['Ethernet0']['media_lanes_mask'] = 0xff
+        task.port_dict['Ethernet0']['tx_power'] = 0
+        task.port_dict['Ethernet0']['laser_freq'] = 0
+        task._gearbox_lanes_dict = {}
+        return task
+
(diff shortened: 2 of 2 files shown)


skv0816-ciena:
fixes the issue #568

aditya-nexthop:
> #### Description
> #### Motivation and Context
> #### How Has This Been Tested?
> #### Additional Information (Optional)

Please can you fill this out detailing the exact scenario you're trying to fix?

skv0816-ciena:
> > #### Description
> > #### Motivation and Context
> > #### How Has This Been Tested?
> > #### Additional Information (Optional)
> 
> Please can you fill this out detailing the exact scenario you're trying to fix?

Done

aditya-nexthop (approved):
Please fix up the coverage check by adding a unit test, but otherwise looks good.
Thanks for fixing this issue.

skv0816-ciena:
> Please fix up the coverage check by adding a unit test, but otherwise looks good. Thanks for fixing this issue.

done

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -1206,9 +1206,13 @@ def process_cmis_state_machine(self, lport):
...
-                        self.log_notice("{}: datapath init not pending".format(lport))
-                        self.force_cmis_reinit(lport, retries + 1)
-                        return
+                        if self.check_config_error(api, host_lanes_mask, ['ConfigSuccess']):
+                            self.log_notice("{}: datapath init not pending but ConfigSuccess; "
+                                            "proceeding with datapath init".format(lport))
+                        else:
+                            self.log_notice("{}: datapath init not pending".format(lport))
+                            self.force_cmis_reinit(lport, retries + 1)
+                            return
```
bgallagher-nexthop: I think this is now dead code. At the start of the branch up on line 1186, we already determine ConfigSuccess is the current state so I don't think we will ever take this path.

It also calls into question the value of this DPInitPending check in general, since this change will effectively make the DPInitPending check just result in a log message

skv0816-ciena: I see that. I'll clean it up and keep it as the informational log; no action is needed on this check. Downstream DP_TXON/DP_ACTIVATE timeouts still retry if init truly fails.

sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py
```diff
@@ -1206,9 +1206,8 @@ def process_cmis_state_machine(self, lport):
                     # Check datapath init pending on module that supports CMIS 5.x
                     majorRev = int(api.get_cmis_rev().split('.')[0])
                     if majorRev >= 5 and not self.check_datapath_init_pending(api, host_lanes_mask):
-                        self.log_notice("{}: datapath init not pending".format(lport))
-                        self.force_cmis_reinit(lport, retries + 1)
-                        return
+                        self.log_notice("{}: datapath init not pending but ConfigSuccess; "
```
prgeor, line 1209: @skv0816-ciena please ask the vendor to fix it in module firmware else we need to dilute many of similar checks to accommodate different module vendor interpretation or implmentation.

skv0816-ciena: This isn't a vendor firmware bug — per issue #568 , DPInitPending (Page 11h) is a transitional flag that software will miss in many cases (more likely with high xcvr processing and high-port-count hwskus). @bobbymcgonigle demonstrated the missed transient with per-lane logging. So asking vendors to "fix firmware" won't help, and I've reproduced the same behavior across multiple vendors' plugs, including raw Page 11h/byte 235 register captures during a forced ApplyDPInit.

Importantly, this doesn't dilute the surrounding checks. The DP_INIT block already returns/reinits unless all host lanes report ConfigSuccess, which per CMIS means the staged control set was accepted — so the DPInitPending check is redundant for correctness, and the datapath is still fully validated afterward (DP_TXON waits for DataPathInitialized, DP_ACTIVATION waits for DataPathActivated). Unlike ConfigStatus/ModuleState/DPState, which are stable states, DPInitPending is the only transient handshake flag here, so dropping reliance on it is a targeted fix.

Proposal: since we no longer gate on it and only log, remove the log entirely.

PR #884, opened 2026-08-26
[CPO] Add CpoStateUpdateTask

bgallagher-nexthop:
#### Description
As outlined in the [xcvrd CPO HLD](https://github.com/sonic-net/SONiC/blob/c334d5cbb1cd7c76500bcdcdadbdab2e627869a8/doc/pmon/xcvrd_cpo.md), a CPO-specific subclass of `SfpStateUpdateTask` is required to handle ELSFP plug-in / plug-out events on CPO hardware.

#### Motivation and Context
Customization of the `SfpStateUpdateTask` is necessary to correctly handle CPO hardware.

Added unit-testing.

Tested on a non-CPO switch manually, to ensure no regression:
- All links came up.
- `TRANSCEIVER_INFO`, `TRANSCEIVER_DOM_THRESHOLD`, VDM threshold tables (`TRANSCEIVER_VDM_{HALARM,LALARM,HWARN,LWARN}_THRESHOLD`) and `TRANSCEIVER_STATUS_SW` all had data populated as expected for each interface.

### DIFF
--- a/sonic-xcvrd/xcvrd/cpo/cpo_state_task.py
+++ b/sonic-xcvrd/xcvrd/cpo/cpo_state_task.py
@@ -7,6 +7,7 @@ try:
     from ..xcvrd import (SfpStateUpdateTask, helper_logger,
                          PHYSICAL_PORT_NOT_EXIST, SFP_EEPROM_NOT_READY)
     from ..xcvrd_utilities import common
+    from ..xcvrd_utilities.xcvr_table_helper import VDM_THRESHOLD_TYPES
     from .db_utils import CPODOMDBUtils, CPOVDMDBUtils
 except ImportError as e:
     raise ImportError(str(e) + " - required module not found")
@@ -90,3 +91,17 @@ class CpoStateUpdateTask(SfpStateUpdateTask):
     def post_port_thresholds_to_db(self, logical_port_name, dom_db_cache=None, vdm_db_cache=None):
         self.dom_db_utils.post_port_dom_thresholds_to_db(logical_port_name, db_cache=dom_db_cache)
         self.vdm_db_utils.post_port_vdm_thresholds_to_db(logical_port_name, db_cache=vdm_db_cache)
+
+    def delete_port_data_from_db(self, logical_port_name, asic_index,
+                                 delete_intf_tbl=False, delete_status_sw_tbl=False):
+        super().delete_port_data_from_db(logical_port_name, asic_index,
+                                         delete_intf_tbl=delete_intf_tbl,
+                                         delete_status_sw_tbl=delete_status_sw_tbl)
+
+        tbl_to_del_list = [
+            self.xcvr_table_helper.get_els_dom_threshold_tbl(asic_index),
+            *[self.xcvr_table_helper.get_els_vdm_threshold_tbl(asic_index, key) for key in VDM_THRESHOLD_TYPES],
+        ]
+        if delete_intf_tbl:
+            tbl_to_del_list.append(self.xcvr_table_helper.get_els_info_tbl(asic_index))
+        common.del_port_sfp_dom_info_from_db(logical_port_name, self.port_mapping, tbl_to_del_list)
--- a/sonic-xcvrd/xcvrd/cpo/cpo_state_task.py
+++ b/sonic-xcvrd/xcvrd/cpo/cpo_state_task.py
@@ -1,12 +1,92 @@
 #!/usr/bin/env python3
 
+import threading
+
 try:
-    from ..xcvrd import SfpStateUpdateTask
+    from swsscommon import swsscommon
+    from ..xcvrd import (SfpStateUpdateTask, helper_logger,
+                         PHYSICAL_PORT_NOT_EXIST, SFP_EEPROM_NOT_READY)
+    from ..xcvrd_utilities import common
+    from .db_utils import CPODOMDBUtils, CPOVDMDBUtils
 except ImportError as e:
     raise ImportError(str(e) + " - required module not found")
+
+platform_chassis = None
 
 
 class CpoStateUpdateTask(SfpStateUpdateTask):
     def __init__(self, namespaces, port_mapping, port_obj_dict, main_thread_stop_event, sfp_error_event):
         super().__init__(namespaces, port_mapping, port_obj_dict, main_thread_stop_event, sfp_error_event)
         self.name = "CpoStateUpdateTask"
+        # Replace the pluggable-transceiver DB utils installed by the base class with the CPO ones
+        self.dom_db_utils = CPODOMDBUtils(port_obj_dict, self.port_mapping, self.xcvr_table_helper,
+                                          self.task_stopping_event, self.logger)
+        self.vdm_db_utils = CPOVDMDBUtils(port_obj_dict, self.port_mapping, self.xcvr_table_helper,
+                                          self.task_stopping_event, self.logger)
+
+    def _get_port_change_event(self, timeout):
+        status, events = platform_chassis.get_elsfp_change_event(timeout)
+        elsfp_events = events.get('elsfp')
+        elsfp_errors = events.get('elsfp_error')
+        return status, elsfp_events, elsfp_errors
+
+    def _get_port_error_description(self, physical_port):
+        port_device = common.get_port_device(physical_port)
+        oe_api = port_device.oe.get_api()
+        return oe_api.get_error_description()
+
+    def _wrapper_is_replaceable(self, physical_port):
+        port_device = common.get_port_device(physical_port)
+        return port_device.is_replaceable()
+
+    def post_port_info_to_db(self, logical_port_name, port_mapping, table, transceiver_dict,
+                             stop_event=threading.Event()):
+        physical_port_list = port_mapping.logical_port_name_to_physical_port_list(logical_port_name)
+        if physical_port_list is None:
+            helper_logger.log_error("No physical ports found for logical port '{}'".format(logical_port_name))
+            return PHYSICAL_PORT_NOT_EXIST
+
+        assert len(physical_port_list) == 1, "Ganged ports are not yet supported on CPO"
+
+        for physical_port in physical_port_list:
+            if stop_event.is_set():
+                break
+
+            if not common._wrapper_get_presence(physical_port):
+                helper_logger.log_notice("Transceiver not present in port {}".format(logical_port_name))
+                continue
+
+            port_name = common.get_physical_port_name(logical_port_name, physical_port, False)
+
+            port_device = common.get_port_device(physical_port)
+            if physical_port in transceiver_dict:
+                oe_info = transceiver_dict[physical_port]
+            else:
+                oe_info = port_device.oe.get_api().get_transceiver_info()
+                transceiver_dict[physical_port] = oe_info
+            elsfp_info = port_device.elsfp.get_api().get_elsfp_info()
+
+            if oe_info is None or elsfp_info is None:
+                return SFP_EEPROM_NOT_READY
+
+            # Publish OE info
+            fvs = swsscommon.FieldValuePairs(
+                [(field, str(value)) for field, value in oe_info.items()] +
+                [('is_replaceable', str(False))]
+            )
+            table.set(port_name, fvs)
+
+            # Publish ELSFP info
+            is_replaceable = self._wrapper_is_replaceable(physical_port)
+            asic_index = port_mapping.get_asic_id_for_logical_port(logical_port_name)
+            els_tbl = self.xcvr_table_helper.get_els_info_tbl(asic_index)
+            fvs = swsscommon.FieldValuePairs(
+                [(field, str(value)) for field, value in elsfp_info.items()] +
+                [('is_replaceable', str(is_replaceable))]
+            )
+            els_tbl.set(port_name, fvs)
+
+
+    def post_port_thresholds_to_db(self, logical_port_name, dom_db_cache=None, vdm_db_cache=None):
+        self.dom_db_utils.post_port_dom_thresholds_to_db(logical_port_name, db_cache=dom_db_cache)
+        self.vdm_db_utils.post_port_vdm_thresholds_to_db(logical_port_name, db_cache=vdm_db_cache)
--- a/sonic-xcvrd/xcvrd/cpo/db_utils.py
+++ b/sonic-xcvrd/xcvrd/cpo/db_utils.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+
+try:
+    from ..dom.utilities.db.utils import DBUtils
+    from ..dom.utilities.dom_sensor.beautify import DOMBeautifyMixin
+except ImportError as e:
+    raise ImportError(str(e) + " - required module not found")
+
+
+class CPODOMDBUtils(DOMBeautifyMixin, DBUtils):
+    """
+    This class provides utility functions for managing DB operations
+    related to DOM on CPO (co-packaged optics) modules.
+    """
+
+    def __init__(self, port_obj_dict, port_mapping, xcvr_table_helper, task_stopping_event, logger):
+        super().__init__(port_obj_dict, port_mapping, task_stopping_event, logger)
+        self.xcvr_table_helper = xcvr_table_helper
+        self.logger = logger
+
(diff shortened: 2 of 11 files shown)


sonic-xcvrd/xcvrd/cpo/cpo_state_task.py
```diff
@@ -1,12 +1,107 @@
...
+        return status, elsfp_events, elsfp_errors
+
+    def _get_port_error_description(self, physical_port):
+        port_device = common.get_port_device(physical_port)
+        oe_api = port_device.oe.get_api()
+        return oe_api.get_error_description()
+
+    def _wrapper_is_replaceable(self, physical_port):
+        port_device = common.get_port_device(physical_port)
+        return port_device.is_replaceable()
```
Natgerbi: why don't we have this function only once? both for sfp and cpo? I can't see what is the benefit of override if get_port_device checking cpo and sfp

sonic-xcvrd/xcvrd/cpo/cpo_state_task.py
```diff
@@ -1,12 +1,107 @@
...
+            fvs = swsscommon.FieldValuePairs(
+                [(field, str(value)) for field, value in elsfp_info.items()] +
+                [('is_replaceable', str(is_replaceable))]
+            )
+            els_tbl.set(port_name, fvs)
+
+
+    def post_port_thresholds_to_db(self, logical_port_name, dom_db_cache=None, vdm_db_cache=None):
+        self.dom_db_utils.post_port_dom_thresholds_to_db(logical_port_name, db_cache=dom_db_cache)
+        self.vdm_db_utils.post_port_vdm_thresholds_to_db(logical_port_name, db_cache=vdm_db_cache)
```
Natgerbi: why do we have it here ? it is the same function as on SfpStateUpdateTask

sonic-xcvrd/xcvrd/dom/utilities/db/utils.py
```diff
@@ -46,19 +46,119 @@ def post_diagnostic_values_to_db(self, logical_port_name, table, get_values_func
...
+                # If cache is enabled and flag values are in cache, just read from cache, no need read from EEPROM
+                flags_dict = db_cache[physical_port]
+            else:
+                # Reading from the EEPROM as the cache is empty
+                flags_dict = get_values_func(physical_port)
+                if flags_dict is None:
+                    self.logger.log_error(f"Post {log_context} to db failed for {logical_port_name} "
+                                          "as no flags found")
+                    return
+                if flags_dict:
```
Natgerbi: you already check above if flags_dict is None, why need another check here?

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -733,6 +696,49 @@ def join(self):
...
+            self.xcvr_table_helper.get_status_flag_clear_time_tbl(asic_index),
+            self.xcvr_table_helper.get_pm_tbl(asic_index),
+            self.xcvr_table_helper.get_firmware_info_tbl(asic_index)
+        ]
+        if delete_intf_tbl:
+            tbl_to_del_list.append(self.xcvr_table_helper.get_intf_tbl(asic_index))
+        if delete_status_sw_tbl:
+            tbl_to_del_list.append(self.xcvr_table_helper.get_status_sw_tbl(asic_index))
+        common.del_port_sfp_dom_info_from_db(logical_port_name, self.port_mapping, tbl_to_del_list)
+
```
Natgerbi, line 733: I think it will be better to have something like:

```suggestion
    def get_tbls_to_clear(self, asic_id, intf=False, status_sw=False):
        tbls = self.core_tbls_to_clear(asic_id)
        if intf:
            tbls += self.intf_tbls_to_clear(asic_id)
        if status_sw:
            tbls += self.status_sw_tbls_to_clear(asic_id)
        return tbls
    def delete_port_data_from_db(self, logical_port_name, asic_index, delete_intf_tbl=False, delete_status_sw_tbl=False):
        common.del_port_sfp_dom_info_from_db(
            port_change_event.port_name, self.port_mapping,
            self.xcvr_table_helper.get_tbls_to_clear(port_change_event.asic_id, status_sw=delete_status_sw_tbl, intf=delete_intf_tbl))
```

and override the internal functions, I think it is more clear.. 

*this is not really functional comment so it is more like my approach and not really mandatory ..

Natgerbi: I see that we are using that clear tables also on DomInfoUpdateTask, so I'm thinking maybe we should return the get_tbls_to_clear from XcvrTableHelper, and maybe to create CpoXcvrTableHelper

sonic-xcvrd/xcvrd/cpo/cpo_state_task.py
bgallagher-nexthop: I had initially done this to avoid any risk of regression in SfpStateUpdateTask. This makes sense to me though, consolidated to a single definition in SfpStateUpdateTask that both tasks use.

sonic-xcvrd/xcvrd/cpo/cpo_state_task.py
bgallagher-nexthop: I had overriden in CPO in case we would later like to customize this logic by caching the static information on a per-device basis, but that is probably premature. Deleted so that we only have a single definition in `SfpStateUpdateTask`.

sonic-xcvrd/xcvrd/xcvrd.py
bgallagher-nexthop: Good point. I have subclassed XcvrTableHelper for CPO and moved more of the table selection logic into that class which allows us to delete the `delete_port_data_from_db` override in `CpoStateUpdateTask` too.

abelamit (dismissed):

sonic-xcvrd/xcvrd/cpo/cpo_state_task.py
```diff
@@ -1,12 +1,90 @@
...
+            return PHYSICAL_PORT_NOT_EXIST
+
+        assert len(physical_port_list) == 1, "Ganged ports are not yet supported on CPO"
+
+        for physical_port in physical_port_list:
+            if stop_event.is_set():
+                break
+
+            if not common._wrapper_get_presence(physical_port):
+                helper_logger.log_notice("Transceiver not present in port {}".format(logical_port_name))
```
aditya-nexthop: ELSFP?

sonic-xcvrd/xcvrd/cpo/cpo_state_task.py
```diff
@@ -1,12 +1,90 @@
...
+                continue
+
+            port_name = common.get_physical_port_name(logical_port_name, physical_port, False)
+
+            port_device = common.get_port_device(physical_port)
+            if physical_port in transceiver_dict:
+                oe_info = transceiver_dict[physical_port]
+            else:
+                oe_info = port_device.oe.get_api().get_transceiver_info()
+                transceiver_dict[physical_port] = oe_info
```
aditya-nexthop: nitpick for readability:

```
if physical_port not in transceiver_dict:
   transceiver_dict[physical_port] = port_device.oe.get_api().get_transceiver_info()
oe_info = transceiver_dict[physical_port]
```

sonic-xcvrd/xcvrd/dom/utilities/db/utils.py
```diff
@@ -46,19 +46,119 @@ def post_diagnostic_values_to_db(self, logical_port_name, table, get_values_func
...
+            values_dict (dict): Pre-read diagnostic values. Beautified in place.
+            beautify_func (function, optional): Function to beautify the values. Defaults to self.beautify_info_dict.
+            enable_flat_memory_check (bool, optional): Flag to check for flat memory support. Defaults to False.
+        """
+        physical_port = self._validate_and_get_physical_port(logical_port_name, enable_flat_memory_check)
+        if physical_port is None:
+            return
+
+        if not values_dict:
+            return
```
aditya-nexthop: nit: can combine as:
```
if physical_port is None or not values_dict:
   return
```

sonic-xcvrd/xcvrd/dom/utilities/status/db_utils.py
```diff
@@ -45,47 +45,11 @@ def post_port_transceiver_hw_status_flags_to_db(self, logical_port_name, db_cach
...
-                    [("last_update_time", self.get_current_time())]
-                )
-                self.xcvr_table_helper.get_status_flag_tbl(asic_index).set(logical_port_name, fvs)
-            else:
-                return
-
-        except NotImplementedError:
-            self.logger.log_notice(f"Post port transceiver hw status flags to db failed for {logical_port_name} "
-                                   "as functionality is not implemented")
-            return
```
aditya-nexthop, line 91: great work getting rid of duplicate code!

sonic-xcvrd/xcvrd/xcvrd.py
aditya-nexthop, line 158: should we consider moving this into its own file?

aditya-nexthop (approved):

sonic-xcvrd/xcvrd/xcvrd.py
bgallagher-nexthop: I think it would make sense to move into its own file, if only to mirror the structure of the other tasks being in their own file. It is probably worth doing in a separate PR though, since it will probably make this already large PR a lot bigger since it isn't a clean lift-and-shift of the class into its own file. A bunch of imports (+ the ones in tests), constants and global helpers also need to be moved along with the class

abelamit (dismissed):

abelamit (dismissed):

sonic-xcvrd/xcvrd/cpo/cpo_state_task.py
```diff
@@ -1,12 +1,88 @@
...
+
+    def create_vdm_db_utils(self, port_obj_dict, port_mapping, xcvr_table_helper,
+                            task_stopping_event, logger) -> CPOVDMDBUtils:
+        return CPOVDMDBUtils(port_obj_dict, port_mapping, xcvr_table_helper, task_stopping_event, logger)
+
+    def _get_port_change_event(self, timeout):
+        status, events = platform_chassis.get_elsfp_change_event(timeout)
+        elsfp_events = events.get('elsfp')
+        elsfp_errors = events.get('elsfp_error')
+        return status, elsfp_events, elsfp_errors
```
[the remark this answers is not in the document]
bgallagher-nexthop: This is done on purpose. The `CpoStateUpdateTask` should never run if `platform_chassis` is not initialized and `get_elsfp_change_event` must be implemented and present for this task to do its job.

sonic-xcvrd/xcvrd/cpo/cpo_state_task.py
```diff
@@ -1,12 +1,88 @@
...
+        return oe_api.get_error_description()
+
+    def post_port_info_to_db(self, logical_port_name, port_mapping, table, transceiver_dict,
+                             stop_event=threading.Event()):
+        physical_port_list = port_mapping.logical_port_name_to_physical_port_list(logical_port_name)
+        if physical_port_list is None:
+            helper_logger.log_error("No physical ports found for logical port '{}'".format(logical_port_name))
+            return PHYSICAL_PORT_NOT_EXIST
+
+        assert len(physical_port_list) == 1, "Ganged ports are not yet supported on CPO"
```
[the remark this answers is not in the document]
bgallagher-nexthop: Done

sonic-xcvrd/xcvrd/cpo/db_utils.py
```diff
@@ -0,0 +1,58 @@
...
+    related to VDM on CPO (co-packaged optics) modules.
+    """
+
+    def __init__(self, port_obj_dict, port_mapping, xcvr_table_helper, task_stopping_event, logger):
+        super().__init__(port_obj_dict, port_mapping, task_stopping_event, logger)
+        self.xcvr_table_helper = xcvr_table_helper
+        self.logger = logger
+
+    def post_port_vdm_thresholds_to_db(self, logical_port_name, db_cache=None):
+        pass
```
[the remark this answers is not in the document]
bgallagher-nexthop: This is intended. VDM support is not yet implemented.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -631,7 +627,7 @@ def task_worker(self, stopping_event, sfp_error_event):
                                         if error_dict:
                                             vendor_specific_error_description = error_dict.get(key)
                                         else:
-                                            vendor_specific_error_description = _wrapper_get_sfp_error_description(key)
+                                            vendor_specific_error_description = self._get_port_error_description(key)
                                         error_descriptions.append(vendor_specific_error_description)
```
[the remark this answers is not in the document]
bgallagher-nexthop: Done

sonic-xcvrd/xcvrd/dom/utilities/db/utils.py
```diff
@@ -91,7 +188,7 @@ def _validate_and_get_physical_port(self, logical_port_name, enable_flat_memory_
 
         physical_port = pport_list[0]
 
-        if physical_port not in self.sfp_obj_dict:
+        if physical_port not in self.port_obj_dict:
             self.logger.log_error(f"Validate and get physical port failed for {logical_port_name} "
                                    "as no sfp object found")
             return None
```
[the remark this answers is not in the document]
bgallagher-nexthop: Done

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -322,6 +230,116 @@ def _mapping_event_from_change_event(self, status, port_dict):
...
+                status, events = platform_chassis.get_change_event(timeout)
+                sfp_events = events.get('sfp')
+                sfp_errors = events.get('sfp_error')
+                return status, sfp_events, sfp_errors
+            except NotImplementedError:
+                pass
+        status, events = platform_sfputil.get_transceiver_change_event(timeout)
+        return status, events, None
+
+    def _wrapper_is_replaceable(self, physical_port):
```
vdahiya12, line 253: could we add the exception path ?
```
if platform_chassis is not None:
        try:
            return platform_chassis.get_sfp(physical_port).is_replaceable()
        except NotImplementedError:
            pass
    return False
```

bgallagher-nexthop: It still exists -- [`get_port_device` returns None if `NotImplementedError` is thrown by `get_sfp`](https://github.com/sonic-net/sonic-platform-daemons/blob/a9d71d7218e72423b3b993c39cfb07f1cdac8705/sonic-xcvrd/xcvrd/xcvrd_utilities/common.py#L136-L137), and this function returns `False` in the event no backing object is found for the given physical port.

sonic-xcvrd/xcvrd/xcvrd.py
```diff
@@ -567,28 +585,27 @@ def task_worker(self, stopping_event, sfp_error_event):
...
                                     if not self.is_warm_fast_reboot_for_lport(logical_port):
                                         media_settings_parser.notify_media_setting(logical_port, transceiver_dict, self.xcvr_table_helper, self.port_mapping)
                                     transceiver_dict.clear()
                             elif value == sfp_status_helper.SFP_STATUS_REMOVED:
-                                # Remove the SFP API object for this physical port
+                                # Remove the API object for this physical port
                                 try:
-                                    sfp = platform_chassis.get_sfp(int(key))
-                                    sfp.remove_xcvr_api()
+                                    device = common.get_port_device(int(key))
```
vdahiya12, line 614: Could we check physical_port in self.port_obj_dict before this call, since common.get_port_device() can return a conventional SFP without the CPO-specific ?

sonic-xcvrd/xcvrd/xcvrd.py
bgallagher-nexthop: Oh sorry, I see what you mean. The `is_replaceable` call is no longer protected. Updated the PR to address this.

sonic-xcvrd/xcvrd/xcvrd.py
bgallagher-nexthop: Good point, I've added a membership check with an error log if the platform hook returns a change event for a port not owned by the task.

vdahiya12 (approved):

PR #886, opened 2026-09-01
[stormond] prevent cumulative FSIO totals from collapsing during FSRO

gauravnagesh-msft:
#### Description

When a storage disk is inaccessible (e.g. a read-only like FSRO event), psutil.disk_io_counters() raises and the StorageCommon FSIO getters return 0. stormond previously fed that 0 into _reconcile_fsio_rw_values(), collapsing the lifetime totals derived from the latest per-poll counter. This was visible in syslog when firmware/health/temp went N/A:

  FS IO Writes: 20,938,935 -> 232,834
  Latest FSIO Writes: 20,706,101 -> 0

A lifetime cumulative counter must never regress. 

```
NOTICE pmon#stormond[40]: Polling Interval set to 3600 seconds
NOTICE pmon#stormond[40]: FSIO JSON file Interval set to 86400 seconds
NOTICE pmon#stormond[40]: Storage Device: nvme0n1, Firmware: XXXXXXXX, health: 100.0%, Temp: 49.0C, FS IO Reads: 57330, FS IO Writes: 20938935
NOTICE pmon#stormond[40]: Latest FSIO Reads: 29166, Latest FSIO Writes: 20706101
NOTICE pmon#stormond[40]: Disk IO Reads: 222,007 [113 GB], Disk IO Writes: 625,653 [320 GB], Reserved Blocks: 100.0
NOTICE pmon#stormond[40]: Last successful sync time to STATE_DB: 2026-08-08 18:36:41

// After disk inaccessible 

NOTICE pmon#stormond[40]: Polling Interval set to 3600 seconds
NOTICE pmon#stormond[40]: FSIO JSON file Interval set to 86400 seconds
NOTICE pmon#stormond[40]: Storage Device: nvme0n1, Firmware: N/A, health: N/A%, Temp: N/AC, FS IO Reads: 28164, FS IO Writes: 232834
NOTICE pmon#stormond[40]: Latest FSIO Reads: 0, Latest FSIO Writes: 0
NOTICE pmon#stormond[40]: Disk IO Reads: 222,007 [113 GB], Disk IO Writes: 625,653 [320 GB], Reserved Blocks: 100.0
NOTICE pmon#stormond[40]: Last successful sync time to STATE_DB: 2026-08-08 19:36:41
```

#### Motivation and Context

To prevent these values from collapsing, when both latest reads and writes are 0 (for an active disk these /proc/diskstats counters are effectively never 0, and a read failure zeroes both together), retain the last known-good latest and total values from STATE_DB so the totals stay monotonic. Requiring both to be 0 avoids false positives on an idle disk; the only false positive is a fully unused disk whose last-good is already 0, making it harmless. Last-good latest (not 0) is persisted so crash-restart reconciliation stays correct, while syslog still shows 0 to signal the failure.

Adds `_get_last_fsio_statedb_value()` helper and tests for the FSRO path.

Tested on local testbed and added testcases `test_get_dynamic_fields_fsio_counters_unavailable`  and `test_get_last_fsio_statedb_value_defaults_to_zero` to verify the scenario and changes.

### DIFF
--- a/sonic-stormond/scripts/stormond
+++ b/sonic-stormond/scripts/stormond
@@ -254,6 +254,14 @@ class DaemonStorage(daemon_base.DaemonBase):
         # that akin to an INIT state. 
 
 
+    # Fetch the last known-good value of an FSIO field from STATE_DB. This is used
+    # to retain the previous latest/total FSIO values (keeping cumulative totals
+    # monotonic) when the live /proc/diskstats counters cannot be read.
+    def _get_last_fsio_statedb_value(self, device, field):
+        value = self.state_db.hget('{}|{}'.format(STORAGE_DEVICE_TABLE, device), field)
+        return "0" if value is None else value
+
+
     def _reconcile_fsio_rw_values(self, fsio_dict, device):
 
         # If stormond is coming up for the first time, neither STATE_DB info nor JSON file would be present.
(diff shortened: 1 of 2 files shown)


judyjoseph (approved):
LGTM

